PR #2, opened 2019-12-13
DDP-4230: SDK component to handle email verification step of Age-Up invitation

fat23cat:
- Added SDK component to handle email verification
- Added Toolkit page
- Added generic AgeUpThankYou component
- Updated Toolkit configuration to handle a case when proxy submits child's email


ddp-workspace/projects/ddp-sdk/src/lib/components/age-up/verifyAgeUp.component.ts
```diff
@@ -0,0 +1,38 @@
...
+            take(1)
+        ).subscribe(status => {
+            if (status === this.EMAIL_VERIFIED) {
+                this.nextUrl.emit(this.url);
+            } else {
+                this.error.emit();
+            }
+        });
+    }
+}
```
MocanaAtBroad, line 32: This `VerifyAgeUpComponent` component looks to me more like a service (i.e., provides functionality that is primarily not visible). The component using this functionality can easily include its own waiting icon.

MocanaAtBroad: Also, I would be inclined to push down into the the new ServiceAgent the interpretation of the `200` code. You can throw an error object that the caller of verifyService.verify can directly or indirectly handle in the observable pipeline.

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/invitationsServiceAgent.service.ts
```diff
@@ -0,0 +1,22 @@
...
+        @Inject('ddp.config') _configuration: ConfigurationService,
+        _http: HttpClient,
+        _logger: LoggingService) {
+        super(_configuration, _http, _logger);
+    }
+
+    public verify(invitationId: string): Observable<number | null> {
+        return this.postObservable('/invitations/verify', { invitationId })
+            .pipe(map(response => response && response.status ? response.status : null));
+    }
```
MocanaAtBroad, line 25: See my comment above about handing status code. 
And if this is really just a "verify worked"/"verify did not work" type of call, perhaps a Observable<boolean> would be more appropriate?

ddp-workspace/projects/ddp-osteo/src/app/app.module.ts
```diff
@@ -63,6 +63,7 @@ tkCfg.parentalConsentUrl = 'parental-consent';
 tkCfg.releaseMinorUrl = 'release-minor-survey';
 tkCfg.releaseUrl = 'release-survey';
 tkCfg.dashboardUrl = 'dashboard';
+tkCfg.doneUrl = 'proxy-thank-you';
```
MocanaAtBroad, line 66: You sure about you want the DONE workflow state to always be mapped to going to the Age Up thank you? (At least that is what it looks like Osteo is now configured to do.

fat23cat: Yes, I discussed it with @yufengwng

yufengwng: Yup, that's the current plan for Osteo's configuration. Better ideas here are welcomed!

ddp-workspace/projects/ddp-sdk/src/lib/components/age-up/verifyAgeUp.component.ts
fat23cat: I tend to leave this as a component because it allows saving age-up components uniformity, and works with queryParams only when it is really necessary because the service complicates the processing queryParams.

I like your idea about handling the error.

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/invitationsServiceAgent.service.ts
MocanaAtBroad: `empty()` appears to be deprecated in favor of `EMPTY`.

ddp-workspace/projects/ddp-sdk/src/lib/components/age-up/verifyAgeUp.component.ts
```diff
@@ -24,9 +24,7 @@ export class VerifyAgeUpComponent implements OnInit {
     }
 
     private verifyInvitation(invitationId: string): void {
-        this.invitationsService.verify(invitationId).pipe(
-            take(1)
-        ).subscribe({
+        this.invitationsService.verify(invitationId).subscribe({
             complete: () => this.nextUrl.emit(this.url),
```
MocanaAtBroad, line 28: You took out the `take(1)`. Not worried about disposing of subscription?

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/invitationsServiceAgent.service.ts
fat23cat: Good catch, thanks.

ddp-workspace/projects/ddp-sdk/src/lib/components/age-up/verifyAgeUp.component.ts
fat23cat: Yes, because this observable has only two states: complete and error, it means that observable finished, so we don't need to unsubscribe.

MocanaAtBroad: Being reading (more!!) about the unsubscribe issue.
I am not saying you are wrong, but I have encountered documentation or StackOverflow entry that would indicate that the subscription would be cleaned up using this type of configuration.
Can you point to a reference? If not, then adding the `take(1)` is not a big deal anyways.

fat23cat: https://stackoverflow.com/a/45478765

MocanaAtBroad (approved):

PR #4, opened 2019-12-13
Prion appengine proof of concept

zyme:
POC for frontend on app engine.  Builds and deploys cleanly, but deployed code has some issues due to missing config values such as `clientID` somewhere along the way.

## Build instructions
1. `npm install` as per usual
2. [Run the deploy script](https://github.com/broadinstitute/ddp-angular/pull/4/files#diff-931ff73d02f12cb802e8396015015137)
3.  VIew in browser!

### DIFF
--- a/ddp-workspace/projects/ddp-prion/app.yaml
+++ b/ddp-workspace/projects/ddp-prion/app.yaml
@@ -1,6 +1,7 @@
 service: prion
-runtime: python
+runtime: python27
 instance_class: B1
+threadsafe: true
 
 manual_scaling:
   instances: 1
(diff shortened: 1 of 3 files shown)


ddp-workspace/projects/ddp-prion/app.yaml
```diff
@@ -0,0 +1,25 @@
+service: prion
+runtime: python27
```
zyme, line 2: Looks weird!  But it works, since this is all static assets.  No point in using node or java, since we're not actually running anything per se, just loading static assets.

fat23cat:
@zyme, @charlottedye can we remove `prion-appengine` branch?

PR #5, opened 2019-12-16
DDP-4295: Lazy resources

fat23cat:
- We can use lazy loading not only for images, but we can also load lazily all resources which have `src` attribute(img, iframe, video). For this reason, the directive was re-implemented and re-named.
- We can configure `rootMargin` property in `IntersectionObserver` object. It allows preload resources while they stay invisible for a user, but a user will see them soon. It improves UX, a user does not need to wait until resources load.


MocanaAtBroad (approved):
Works nice! I am curious if we could extend this to work with the Twitter and Instagram feeds. Presumably we don't have to load the components or the supporting external scripts until they become visible...

fat23cat:
> Works nice! I am curious if we could extend this to work with the Twitter and Instagram feeds. Presumably we don't have to load the components or the supporting external scripts until they become visible...

Yeah, I thought about it. We can create own IntersectionObservers for widgets and run ScriptLoader and show widget when browser viewport intersects with a widget.

PR #13, opened 2019-12-20
DDP-4233 Angular email question

MocanaAtBroad:
Work on this PR is around `activityEmailAnswer.component.ts` and its validation.
Note that this seems the first component where we don't get the error messages when validation is incorrect from server.
Tried to internationalize messages, so created `en.json` for all projects except basil-app.
Unfortunately we don't yet have an activity that includes an EMAIL question to kick around. For my local testing I just modified the Osteo configuration and added email questions.
I did include a new sandbox component, but it does look different.
What it looks like in Osteo (Note that I set up question with a placeholder value of 'Email'):
Screen Shot 2019-12-20 at 5 07 53 PM
The confirmation placeholder is generated by adding an internationalized prefix to the question placeholder value (possible future improvement is to include separate placeholder for confirmation in API).
I sprinkled comments in files.


ddp-workspace/projects/ddp-angio/src/assets/i18n/en.json
```diff
@@ -568,6 +568,13 @@
             "PluralForm": "characters remaining",
             "SingularForm": "character remaining"
         },
+        "EmailEntry": {
+          "ConfirmationEmailPlaceholder": "Confirm {{value}}",
```
MocanaAtBroad, line 572: Tried to do interpolation.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityAnswer.component.ts
```diff
@@ -13,7 +16,7 @@ import { BlockVisibility } from '../../models/activity/blockVisibility';
                                  [readonly]="readonly"
                                  (valueChanged)="onChange($event)">
     </ddp-activity-boolean-answer>
-    <ddp-activity-text-answer *ngIf="isTextQuestion(block) && block.shown"
+    <ddp-activity-text-answer *ngIf="isTextQuestion(block) && !isEmailQuestion(block) && block.shown"
                               [block]="block"
```
MocanaAtBroad, line 20: From API perspective, email question is just a text question. But it works and looks very different than a regular text question, so some special treatment.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityEmailAnswer.component.ts
```diff
@@ -0,0 +1,106 @@
...
+  ngOnInit(): void {
+    this.initFormControls();
+    this.initPlaceholders();
+  }
+
+  private initFormControls() {
+    this.formControls = [0, 1].map(_ => new FormControl({
+      value: this.block.answer,
+      disabled: this.readonly
+    }, [Validators.email]));
```
MocanaAtBroad, line 56: Reactive component includes email validator, so used reactive instead of ngModel. Also more flexibility.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityEmailAnswer.component.ts
```diff
@@ -0,0 +1,106 @@
...
+    if (changes['block'] && !changes['block'].isFirstChange()) {
+      this.formControls.forEach(control => control.patchValue(this.block.answer, { onlySelf: true, emitEvent: false }));
+    }
+    if (!_.isUndefined(changes['readonly']) && !changes['readonly'].isFirstChange()) {
+      this.formControls.forEach(control =>
+        this.readonly ?
+          control.disable({ onlySelf: true, emitEvent: false }) :
+          control.enable({ onlySelf: true, emitEvent: false }));
+    }
+  }
```
MocanaAtBroad, line 96: The whole onChanges business is to enable pressing the Update button in sandbox to change the parameters used there. Not relevant under "normal" circumstances.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityTextAnswer.component.ts
```diff
@@ -83,12 +83,12 @@ export class ActivityTextAnswer implements OnChanges, OnInit {
     }
 
     public isTextQuestion(block: ActivityTextQuestionBlock): boolean {
-        return !block.inputType || block.inputType.toLowerCase() === InputType.Text
-            || block.inputType.toLowerCase() === InputType.Signature;
+        return !block.inputType || block.inputType === InputType.Text
+            || block.inputType === InputType.Signature;
```
MocanaAtBroad, line 87: Since I needed to add the `Email` InputType, I took the opportunity to just use it as straight up `enum` (i.e., no need to compare strings)

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityEmailQuestionBlock.ts
```diff
@@ -0,0 +1,23 @@
+import { ActivityTextQuestionBlock } from './activityTextQuestionBlock';
+
+export class ActivityEmailQuestionBlock extends ActivityTextQuestionBlock {
+  public email: string | null = null;
+  public emailConfirmation: string | null = null;
+  constructor() {
```
MocanaAtBroad, line 6: This is one of the weirdnesses that needed to figure out with this component. For all others we have had just one value that we care about (granted ,that value could be an array). Here server only cares about the email, but client cares about confirmation value as well for validation purposes, so had to get creative to fit into our existing scheme of one block/one value.

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityEmailQuestionBlock.ts
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  get answer(): string | null {
+    return this.email;
+  }
+
+  set answer(value: string | null) {
+    this.email = this.emailConfirmation = value;
+    this.validate();
+  }
```
MocanaAtBroad, line 17: If we do set a value for the block, we are going to start off email and confirm value being the same. Only the component that works with emails will care.

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityEmailValidatorRule.ts
```diff
@@ -0,0 +1,58 @@
+import { ActivityAbstractValidationRule } from './activityAbstractValidationRule';
+import { ActivityEmailQuestionBlock } from '../../../models/activity/activityEmailQuestionBlock';
+import { NGXTranslateService } from '../../../services/internationalization/ngxTranslate.service';
+import { take } from 'rxjs/operators';
+
+// Need to hold off on this expression for now until https://broadinstitute.atlassian.net/browse/DDP-4311 is fixed
+// const EMAIL_REGEXP =
+   // /^(?=.{1,254}$)(?=.{1,64}@)[a-zA-Z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-zA-Z0-9!#$%&'*+/=?^_`{|}~-]+)*@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/;
+
```
MocanaAtBroad, line 9: Started with this expression (borrowed from Angular), but was letting through values that the server was rejecting.
So below I reproduced the regex used on the Pepper back-end.

ddp-workspace/projects/ddp-sdk/src/lib/services/internationalization/ngxTranslate.service.ts
```diff
@@ -7,11 +7,11 @@ import { mergeMap, merge } from 'rxjs/operators';
...
 
-    public getTranslation(word: string | Array<string>): Observable<object | string> {
+    public getTranslation(word: string | Array<string>, interpolateParams?: Object): Observable<object | string> {
         return of(null).pipe(
             merge(this.translate.onLangChange),
             merge(this.translate.onDefaultLangChange),
-            mergeMap(() => this.translate.get(word))
+            mergeMap(() => this.translate.get(word, interpolateParams))
         );
     }
```
MocanaAtBroad, line 16: Interpolation params can be used now.

fat23cat:
@MocanaAtBroad, I want to discuss changes in architecture and new component in general.
1. Components should receive text parameters from the backend:
image
Moreover, the second input should be conditional. I believe @ssettipalli implemented the component's parameters on the backend.
2. Initially, looking through PR I got the impression that Email question it's a new Question block, but in reality, it's subtype for TEXT question block. **IMO** it is confusing and complicates components' structure maintainability and readability. I think:
   - We shouldn't create `ActivityEmailQuestionBlock` we should add new optional parameters in the `ActivityTextQuestionBlock`
   - We shouldn't add `ActivityEmailAnswerComponent` directly in the `ActivityAnswerComponent`, it should be a part of `ActivityTextAnswer`. It should look like this:
image
(Pseudocode example)
   - `ActivityEmailAnswerComponent` should be fully independent - it's not necessary to create validation rules, we can use Angular Forms, we should validate values, show error messages if email mismatch, etc. The component should emit/save an answer in the `ActivityTextQuestionBlock` only if it's valid. It allows reuse `ActivityTextQuestionBlock` instead of creating new structures.

fat23cat:
@MocanaAtBroad, @zyme, due to the fact that the weekend is in the USA, and then the weekend will be in Russia, we don't have a possibility to discuss these changes, so, I opened another draft PR #16  with fixes and improvements which I suggested above. Please, take a look when you back from weekends.

PR #17, opened 2019-12-25
DDP-4312: Workflow progress component is shown on the Provide contact information page

fat23cat:
Bug description is here https://broadinstitute.atlassian.net/browse/DDP-4312


ddp-workspace/projects/ddp-osteo/src/app/components/workflow-progress/workflow-progress.component.ts
```diff
@@ -13,6 +13,7 @@ export class WorkflowProgressComponent implements OnChanges {
   @Input() public currentActivity: string;
```
MocanaAtBroad: Would you mind annotating with what this attribute means? Is this the activity instance guid or activity  guid or something else? (I think it is the Activity GUID) but needed to do some digging...

ddp-workspace/projects/ddp-osteo/src/app/components/workflow-progress/workflow-progress.component.ts
```diff
@@ -13,6 +13,7 @@ export class WorkflowProgressComponent implements OnChanges {
...
   @Input() public workflowStartSectionsVisibility: number | null;
```
MocanaAtBroad, line 17: Similar annotation/comment needed here.

MocanaAtBroad (dismissed):

MocanaAtBroad (approved):

PR #18, opened 2019-12-25
DDP-4313: Provide contact information page does not match layout

fat23cat:
The description is here - DDP-4313
Backend changes - https://github.com/broadinstitute/ddp-study-server/pull/24


MocanaAtBroad (commented):
Not seeing where this style is used.

fat23cat:
> Not seeing where this style is used.

It used on the backend side - https://github.com/broadinstitute/ddp-study-server/blob/develop/study-builder/studies/osteo/child-contact.conf#L47

MocanaAtBroad (approved):

PR #19, opened 2019-12-26
Angular update

fat23cat:
Last Angular and Angular Material updates.


Dockerfile
```diff
@@ -30,7 +30,7 @@ RUN ln -s /usr/local/bin/node /usr/local/bin/nodejs
 
 # install angular in a separate cached layer
 RUN mkdir -p /ddp-workspace/projects
-ENV ANGULAR_CLI @angular/cli@8.2.0
+ENV ANGULAR_CLI @angular/cli@8.3.21
 RUN npm install -g $ANGULAR_CLI
```
MocanaAtBroad, line 34: @zyme We still using Dockerfile s for deployment or have we moved to using Google App Engine configurations?

MocanaAtBroad (commented):
Changes look good `ng serve` and prod builds I attempted worked without a problem. Not sure if we still using Dockerfile so have asked Zim to chime in.

Dockerfile
fat23cat: @zyme, can you help us? ^^

zyme: Sorry for the delay!  Yes, we are still using Docker for our deploys--hopefully not too much longer, but we will need to keep doing these sorts of changes for now.

MocanaAtBroad (approved):

PR #1, opened 2019-12-12
DDP-4292: Refactor mail address component

MocanaAtBroad:
Picking up the pieces from PR in old repo at https://github.com/broadinstitute/ddp/pull/1056
Trying to address comments in old PR there with links to corresponding commits in this repo.
**May 8 2020**
- Added test result reporting which forgot about first time around.
- Appears to be working (test failure fails job AND test result show in CircleCi).
https://app.circleci.com/pipelines/github/broadinstitute/ddp-angular/1356/workflows/74ebc0d1-5074-40e3-a4de-460015be14c9/jobs/1715/tests
- Added .gitignore update
- Fixed null check.

**May 7 2020**
Cleaned up all the tests we have in repo. They all now run. Recent commits change the updates
Also added a run-tests command to CircleCi that executes whenever we run a build.
Finally, had to figure out how to setup and run Chrome headless in Docker. See Dockerfile and ddp-sdk/karma.conf.js for those changes.
This CircleCI workflow run shows the tests in action:
https://app.circleci.com/pipelines/github/broadinstitute/ddp-angular/1328/workflows/b90ed55d-5b60-48b5-976c-da936d12d3e1/jobs/1698

**Dec 16 2019**
Still need to deal with the console logs and a couple of other things.

PR #21, opened 2020-01-03
Tweaks to Pavel's implementation of email text question

MocanaAtBroad:
Was able to try out proposed version against local environment with email questions with confirmations in wizard activity and "regular" activities and found some problems.
Specifically, some email values were being sent to server that server was rejecting. This was due to format used in local email string validation did not match the format on server.
Another problem was sending `null` for a value. Server does not like that, so we are now sending empty strings.
Also, the validation of emails was being ignored in wizard (I think this was a more general problem in wizard that I have tried to address here as well).
I think my commit messages are fairly descriptive so for review might make sense to go through commit changes rather (or in addition) to just file changes.


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityEmailInput.component.ts
```diff
@@ -66,63 +73,82 @@ export class ActivityEmailInput implements OnChanges {
...
+        const EmailPatternValidator = Validators.pattern(EMAIL_REGEXP);
+
         if (this.block.confirmEntry) {
+
             this.emailForm = this.formBuilder.group({
                 email: new FormControl({
                     value: this.block.answer,
                     disabled: this.readonly
-                }, Validators.email),
+                }, EmailPatternValidator),
```
MocanaAtBroad, line 104: Use our own validator to match back-end email format checker.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityEmailInput.component.ts
```diff
@@ -66,63 +73,82 @@ export class ActivityEmailInput implements OnChanges {
...
+                this.block.setAnswer(cleanedFormData.email, false);
+                this.block.confirmEntry && (this.block.confirmationValue = cleanedFormData.confirmEmail);
+            }),
+            // we only emit data we know server will accept
+            filter((_) => this.block.canPatch()),
+            map((cleanedFormData) => !!cleanedFormData.email ? cleanedFormData.email : ''),
+            distinctUntilChanged(),
+            tap((val) => this.valueChanged.emit(val))
+        ).subscribe();
     }
```
MocanaAtBroad, line 137: Big change here it to handle the value changes from reactive pipeline rather than from an onChange handler.
We get the raw changes, clean them up, store them in the block, call the block validation and emit if data looks valid.
Note that we send valid emails or empty strings.
Also noteworthy is that we store the confirmationValue in the block. We do this because our pre-existing validation scheme has validators working off data values in block, so that is why we extended the text block and set the confirmationValue; to allow the EmailValidatorRule to figure out if our email is valid or not.

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityTextQuestionBlock.ts
```diff
@@ -11,12 +11,27 @@ export class ActivityTextQuestionBlock extends ActivityQuestionBlock<string> {
...
+
+    public set answer(value: string | null) {
+        this.setAnswer(value);
+    }
+
+    setAnswer(value: string | null, doValidation: boolean = true): void {
+      this._answer = value;
+      this.confirmEntry && (this.confirmationValue = value);
+      doValidation && this.validate();
     }
```
MocanaAtBroad: our "blocks" only take a single answer value, but we need to somehow initialize the confirmationValue .

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/activityQuestionConverter.service.ts
```diff
@@ -125,7 +132,7 @@ export class ActivityQuestionConverter {
                 }
             },
             {
-                type: 'AGREEMENT', func: (questionJson) => {
+                type: 'AGREEMENT', func: () => {
```
MocanaAtBroad, line 128: Getting rid of tslint warning.

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityEmailValidatorRule.ts
```diff
@@ -0,0 +1,53 @@
...
+  constructor(
+    question: ActivityTextQuestionBlock, translate: NGXTranslateService) {
+    super(question);
+    translate.getTranslation(TRANSLATION_KEYS
+    )
+      .pipe(take(1))
+      .subscribe((vals) => {
+        this.badEmailMessage = vals[TRANSLATION_KEYS[0]];
+      });
+  }
```
MocanaAtBroad: A bit awkward to get translated message into validator. Open to suggestions.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityTextAnswer.component.ts
```diff
@@ -90,7 +90,9 @@ export class ActivityTextAnswer implements OnChanges, OnInit {
     }
 
     public onChange(value: string): void {
-        const answer = value ? value : null;
+        // turns out server will not accept null for text questions
+        // send emit empty string for falsy values
+        const answer = value ? value : '';
```
fat23cat, line 95: Actually, answers can be nullable https://github.com/broadinstitute/ddp-study-server/blob/develop/pepper-apis/docs/specification/src/components/schemas.yml#L259

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityEmailInput.component.ts
```diff
@@ -66,63 +73,82 @@ export class ActivityEmailInput implements OnChanges {
...
+
             this.emailForm = this.formBuilder.group({
                 email: new FormControl({
                     value: this.block.answer,
                     disabled: this.readonly
-                }, Validators.email),
+                }, EmailPatternValidator),
                 confirmEmail: new FormControl({
-                    value: this.block.answer,
+                    value: this.block.confirmationValue,
```
fat23cat: We can save on the backend only valid answer(email), so, we don't need to store confirmationValue in the block and we can show the answer in both fields.

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityTextQuestionBlock.ts
fat23cat: Marco, I want to discuss this logic...
The main idea that I wanted to show in my implementation is that blocks shouldn't know about the confirmation filed.
Email input type no different from other components - there should store only one answer, so, can save the email input type answer in the block.answer only if form valid, we shouldn't store both answers. Block shouldn't validate answers, we can delegate it to Angular form in the email input type component.
I am inclined to believe that we should remove new logic in the `ActivityQuestionBlock`, because we mixed validation logic and storing answers; remove `ActivityEmailValidatorRule`, because we can delegate validation to Angular form; remove this logic https://github.com/broadinstitute/ddp-angular/blob/f3372ae9cffe9960ccd9594a234e37e5160a6c6a/ddp-workspace/projects/ddp-sdk/src/lib/services/activity/activityQuestionConverter.service.ts#L66 because we delegate validation to Angular form.
My idea is - we don't need external validation rules, Angular form in the email input type component can validate all the stuff and store only valid email address.

MocanaAtBroad: Thanks for taking a look during vacation. Hopefully you still taking time to relax.
We can talk further about this topic on Monday, but just know that I am not adding this without careful consideration.

My #1 least favorite thing in our system is how we do validations (stateful, triggered multiple times and sometimes by setting a flag, etc.etc) . But as things stand right now, validation error messages displayed on activity instance pages ("wizard" and "regular") are generated by validator rules that are attached to question "blocks".

The email validator rule could of course be added in some other way than what I have done here, but essentially same functionality needs to be added somewhere.

I like what you did with FormGroup validator (and left it in there mostly unchanged), but this validation would not prevent the final activity PUT of the activity or display an error message similar to what we have in the other question components (i.e., `<ddp-validation-message>`).

At any rate, don't worry about this now. Let's talk Monday. We should talk about just validations in general too, which as I said in the beginning, not really happy with.

fat23cat: Don't worry, I had a few munites to take a look at PR, but I swear, I won't open GitHub till the end of the vacation 😄

> but this validation would not prevent the final activity PUT of the activity or display an error message similar to what we have in the other question components (i.e., <ddp-validation-message>).

I guess the REQUIRED validation rule should handle it.
Moreover, for backend valid answers for email input type are strings and null, my code
```TS
    public onChange(answer: string): void {
        const allFieldsTouched = Object.values(this.emailForm.controls).every(control => control.touched);
        if (this.emailForm.valid) {
            this.saveAnswer(answer);
        } else if (allFieldsTouched) {
            this.saveAnswer(null);
        }
    }
```
satisfied these rules.

Your changes make this question required if the user inputs invalid email or emails are not equal. This question maybe is not required for activity, so it shouldn't affect activity validation. We should only check that the user inputs a valid email or confirms the email correctly(so, if email input form valid - the user answered the question if invalid - the user didn't answer the question, nothing more, mismatched responses or invalid email do not mean that the activity is not valid). Activity validators shouldn't check that the answer contains correct email, the component has already done this.

But okay, let's discuss it on the next week, not on Monday, on Thursday, because we're coming back from vacation on Thursday.

fat23cat: So, since I'm back, let's return to the discussion 🙂
My strong opinion - right now we shouldn't change validation behavior for one block. If we want to do so, we need to discuss new validation architecture in general. 

So, talking about email input type, your changes suggest that if emails mismatched or a user inputted invalid email, the block becomes invalid for activity. But as I mentioned above, it implicitly does question required and it can affect UX when a question shouldn't be required. Email input type logic should be: if email input form valid - the user answered the question if invalid - the user didn't answer the question. The REQUIRED validation rule will control that a user answered on this question.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityTextAnswer.component.ts
MocanaAtBroad: The server rejects NULLs for text questions. Tested it out and saw the backing Java code and had discussion with @yufengwng about this. I think rejecting NULLs is not unreasonable, but we should update the API doc.

fat23cat: Good to know, thanks.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityEmailInput.component.ts
yufengwng: I'm curious why the email check format doesn't match. AFAIK the server is using a very generous format, basically allowing anything that's not whitespace. If Angular uses a more stricter format, then it ought to work?

MocanaAtBroad:
I found that the Angular email format accepts: marco@acme while server
insists on a top-level domain be included, e.g., marco@acme.com .

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityEmailInput.component.ts
MocanaAtBroad: The Angular format allows for emails of the form `marco@acme` . Pepper server insists on having the top-level domain, e.g., `marco@acme.com` .

yufengwng: Okay, that solves the mystery.

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/activityQuestionConverter.service.ts
```diff
@@ -27,7 +28,8 @@ export class ActivityQuestionConverter {
     constructor(
         private validatorBuilder: ActivityValidatorBuilder,
         private suggestionBuilder: ActivitySuggestionBuilder,
-        private logger: LoggingService) {
+        private logger: LoggingService,
+        private translate: NGXTranslateService) {
```
fat23cat: We don't use NGXTranslateService here, please remove.

fat23cat (approved):
Looks and works well, just one remark.

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/activityQuestionConverter.service.ts
MocanaAtBroad: 2eab8c7e28d5a1c10ff66e86cec07318bc7a7581

PR #22, opened 2020-01-07
auth: pass along params to local registration

yufengwng:
While testing osteo age-up locally, I noticed that we're not passing `invitation_id` along via the local registration codepath. This PR attempts to solve that by passing via session storage.

CC @fat23cat

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/login/auth0-code-callback.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/login/auth0-code-callback.component.ts
@@ -32,14 +32,20 @@ export class Auth0CodeCallbackComponent implements OnInit, OnDestroy {
         if (authCode) {
             this.log.logEvent(this.LOG_SOURCE, 'Will login to ' + this.configuration.localRegistrationUrl + ' using code ' + authCode);
 
-            const tempSession = this.session.isTemporarySession() ? this.session.session : null;
+            const value = sessionStorage.getItem('localAuthParams') || "{}";
+            const params: object = JSON.parse(value);
+            sessionStorage.removeItem('localAuthParams');
+
             const registrationPayload = {
                 auth0ClientId: this.configuration.auth0ClientId,
                 studyGuid: this.configuration.studyGuid,
                 auth0Code: authCode,
                 redirectUri: this.configuration.auth0CodeRedirect,
-                ...(tempSession && {
-                    tempUserGuid: tempSession.userGuid
+                ...(params['temp_user_guid'] && {
+                    tempUserGuid: params['temp_user_guid']
+                }),
+                ...(params['invitation_id'] && {
+                    invitationId: params['invitation_id']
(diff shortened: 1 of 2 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/login/auth0-code-callback.component.ts
```diff
@@ -32,14 +32,20 @@ export class Auth0CodeCallbackComponent implements OnInit, OnDestroy {
         if (authCode) {
             this.log.logEvent(this.LOG_SOURCE, 'Will login to ' + this.configuration.localRegistrationUrl + ' using code ' + authCode);
 
-            const tempSession = this.session.isTemporarySession() ? this.session.session : null;
+            const value = sessionStorage.getItem('localAuthParams') || "{}";
```
fat23cat:
```suggestion
            const value = sessionStorage.getItem('localAuthParams') || '{}';
```

fat23cat: And probably it would be better to make work with `sessionStorage` in separate method:
``` ts
private getLocalAuthParams(): object {
   const params = sessionStorage.getItem('localAuthParams') || '{}';
   sessionStorage.removeItem('localAuthParams');
   return params;
}
```

ddp-workspace/projects/ddp-sdk/src/lib/components/login/auth0-code-callback.component.ts
```diff
@@ -32,14 +32,20 @@ export class Auth0CodeCallbackComponent implements OnInit, OnDestroy {
...
+            sessionStorage.removeItem('localAuthParams');
+
             const registrationPayload = {
                 auth0ClientId: this.configuration.auth0ClientId,
                 studyGuid: this.configuration.studyGuid,
                 auth0Code: authCode,
                 redirectUri: this.configuration.auth0CodeRedirect,
-                ...(tempSession && {
-                    tempUserGuid: tempSession.userGuid
+                ...(params['temp_user_guid'] && {
```
zyme, line 41: Nice catch--I had forgotten about local registration!  When we move to app engine, it should be straightforward to spin up stuff from a branch on-the-fly (albeit still sharing GCP databases), which might overall be faster/simpler than our current special case local registration setup.

ddp-workspace/projects/ddp-sdk/src/lib/components/login/auth0-code-callback.component.ts
zyme: How about `resetLocalAuthParams()`?  The side effect of clearing sessionStorage in a getter is a bit unusual.

yufengwng: done

fat23cat (approved):

PR #35, opened 2020-01-28
Toolkit components optimization

fat23cat:
When I started work on osteo, I was creating universal components with legacy design and new design in the Toolkit.
But now, I figured out that it increases the app's bundle size, so, I extracted redesigned templates into separate components and reused TypeScript code. This allowed reducing bundle size by Angular tree-shaking.
Osteo main bundle before optimization:
image
Osteo main bundle after optimization:
image


MocanaAtBroad (approved):
Everything seems to still be working and build smaller, so 👍

PR #37, opened 2020-01-29
Fix flashing error message when saving email entry

MocanaAtBroad:
Two line change. The comment explains the problem and fix.


fat23cat (approved):
Works well!

PR #46, opened 2020-02-07
DDP-4397: Update videos on the osteo home page

fat23cat:
Just added real links on videos

### DIFF
--- a/ddp-workspace/projects/ddp-osteo/src/app/components/welcome/welcome.component.html
+++ b/ddp-workspace/projects/ddp-osteo/src/app/components/welcome/welcome.component.html
@@ -19,7 +19,7 @@
             </div>
             <div class="youtube-video">
                 <iframe [title]="'HomePage.IntroductionSection.IframeTitle' | translate" class="youtube-video__iframe"
-                    src="https://www.youtube.com/embed/VAvMSzBJaqQ" frameborder="0"
+                    src="https://www.youtube.com/embed/d6UcF6ngNKg" frameborder="0"
                     allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture"
                     allowfullscreen></iframe>
             </div>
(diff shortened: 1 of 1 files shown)


zyme (changes requested):
This new alternative approach does look faster and gives us more control, which is nice!

However, we need to stick with youtube hosting here--even with discounts, the financial cost of us hosting and egressing content isn't something we can afford.

fat23cat:
> This new alternative approach does look faster and gives us more control, which is nice!
> 
> However, we need to stick with youtube hosting here--even with discounts, the financial cost of us hosting and egressing content isn't something we can afford.

Okay... How about Google Drive? We can store videos there, we only need to make these videos public, or it also will create additional financial costs?

MocanaAtBroad (approved):
Code looks good and is working correctly for me, so I am approving.
However, one thing that I am not sure we can control (let's investigate options); the initial/static image that is shown when the video has not yet played.
In the top video, the image seems OK to me, but the bottom one looks like the girl got caught in mid-blink; a few frames earlier or later would be OK.
As I understand it, Youtube by default show a the frame that is the very middle of the stream. Maybe there is a way to select a different one?

fat23cat:
> Code looks good and is working correctly for me, so I am approving.
> However, one thing that I am not sure we can control (let's investigate options); the initial/static image that is shown when the video has not yet played.
> In the top video, the image seems OK to me, but the bottom one looks like the girl got caught in mid-blink; a few frames earlier or later would be OK.
> As I understand it, Youtube by default show a the frame that is the very middle of the stream. Maybe there is a way to select a different one?

Youtube doesn't allow to change video's 'poster', we can do it only with `<video>`

MocanaAtBroad:
It appears that the uploader of the video can select one of three frames that YouTube generates to be the "poster". They can also create custom "thumbnails" where they can do whatever they want for the poster. Might be worth letting CountMeIn folks of their options.

PR #51, opened 2020-02-14
DDP-4396: Google Analytics doesn't cover all osteo website

fat23cat:
- GA added for osteo static pages(used the same approach as angio, brain, etc.)
- Updated GA for the activity component(send activity even model doesn't' have a name, send section number if section doesn't have a name, send `closeSurvey` event if an activity is read-only)
- Used enum instead of strings for events
- Noticed that 'PREV' button on the activity a was too compressed, updated CSS


zyme (approved):

zyme:
After you merge this to `develop` and verify, can you cherry pick to `v1.7.2` so we can push a patch to staging?

PR #61, opened 2020-02-27
Images optimization 

fat23cat:
Images optimization by https://squoosh.app/ and converting some .svg -> .png allowed reduce images size:
Angio 3.3 MB -> 0.95 MB
Brain 6.94 -> 2.55 MB
MBC 2.07 MB -> 1.44 MB


MocanaAtBroad (approved):
They looks good! Noticed in a few images _VERY_ subtle differences, but only because I was comparing the old site to new side by side.
Interesting that PNG are really just as good as the SVG. And some of our SVGs were really gigantic bandwith pigs. Anyway, this is very nice!

PR #62, opened 2020-02-28
CircleCI based build of new build/deploy Angular app deployment process on GAE

MocanaAtBroad:
**March 5, 2019 Update:**
I am calling this done. There are some additional things that would be nice to do for which I have created tickets:
https://broadinstitute.atlassian.net/browse/DDP-4479 Remove GAE service versions that are no longer needed
https://broadinstitute.atlassian.net/browse/DDP-4478 Update Angular CircleCi configuration to delete builds from Google buckets that have not been deployed for a while
https://broadinstitute.atlassian.net/browse/DDP-4425 Create mechanism to keep Docker image used in Angular build fresh and in sync with CircleCi build
https://broadinstitute.atlassian.net/browse/DDP-4422 Update CircleCI Angular build configuration so that only project that gets built is one with code changes

Added everyone as reviewer as this PR is of general interest.

The process we are trying to enable:
https://www.lucidchart.com/documents/edit/2f3ef2a3-9bef-477e-9aa3-40188653e6da/0_0?shared=true

The good stuff is in the file .circleci/config.yml .

Tried as best I could to be keep clean, and DRY . CircleCi and YAML itself provide some capabilities to refactor functionality that I used.

Recommend reading the `config.yml` starting from the "workflows:" section towards bottom of the document. That is were all the work that happens gets initiated from.

I configured Google App Engine on GCP and added the necessary credentials to deploy to dev, test, and staging to Vault. Have not messed with prod yet.

The CircleCi pipelines are accessible here:
https://app.circleci.com/github/broadinstitute/ddp-angular/pipelines
If you cannot see them, Zim might have to do some admin magic to allow you to get to the console.


.circleci/config.yml
```diff
@@ -183,11 +395,11 @@ workflows:
...
           name: osteo-nightly
-      - angio:
+      - angio-build-and-deploy:
           name: angio-nightly
-      - brain:
+      - brain-build-and-deploy:
           name: brain-nightly
-      - mbc:
+      - mbc-build-and-deploy:
           name: mbc-nightly
```
MocanaAtBroad, line 426: I left the nightly builds, but I suspect we don't need them. Originally idea was to do --aot builds only at night to make them quicker (I think). Reality is we only save a couple of minutes and saves us pain to find out next day that aot broke the build.

.circleci/config.yml
```diff
@@ -75,106 +66,327 @@ commands:
...
+            echo "Checking for ${TAR_FILE_URL_PATTERN}"
+            # For some reason following line generating exit code 1 every time in CircleCI. Use set +e so script can continue
+            TAR_FILE_URL=`gsutil ls $TAR_FILE_URL_PATTERN  | head -1`
 
+            # if not found, let's kick off a build
+            if [ -z $TAR_FILE_URL ]
+              then
+                echo "No build found for <<parameters.study_key>> with git sha ${SHORT_GIT_SHA}. Initiating a build for it."
+                curl -u "${CIRCLE_API_TOKEN}:" -X POST --header "Content-Type: application/json" -d '{
+                  "branch": "<< pipeline.git.branch >>",
```
MocanaAtBroad: CIRCLE_API_TOKEN injected from the CircleCi configuration. Currently using my personal one (as far as I know there isn't a "service account" type of token).

.circleci/config.yml
```diff
@@ -75,106 +66,327 @@ commands:
...
+                  "parameters": {
+                      "study_key": "<<parameters.study_key>>",
+                      "do-builds": true
+                  }
+                }' https://circleci.com/api/v2/project/gh/broadinstitute/ddp-angular/pipeline
+              else
+                echo "Build for <<parameters.study_key>> found at URL: ${TAR_FILE_URL}. No new build will be initiated"
+            fi
```
MocanaAtBroad, line 265: CIRCLE_API_TOKEN injected from the CircleCi configuration. Currently using my personal one (as far as I know there isn't a "service account" type of token).

MocanaAtBroad: A bit of a disappointment w CircleCi that I needed to call the API to conditionally run individual jobs. Works well, but would be nicer if there was runtime conditionals available at the workflow level.

.circleci/config.yml
```diff
@@ -75,106 +66,327 @@ commands:
...
-            echo $GCLOUD_SERVICE_KEY | gcloud auth activate-service-account --key-file=-
-            # ran out of time, but demonstration of basic idea extracting data from Vault directly. Needs to be wrapped in nicer way!
-            #gcloud --quiet config set compute/zone `vault read  -format=json  secret/pepper/$DEPLOY_ENV/v1/conf | jq -r '.data.gcp.computeZone'`
-      - run:
-          name: <<parameters.study_key>> deploy to GAE
-          command: gcloud app deploy --version="${CIRCLE_SHA1}" --project "broad-ddp-${ENVIRONMENT}" "${ANGULAR_PROJECT_DIR_PATH}/app.yaml"
+            set -u
+            DATE=`date +%F`
+            TAR_NAME="${ANGULAR_PROJECT_NAME}_${DATE}_${SHORT_GIT_SHA}.tar.gz"
+            TAR_PATH="/tmp/${TAR_NAME}"
```
MocanaAtBroad, line 215: Note the format chosen for the tar archive name. Included date because I wanted to a have a clue on when the build occurred (outside of just the timestamp cuz that can get messed up with a copy).
Fully aware that newer build on same day of same project and SHA will clobber earlier one. I am perfectly fine with that, but welcome disagreements.

.circleci/config.yml
```diff
@@ -75,106 +66,327 @@ commands:
...
+    description: "Build DDP study app"
+    steps:
       - run:
-          name: <<parameters.study_key>> ng build
-          command: ng build $ANGULAR_PROJECT_NAME $NG_BUILD_OPTS --base-href=/ --output-path=$ANGULAR_PROJECT_DIR_PATH/dist --verbose=true --progress=true
+          name: ng build
+          command: |
+            set -u
+            ng build $ANGULAR_PROJECT_NAME --prod --aot --base-href=/ --output-path=$OUTPUT_DIR --verbose=true --progress=true
           working_directory: ~/repo/ddp-workspace
```
MocanaAtBroad, line 204: I just force aot (though my understanding is that the `prod` flag picks up the build settings from Angular config file anyways, so might be redundant.

.circleci/config.yml
```diff
@@ -75,106 +66,327 @@ commands:
...
       - run:
-          name: <<parameters.study_key>> npm install
+          name: npm install
           command: npm install
           working_directory: ~/repo/ddp-workspace
       - save_cache:
           key: *npm_cache_key
           paths:
             - ./node_modules
+
```
MocanaAtBroad, line 195: Pavel has indicated correctly that `npm ci` is considered the more reproducible way of setting up the node project initially. My observation was that it was faster doing npm install and since we always start from a clean slate (except for cached node_modules sometimes), I am sticking with `npm install`. 
Happy to revisit this later.

.circleci/config.yml
```diff
@@ -75,106 +66,327 @@ commands:
...
+          command: |
+            echo 'export ANGULAR_PROJECT_NAME=ddp-<<parameters.study_key>>' >> $BASH_ENV
+            echo 'export ANGULAR_PROJECT_DIR_PATH=projects/ddp-<<parameters.study_key>>' >> $BASH_ENV
+            source $BASH_ENV
+            OUTPUT_DIR="${ANGULAR_PROJECT_DIR_PATH}/dist"
+            echo "export OUTPUT_DIR=$OUTPUT_DIR" >> $BASH_ENV
+            echo "export SHORT_GIT_SHA=`echo '<< pipeline.git.revision >>' | cut -c 1-7`" >> $BASH_ENV
+            echo 'export BUILDS_BUCKET=ddp-angular-builds' >> $BASH_ENV
+            source $BASH_ENV
+
```
MocanaAtBroad, line 179: Exhibiting hear some of the weirdnesses with CircleCi: they use heavily environment variables, but make it super awkward to create new ones. Also notice how we mix in global build variables ( pipeline.*) with "local" parameters (parameters.*).
Most of the troubleshooting putting this thing together were errors handling strings and environment variables!

.circleci/config.yml
```diff
@@ -75,106 +66,327 @@ commands:
...
+          command: |
+            set -u
+            ./readvault.sh secret/pepper/${ENVIRONMENT}/v1/conf .data.gcp.serviceKey | gcloud auth activate-service-account --key-file=-
+            gcloud --quiet config set compute/zone ${GOOGLE_COMPUTE_ZONE}
+      - run:
+          name: Deploy to GAE
+          command: |
+            set -u
+            gcloud app deploy --version="${SHORT_GIT_SHA}" --stop-previous-version --project "broad-ddp-${ENVIRONMENT}" "${ANGULAR_PROJECT_DIR_PATH}/app.yaml"
+          working_directory: ~/repo/ddp-workspace
```
MocanaAtBroad, line 144: The "version" strings are available in the Google App Engine service console (I am sure they are available via APIs as well which might be handy at some point to get to).
Also note the --stop-previous-version flag. Without it the all previous deployments are still accessible from a different URL. Really interesting but undesirable.

Screen Shot 2020-02-28 at 6 13 06 PM

.circleci/config.yml
```diff
@@ -75,106 +66,327 @@ commands:
...
-      # todo: conditionally don't build if we already have cached the tarball for the SHA+ENV somewhere (circle artifacts?  github package manager?  GCP bucket?
-      # for better build reproducibility, we want to move the same build artifacts to different environments and not rebuild them and
-      # suffer the pain of npm caching, package-lock, and auto-upgrade pitfalls
+      - run:
+          name: Setup gcloud context
+          command: |
+            set -u
+            ./readvault.sh secret/pepper/${ENVIRONMENT}/v1/conf .data.gcp.serviceKey | gcloud auth activate-service-account --key-file=-
+            gcloud --quiet config set compute/zone ${GOOGLE_COMPUTE_ZONE}
+      - run:
```
MocanaAtBroad, line 139: GOOGLE_COMPUTE_ZONE comes from the CircleCI project configuration. At some point we ought to mix up zones a bit. All deployments are going to same zone.

.circleci/config.yml
```diff
@@ -75,106 +66,327 @@ commands:
...
-          name: <<parameters.study_key>> show environment variables
-          command: env
+          name: Retrieve and expand build archive for << parameters.study_key >> with git SHA << pipeline.git.revision >>
+          command: |
+            set -u
+            # we are getting our builds from the one bucket
+            ./readvault.sh secret/pepper/dev/v1/conf .data.gcp.serviceKey | gcloud auth activate-service-account --key-file=-
+            TAR_FILE_URL_PATTERN="gs://${BUILDS_BUCKET}/${ANGULAR_PROJECT_NAME}/${ANGULAR_PROJECT_NAME}_*_${SHORT_GIT_SHA}.tar.gz"
+            TAR_FILE_URL=`gsutil ls $TAR_FILE_URL_PATTERN  | sort -r | head -1`
+
```
MocanaAtBroad, line 94: Here we look for through the  build archive (if multiple builds of the same SHA pickup the newest one).

.circleci/config.yml
```diff
@@ -75,106 +66,327 @@ commands:
...
               only:
                 - develop
-      - brain:
+      - brain-build-and-deploy:
+          <<: *build-and-deploy-filters
+      - angio-build-and-deploy:
+          <<: *build-and-deploy-filters
+      - mbc-build-and-deploy:
+          <<: *build-and-deploy-filters
+
```
MocanaAtBroad, line 375: Above is the "develop" workflow. If you trace back to each of the jobs (they run in parallel, by the way), you will find that I don't save tar archives for them. Seemed unnecessary, but welcome disagreements.

.circleci/config.yml
```diff
@@ -75,106 +66,327 @@ commands:
...
-      - mbc:
+                - /^rc.*/
+                - /^hotfix.*/
+
+  build-app-workflow:
+    when: << pipeline.parameters.do-builds >>
+    jobs:
+      - app-build-and-store-job:
+          <<: *build-and-store-filters
+
```
MocanaAtBroad, line 393: The two workflows above work together and off the same branches; the first one triggers the second one.

.circleci/config.yml
```diff
@@ -75,106 +66,327 @@ commands:
...
-    # run a build on develop using --prod --aot as an early warning for prod/aot issues
+                - test
+                - staging
+                - production
+      - brain-deploy-only:
+          <<: *deploy-filters
+      - mbc-deploy-only:
+          <<: *deploy-filters
+      - angio-deploy-only:
+          <<: *deploy-filters
```
MocanaAtBroad, line 409: This is the workflow for deployments of saved builds. One thing to notice is that the build-only workflow is detached from the deploy-only workflow. So if:
a) the process of creating the rc_* or hotfix_* branch is not followed.
b) You move PUSH to test or staging a few seconds after having pushed to rc_* or hotfix_*
The deploy will fail because the archive will not exist yet.

.circleci/config.yml
```diff
@@ -75,106 +66,327 @@ commands:
...
+                      "study_key": "<<parameters.study_key>>",
+                      "do-builds": true
+                  }
+                }' https://circleci.com/api/v2/project/gh/broadinstitute/ddp-angular/pipeline
+              else
+                echo "Build for <<parameters.study_key>> found at URL: ${TAR_FILE_URL}. No new build will be initiated"
+            fi
 
 jobs:
-  osteo:
```
MocanaAtBroad, line 118: In circleci, each "job" means (at least by default) launching a build machine/docker image. Below the jobs are just declared. They need to be launched by workflows under defined in the lower section.

ddp-workspace/projects/ddp-angio/app.yaml
```diff
@@ -1,6 +1,11 @@
 service: angio
 runtime: python37
 
+# This setting will not matter except to prevent default automatic-scaling which prevents disabling
+# earlier versions of this service
+basic_scaling:
+  max_instances: 1
+
```
MocanaAtBroad, line 8: As comment indicates, this option allows us to use the `gcloud app deploy --stop-previous-version` flag.

ddp-workspace/readvault.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+if [ "$1" == "" ]  ||  [ "$2" == "" ];
+then
+	echo "USAGE: readvault.sh vaultkey propertypath"
+	echo "EXAMPLE: readvault.sh secret/pepper/dev/v1/conf .data.gcp.serviceKey.token_uri"
+	exit 1
+fi
+vault read -format=json $1 | jq -r $2
```
MocanaAtBroad: Just trying to keep code a little more readable within config.yml . Maybe there are other chunks in the config.yml that can be refactored to external scripts?

findModifiedComponents.sh
```diff
@@ -0,0 +1,22 @@
...
+NG_PROJECT_PATH_PREFIX='ddp-workspace\/projects\/'
+SHARED='_SHARED_'
+SED_CMD='gsed -E'
+git diff --name-only $1 $2 \
+| ${EXCLUDE_CMD} \ `# Exclude from changes files that we want to ignore` \
+| ${SED_CMD} "/${NG_PROJECT_PATH_PREFIX}[^\/]*/! s/.*/${SHARED}/" `# If does not match Angular project path we call it _SHARED_` \
+| ${SED_CMD} "s_${NG_PROJECT_PATH_PREFIX}([^\/]*)/.*_\1_" `# If it does match Angular project path, save only the project dir name` \
+| sort \
+| uniq
+
```
MocanaAtBroad: This script is not yet being used, but keeping it because would like to in future rev of CircleCi configuration.
Current config.yml will build all the projects in development every time a PUSH occurs. This script tries to figure one from using git diff what projects are really affected.
So can eventually be used to avoid unnecessary build/deploys.

readvault.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+if [ "$1" == "" ]  ||  [ "$2" == "" ];
+then
+	echo "USAGE: readvault.sh vaultkey propertypath"
+	echo "EXAMPLE: readvault.sh secret/pepper/dev/v1/conf .data.gcp.serviceKey.token_uri"
+	exit 1
+fi
+vault read -format=json $1 | jq -r $2
```
MocanaAtBroad, line 8: Just trying to keep code a little more readable within config.yml . Maybe there are other chunks in the config.yml that can be refactored to external scripts?

.circleci/config.yml
yufengwng: Maybe we have already explored it and that I missed the discussion: will it be possible to use CircleCI's Artifacts storage instead of buckets?

MocanaAtBroad: It's a good question, Yufeng.
I had assumed that we would use the CircleCi artifact capabilities for handling Angular builds and back-end jars.
On a close reading of the docs and blog posts, though it turns out that their "artifact" support is extremely limited: their configuration API allow for storing the artifacts and not reading them. Furthermore, they warn in fine print:
"artifacts are meant to be useful around the time of the build. It is best practice not to rely on artifacts as a distribution mechanism with long term guarantees".
You can download stuff from their website, but that is pretty much it.
(I am sure there would have been someway to get at the stuff directly using S3 APIs, but did not seem worth the trouble).
It seemed to me that a more realistic option to sticking the builds in our own bucket would have been to use an honest to goodness software artifact repo like artifactory. Seemed like more trouble than what was worth, but certainly this is an option that could be reconsidered at some point.

yufengwng: Those are good considerations. I have not looked at their docs too closely, but it does seems like you can "read" the artifacts by calling their API with `curl`. And you can get the latest artifact too! This seems comparable to downloading from our bucket.

The fine print does sound concerning, but I don't think we need these build artifacts to be long-lived since most of the time we're deploying them within a sprint or two? Unless we _do in fact_ want to keep build artifacts around... If not and we keep with the buckets, maybe we should have a process to occasionally delete old artifacts from buckets (especially dev/test builds since those can accumulate quickly). I just hope we don't end up hand-rolling our own artifact management system.

MocanaAtBroad: You can indeed use `curl` to download files and get the latest, but for some reason they don't provide a way to list them or a configuration option to retrieve a build to use on a deployment that is the counterpart to saving the artifact (something that I found very disconcerting).
The fact that they warn you about not depending on artifacts long term without telling you how long you can indeed count on them (a day, a month, a year?) being there is troubling.
My meta-take away was that the feature was half-cooked and treated as an afterthought.
Replacing it with our own bucket seemed safer and just as easy for what we needed at this time.
Maybe they'll improve the feature or we could jump on Artifactory if our needs grew in complexity. Either ought to be pretty easy to plug into place.
Cleanup can eventually be a concern. We could look at removing builds that are older than month that are not currently deployed or something like that?

yufengwng: Sounds good. I just found out that Google Buckets [has nifty features to automate object deletion](https://cloud.google.com/storage/docs/lifecycle) so that's something we can explore.

readvault.sh
yufengwng: I think there's a convention of putting CI related scripts and files in a top-level directory called `ci`. You might want to consider that convention instead of a `build-utils` directory.

build-utils/sheetappend/index.js
```diff
@@ -0,0 +1,78 @@
...
+            }
+            process.exit(1);
+        } else {
+            console.log("%o", response.data);
+        }
+    };
+    client.request({method: 'POST', url, data}, callback);
+}
+
+ main().catch(console.error);
```
MocanaAtBroad, line 92: Utility to append rows to a Google sheet. We read Google credentials from stdin same way the Google utilities do.

readvault.sh
MocanaAtBroad: Got a repo example? Would like to see what they are doing.
Was considering moving all the scripts and config files, like Dockerfile that have to do with building and deploying off the root into some other place. At first blush referring to all this sort of stuff as ci does not sound very descriptive of intent, but really open to ideas.

yufengwng: Tried finding example, but they're not in the JavaScript world. Some that are seems to just put scripts in the `.circleci` directory: [angular](https://github.com/angular/angular/tree/master/.circleci), [react-native](https://github.com/facebook/react-native/tree/master/.circleci), [electron](https://github.com/electron/electron/tree/master/.circleci). Probably not representative -- couldn't find anything substantial since they either don't use Circle or their `config.yml` is simple.

.circleci/config.yml
```diff
@@ -75,106 +69,343 @@ commands:
...
+            TZ=America/New_York
+            DATE=`TZ=$TZ date +%F`
+            TIME=`TZ=$TZ date +%T`
+            COMPONENT='<<parameters.study_key>> Angular client'
+            BRANCH=<< pipeline.git.branch >>
+            TAG=`git tag --points-at HEAD | xargs`
+            AUTHOR=`git show -s --format='%ae' HEAD`
+            readvault.sh secret/pepper/${ENVIRONMENT}/v1/conf .data.gcp.serviceKey | \
+            sheetappend $RELEASE_SHEET_ID $DATE $TIME "$COMPONENT" $ENVIRONMENT $BRANCH $TAG $SHORT_GIT_SHA "$AUTHOR" $CIRCLE_BUILD_NUM "$CIRCLE_BUILD_URL"
+
```
MocanaAtBroad, line 160: Record data in newly created deployments sheet. The rows with the URLs reflect what we are current writing to sheet (Note that now we add data rows to top of sheet, right below the header):
Screen Shot 2020-03-05 at 2 27 40 PM

readvault.sh
MocanaAtBroad: Good examples! I am inclined to leave things as they are for now, but certainly open to change things as we learn more and/or our needs expand.

.circleci/config.yml
MocanaAtBroad: I'll create a ticket to add cleanup. I think our needs will be a little more complicated than just letting GCP remove old files. We probably will want to query the GCP to find out  what is currently running there and then remove builds that have not run in a while (where "a while" is TBD).

zyme (approved):
Woo hoo!

zyme (approved):

PR #63, opened 2020-03-02

fat23cat:
Bug description and steps of reproduction are here: https://broadinstitute.atlassian.net/browse/DDP-4304
Actually, I could reproduce the bug only once, but my impression is that this is occurring by last-second click on the Continue button and slow Internet connection. 
My idea - give extra time in case a user will interact with Sign Out and Continue buttons.
Presently, the session expires at the same time with the counter in the SessionWillExpire dialog, I think it may сause the bug when a user interacts with Sign Out and Continue buttons and makes a last-second click. So I've added "extra time", the dialog will show that a user has only 1 second left, but the session will be alive 1 minute 1 second that allows interacting with the website when a user clicks Continue and renews token even with slow internet.

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/services/sessionMemento.service.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/services/sessionMemento.service.ts
@@ -8,7 +8,7 @@ import { filter, map, mergeMap, startWith } from 'rxjs/operators';
 export class SessionMementoService implements OnDestroy {
     private readonly SESSION_KEY: string = 'session_key';
     private readonly TOKEN_KEY = 'token';
-    private readonly MSECS_TO_NOTIFICATION = 300000;
+    private readonly MSECS_TO_NOTIFICATION = 360000;
 
     private sessionSubject: BehaviorSubject<Session | null> = new BehaviorSubject<Session | null>(this.session);
     private renewSubject: Subject<number> = new Subject<number>();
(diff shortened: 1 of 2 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/services/sessionMemento.service.ts
```diff
@@ -8,7 +8,7 @@ import { filter, map, mergeMap, startWith } from 'rxjs/operators';
 export class SessionMementoService implements OnDestroy {
     private readonly SESSION_KEY: string = 'session_key';
     private readonly TOKEN_KEY = 'token';
-    private readonly MSECS_TO_NOTIFICATION = 300000;
+    private readonly MSECS_TO_NOTIFICATION = 360000;
```
zyme, line 11: Would it help to add more "leeway" to the [token expiration server-side](https://www.javadoc.io/doc/com.auth0/java-jwt/3.0.1/com/auth0/jwt/JWTVerifier.Verification.html#acceptLeeway-long-)?  I think we're currently at 10 seconds.

MocanaAtBroad (approved):

ddp-workspace/projects/ddp-sdk/src/lib/services/sessionMemento.service.ts
fat23cat: I think there should be enough changes only on the frontend side.

PR #67, opened 2020-03-10

fat23cat:


ddp-workspace/projects/toolkit/src/lib/components/dialogs/sessionWillExpire.component.ts
```diff
@@ -56,7 +56,7 @@ export class SessionWillExpireComponent implements OnInit, OnDestroy {
                 this.timeLeft = `${formattedMinutes}:${formattedSeconds}`;
             } else {
                 this.timeLeft = '00:00';
-                this.auth0.handleExpiredSession();
+                this.auth0.logout('session-expired');
```
fat23cat, line 59: If the time is up, log out and redirect a user to `/session-expired`

ddp-workspace/projects/ddp-sdk/src/lib/interceptors/auth-interceptor.service.ts
```diff
@@ -0,0 +1,20 @@
...
+        private auth0: Auth0AdapterService) { }
+
+    public intercept(req: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
+        if (this.session.isSessionExpired()) {
+            this.auth0.logout('session-expired');
+        } else {
+            return next.handle(req);
+        }
+    }
+}
```
fat23cat, line 20: Angular HttpInterceptors allow intercepting all HTTP requests, so we can check that the session is alive and prevent sending invalid requests. Unauthenticated requests will work as before.

ddp-workspace/projects/ddp-sdk/src/lib/services/sessionMemento.service.ts
```diff
@@ -159,11 +159,7 @@ export class SessionMementoService implements OnDestroy {
...
     }
 
-    private hasOnlyUserGuid(session: Session): boolean {
-        return !session.idToken && !!session.userGuid;
-    }
-
-    private isSessionExpired(): boolean {
+    public isSessionExpired(): boolean {
         const session = this.sessionSubject.value;
         if (session === null) {
```
zyme, line 164: I think the brackets aren't balanced here, resulting in this always returning `false`?  Or am I misreading?

ddp-workspace/projects/ddp-sdk/src/lib/interceptors/auth-interceptor.service.ts
zyme: Will this only pick up on traffic back to pepper backend?  Or would it also run for calls to other URLs?

zyme (commented):
I like the approach of one interceptor to rule them all.  Overall looks good, but a few questions inline.

ddp-workspace/projects/ddp-sdk/src/lib/interceptors/auth-interceptor.service.ts
fat23cat: This will pick up on traffic only back to pepper backend.

ddp-workspace/projects/ddp-sdk/src/lib/services/sessionMemento.service.ts
fat23cat: Hm, I think brackets are balanced.
If the session observable is alive we will skip `return false;` and jump to `return new Date().getTime() > session.expiresAt;`

fat23cat: Probably GitHub cut some lines in the code review tab and confused you... this method:
``` ts
    public isSessionExpired(): boolean {
        const session = this.sessionSubject.value;
        if (session === null) {
            return false;
        }

        return new Date().getTime() > session.expiresAt;
    }
```

zyme (approved):

PR #68, opened 2020-03-11
DDP-4432: Pass user's language selection to Auth0 registration

fat23cat:


ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
```diff
@@ -113,7 +115,8 @@ export class Auth0AdapterService implements OnDestroy {
             }),
             ...(additionalParams && {
                 ...additionalParams
-            })
+            }),
+            language: this.language.getCurrentLanguage()
```
fat23cat, line 119: @yufengwng, have we got the final decision about the parameter's name?

yufengwng: I think I like just calling it `language`. I feel `selected_language` might be too specific (imagine if in the future a client automatically detects the language instead of having a user "select" it).

ddp-workspace/projects/ddp-sdk/src/lib/services/languageService.service.ts
```diff
@@ -0,0 +1,11 @@
...
+import { TranslateService } from '@ngx-translate/core';
+
+@Injectable()
+export class LanguageService {
+    constructor(private translate: TranslateService) { }
+
+    public getCurrentLanguage(): string {
+        return this.translate.currentLang;
+    }
+}
```
fat23cat, line 11: I decided to create `LanguageService` because in the future we will need to manage/store chosen language in the LocalStorage when we will create LanguageComponent, so it just the initial step in this direction.

zyme (dismissed):

zyme (approved):

PR #70, opened 2020-03-11
Fix handling of hotfix branches on CircleCi

MocanaAtBroad:
Merged into branch angular-appengine-deployment-ctd-2 and corresponding PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,6 +51,9 @@ commands:
               develop)
                 DEPLOY_ENV=dev
               ;;
+              circle-ci-update-test)
+                DEPLOY_ENV=dev
+              ;;
               circleci-update-build-bucket)
                 DEPLOY_ENV=dev
               ;;
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,7 +93,7 @@ commands:
           command: |
             set -u
             # we are getting our builds from the one bucket
-            readvault.sh secret/pepper/${ENVIRONMENT}/conf .data.gcp.serviceKey | gcloud auth activate-service-account --key-file=-
+            readvault.sh secret/pepper/${ENVIRONMENT}/v1/conf .data.gcp.serviceKey | gcloud auth activate-service-account --key-file=-
             TAR_FILE_URL_PATTERN="gs://${BUILDS_BUCKET}/${ANGULAR_PROJECT_NAME}/${ANGULAR_PROJECT_NAME}_*_${SHORT_GIT_SHA}.tar.gz"
             TAR_FILE_URL=`gsutil ls $TAR_FILE_URL_PATTERN  | sort -r | head -1`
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -21,7 +21,6 @@ commands:
       - setup-build-workspace
       - setup-shared-env:
           study_key: << parameters.study_key >>
-      - set-deployment-environment
       - build-app
```
MocanaAtBroad, line 24: When we build and store (like when we create a hotfix_* or rc_* branch), we might not have a branch that we can map to an environment, so this will fail.

.circleci/config.yml
```diff
@@ -224,7 +223,7 @@ commands:
           name: Store build archive
           command: |
             set -u
-            readvault.sh secret/pepper/${ENVIRONMENT}/v1/conf .data.gcp.serviceKey | gcloud auth activate-service-account --key-file=-
+            readvault.sh secret/pepper/dev/v1/conf .data.gcp.serviceKey | gcloud auth activate-service-account --key-file=-
             gsutil cp  ${TAR_PATH} "gs://${BUILDS_BUCKET}/${ANGULAR_PROJECT_NAME}/${TAR_NAME}"
```
MocanaAtBroad, line 227: And since there is no ENVIRONMENT , then we just used a fixed one.

.circleci/config.yml
```diff
@@ -236,14 +235,13 @@ commands:
...
-      - set-deployment-environment
       - run:
           name: Initiate build for <<parameters.study_key>> and git SHA << pipeline.git.revision >> if existing one not found
           command: |
             set -u
             set +e
             # Looking for latest build of this project in Google bucket
-            readvault.sh secret/pepper/${ENVIRONMENT}/v1/conf .data.gcp.serviceKey | gcloud auth activate-service-account --key-file=-
+            readvault.sh secret/pepper/dev/v1/conf .data.gcp.serviceKey | gcloud auth activate-service-account --key-file=-
             TAR_FILE_URL_PATTERN="gs://${BUILDS_BUCKET}/${ANGULAR_PROJECT_NAME}/${ANGULAR_PROJECT_NAME}_*_${SHORT_GIT_SHA}.tar.gz"
```
MocanaAtBroad, line 245: Same thing here when we launch the build and stores.

.circleci/config.yml
```diff
@@ -224,7 +223,7 @@ commands:
           name: Store build archive
           command: |
             set -u
-            readvault.sh secret/pepper/${ENVIRONMENT}/v1/conf .data.gcp.serviceKey | gcloud auth activate-service-account --key-file=-
+            readvault.sh secret/pepper/dev/v1/conf .data.gcp.serviceKey | gcloud auth activate-service-account --key-file=-
```
zyme: As per our chat, the plan here is to switch this to `prod` whenever we build code, since the location of the artifact will always be the production bucket.

zyme (approved):

PR #71, opened 2020-03-13
DDP-3957: Google Analytics service refactoring

fat23cat:
Details and implementation sketch are here: https://broadinstitute.atlassian.net/browse/DDP-3957


ddp-workspace/projects/toolkit/src/lib/components/app/app.component.ts
```diff
@@ -228,4 +242,12 @@ export class AppComponent implements OnInit, AfterViewInit, OnDestroy {
...
         }
     }
+
+    private initAnalyticsListener(): void {
+        const events = this.analytics.analyticEvents.subscribe((event: AnalyticsEvent) => {
+            ga('send', event);
+            ga('platform.send', event);
+        });
+        this.anchor.addNew(events);
+    }
```
MocanaAtBroad: Would it be possible to move this subscription to `app.module.ts` ? I think of app.module as the central place where the application is configured and this subscription is pretty central and important; makes it more visible to anyone figuring out how we setup/handle GA.

ddp-workspace/projects/toolkit/src/lib/components/app/app.component.ts
```diff
@@ -9,15 +9,27 @@ import { WarningComponent } from '../dialogs/warning.component';
...
+    AnalyticsEvent
+} from 'ddp-sdk';
 
 declare global {
     interface Window {
         twttr: any;
     }
 }
 
+declare const ga: Function;
```
MocanaAtBroad: Would be nice if we had @types for Google Analytics .

ddp-workspace/projects/ddp-angio/src/app/app.module.ts
```diff
@@ -134,7 +134,7 @@ export function translateFactory(translate: TranslateService, injector: Injector
 export class AppModule {
   constructor(
     private router: Router,
-    private analytics: GoogleAnalyticsEventsService) {
+    private analytics: AnalyticsEventsService) {
     this.router.events.subscribe(event => {
       if (event instanceof NavigationEnd) {
         this.analytics.emitNavigationEvent();
```
MocanaAtBroad: Is it possible to factor this into toolkit? Guess we would need access to router?

ddp-workspace/projects/ddp-sdk/src/lib/models/analyticsActionTypes.ts
```diff
@@ -1,4 +1,4 @@
-export enum GoogleAnalytics {
+export enum AnalyticsActionTypes {
```
MocanaAtBroad: We should consider removing `clickedCountMeIn` from here. This is more of a toolkit thing.

ddp-workspace/projects/ddp-sdk/src/lib/services/analyticsEvents.service.ts
```diff
@@ -1,45 +1,46 @@
...
+import { AnalyticsEvent } from '../models/analyticsEvent';
 
 @Injectable()
-export class GoogleAnalyticsEventsService {
+export class AnalyticsEventsService {
+  private events = new Subject<AnalyticsEvent>();
+
   public emitCustomEvent(
     eventCategory: string,
     eventAction: string,
```
MocanaAtBroad, line 11: Noticing that in the great majority of the calls to this method we specify for both the `eventCategory` argument and the `eventAction` argument an enum value from `AnalyticsActionTypes`  suggesting that semantically something is probably incorrect here. Either the enums are not really "action types" or the argument names are misleading. Could you take a look at Google Analytics API to see how these argument align with what gets reported back to Google Analytics?

ddp-workspace/projects/ddp-angio/src/app/app.module.ts
fat23cat: Ok, I’ll try and see how it will work in the Toolkit.

ddp-workspace/projects/ddp-sdk/src/lib/models/analyticsActionTypes.ts
fat23cat: Why? I think ROPES rely on the events which defined in the enum.

ddp-workspace/projects/toolkit/src/lib/components/app/app.component.ts
fat23cat: Tried to use types, but didn't find appropriate types in the library https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/google.analytics/index.d.ts

ddp-workspace/projects/ddp-sdk/src/lib/services/analyticsEvents.service.ts
fat23cat: Function's arguments match Google Analytics API, but okay, I will refactor enum to make it more semantically correct.

ddp-workspace/projects/ddp-sdk/src/lib/models/analyticsActionTypes.ts
MocanaAtBroad: Maybe CountMeIn is a CountMeInAnalyticsType toolkit? API calls just take strings so that would work. Making observation that everything else there is fairly generic and could be used by any other non-CMI project. This is the one exception.
Not a huge deal. Just a string, but mindset ought to be that SDK knows nothing about CMI.
Don't get hung up on this. Just make best decision. This is an observation, not a direction tor change.

ddp-workspace/projects/ddp-sdk/src/lib/services/analyticsEvents.service.ts
MocanaAtBroad: Same comment with this one: this is an observation, not a direction to change. If you can come up with a simple solution to refactor to make it cleaner awesome.

ddp-workspace/projects/ddp-sdk/src/lib/models/analyticsActionTypes.ts
fat23cat: I think we need to leave it as is. SDK doesn't emit any events with this value, but I'd prefer to store all analytics constants in one place.

ddp-workspace/projects/ddp-sdk/src/lib/services/analyticsEvents.service.ts
fat23cat: Done :)

ddp-workspace/projects/toolkit/src/lib/components/app/app.component.ts
MocanaAtBroad: I am sorry because I think my comment here was unclear/misleading. What I was hoping was that this could go in the `toolkit.module.ts` (not in in app.module.ts. Toolkit does not an app.module.ts). Intention is that the wiring up the GA would happen in in the *.module.ts instead of component. 
I think it that what you did, moving it to each app's app.module.ts is better than component, but with they all have the toolkit dependency, there is no reason for duplication.
Again, sorry. I was not clear in my original comment.

ddp-workspace/projects/ddp-sdk/src/lib/models/analyticsEventCategories.ts
```diff
@@ -0,0 +1,9 @@
+export enum AnalyticsEventCategories {
+    ClickedCountMeIn = 'clickedCountMeIn',
+    Social = 'social',
+    CloseSurvey = 'closeSurvey',
+    SubmitSurvey = 'submitSurvey',
+    Dashboard = 'dashboard',
+    Authentication = 'authentication',
+    MailingList = 'mailingList'
+}
```
MocanaAtBroad, line 9: Do we have room for modifying these strings for categories and actions?
I think the separation of actions and categories is a good idea, but the strings underneath don't quite seem to match the definition of a category or action: e.g., closeSurvey seems more like an action than a category. ClickedCountMeIn also sounds more like an action (both are verbs). 'fromMainPage' sounds more like a source than an action.
If these are the strings we are stuck using, then maybe these are just tags (primary and secondary).

fat23cat: You are probably right that this is not semantically, but I split values in enums not by name, but by how they are used in GA(these values were determined for angio gen2 website)

MocanaAtBroad: It is what it is!

MocanaAtBroad (approved):

PR #73, opened 2020-03-18
Hotfix 1.7.5 circleci updates

MocanaAtBroad:


zyme (approved):

.circleci/config.yml
```diff
@@ -173,14 +174,15 @@ commands:
       - run:
           name: Set environment variables to build <<parameters.study_key>>
           command: |
+            echo "export VAULT_TOKEN=`vault write -field=token auth/approle/login role_id=$VAULT_ROLE_ID secret_id=$VAULT_ROLE_SECRET_ID`" >> $BASH_ENV
```
yufengwng, line 177: This is _**writing**_ to some `auth/approle/login` secret in vault and that somehow produces a vault token that we're storing in bash's env. Am I reading this right??

yufengwng: I'm assuming `auth/...` is some other secrets engine that devops has setup and somehow produces a login token, which we're outputting using the `-field` option?

MocanaAtBroad: This is a Vault feature:
https://www.vaultproject.io/docs/auth/approle/
Without the `-field` option you get a whole list of keys and values.

MocanaAtBroad: And you are reading correctly. We are creating a script line that looks something like:
export VAULT_TOKEN=xxxxxx
that gets appended to BASH_ENV
This is how the CircleCI docs explain how to add env variables that are used when executing other commands.

yufengwng: Interesting, good to know!

yufengwng (approved):
👍

PR #78, opened 2020-04-08
Script to trigger build/deploy to dev and other refactors and improvements

MocanaAtBroad:
Added script for situations where we need to deploy to dev from own branch and want to avoid hacking config.yml script further.
**Updates on April 9, 2020**
- Some refactoring of the CircleCi configuration. Objective was to remove need to keep passing around study guids (we should know what they are) and to remove repetitive volume.
- Script for remote build deploys no longer needs the study guid to be specified.
- Updated Docker file to be based on Alpine . It is __**MUCH**__ faster! Loading the image from CircleCi went down from 50 to 80 seconds down to 10 to 15 sec.
Image size went from 760MB to 160MB. Whoa!

Comments added throughout.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,17 +3,11 @@ version: 2.1
 references:
   npm_cache_key: &npm_cache_key
                    v1-dependency-npm-{{ checksum "package-lock.json" }}
+  # Note: important that values of guids and keys have matching order!"
   study_keys: &study_keys
                 "osteo brain angio mbc"
   study_guids: &study_guids
                 "CMI-OSTEO cmi-brain ANGIO cmi-mbc"
-#  deploy_branch_filters: &deploy_branch_filters
-#    filters:
-#      branches:
-#        only:
-#          - test
-#          - staging
-#          - production
 
 # Container for all the builds
 executors:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -477,6 +477,7 @@ workflows:
             branches:
               only:
                 - develop
+                - trigger-build-deploy
 
   build-and-deploy-workflow:
     when: << pipeline.parameters.do-builds >>
@@ -525,7 +526,6 @@ workflows:
             branches:
               only:
                 - develop
-                - trigger-build-deploy
     jobs:
       - osteo-build-and-deploy:
           name: osteo-nightly
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -401,6 +401,16 @@ jobs:
       - app-build-and-store:
           study_key: << pipeline.parameters.study_key >>
 
+  deploy-stored-build-job:
+    parameters:
+      study_key:
+        type: string
+    executor:
+      name: commit-executor
+    steps:
+      - deploy-stored-build:
+          study_key: <<parameter.study_key>>
+
   osteo-deploy-only:
     executor:
       name: commit-executor
@@ -506,7 +516,7 @@ workflows:
 
   deploy-all-apps-workflow:
     jobs:
-      - deploy-stored-build: &deploy_branch_filters
+      - deploy-stored-build-job: &deploy_branch_filters
           filters:
             branches:
               only:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -525,6 +525,7 @@ workflows:
             branches:
               only:
                 - develop
+                - trigger-build-deploy
     jobs:
       - osteo-build-and-deploy:
           name: osteo-nightly
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,6 +33,9 @@ commands:
     parameters:
       study_key:
         type: string
+      study_guid:
+        type: string
+        default: NOTNEEDED
       skip_build:
         type: boolean
         default: false
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -226,7 +226,9 @@ commands:
       - checkout:
           path: ~/repo
       - restore_cache:
-          key: *npm_cache_key
+          keys:
+            - *npm_cache_key
+            - v1-dependency-npm-
       - run:
           name: npm install
           command: npm install
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -402,11 +402,14 @@ jobs:
           study_key: << pipeline.parameters.study_key >>
 
   deploy-stored-build-job:
+    parameters:
+      study_key:
+        type: string
     executor:
       name: commit-executor
     steps:
       - deploy-stored-build:
-          study_key: << pipeline.parameters.study_key >>
+          study_key: << parameters.study_key >>
 
   osteo-deploy-only:
     executor:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -158,7 +158,7 @@ commands:
           name: Update GAE dispatch config
           command: |
             set -u
-            gcloud app deploy --project "broad-ddp-${ENVIRONMENT}" "${ANGULAR_PROJECT_DIR_PATH}/output-config/dispatch.yaml"
+            gcloud app deploy --quiet --project "broad-ddp-${ENVIRONMENT}" "${ANGULAR_PROJECT_DIR_PATH}/output-config/dispatch.yaml"
```
MocanaAtBroad, line 150: Does not appear that dispatch rule was being applied.

fat23cat:
Am I right that with this script we just need to add new `study_key`, `study_guid` and executors in circleci config without specifying local branch to deploy new app into GAE?

MocanaAtBroad:
> Am I right that with this script we just need to add new `study_key`, `study_guid` and executors in circleci config without specifying local branch to deploy new app into GAE?

No. You execute the script locally/from your computer. Will need to generate a CircleCi token as described, but that is it.

fat23cat:
@MocanaAtBroad, could you provide example of how to run the script?

fat23cat (dismissed):

.circleci/config.yml
```diff
@@ -3,12 +3,17 @@ version: 2.1
 references:
   npm_cache_key: &npm_cache_key
                    v1-dependency-npm-{{ checksum "package-lock.json" }}
+  # Note: important that values of guids and keys have matching order!"
+  study_keys: &study_keys
+                "osteo brain angio mbc"
+  study_guids: &study_guids
+                "CMI-OSTEO cmi-brain ANGIO cmi-mbc"
```
MocanaAtBroad, line 11: Place this here so can be referenced using YAML keys. This is now the only place where Guids show up in script.

.circleci/config.yml
```diff
@@ -3,12 +3,17 @@ version: 2.1
...
 
 # Container for all the builds
 executors:
   commit-executor:
     docker:
-      - image: broadinstitute/study-website-build:03-26-2020A
+      - image: broadinstitute/study-website-build:04-09-2020A
```
MocanaAtBroad, line 16: The updated Docker image.

.circleci/config.yml
```diff
@@ -87,8 +88,6 @@ commands:
     parameters:
       study_key:
         type: string
-      study_guid:
-        type: string
     steps:
```
MocanaAtBroad, line 84: Can go away!

.circleci/config.yml
```diff
@@ -117,22 +116,19 @@ commands:
...
-          study_guid: << parameters.study_guid >>
 
   deploy-exploded-build:
     description: Deploy files to GAE
     parameters:
       study_key:
         type: string
-      study_guid:
-        type: string
     steps:
```
MocanaAtBroad, line 118: Ditto.

.circleci/config.yml
```diff
@@ -117,22 +116,19 @@ commands:
...
       - set-deployment-environment
       - run:
           name: <<parameters.study_key>> generate config files and copy pepperConfig.js
           command: |
             set -u
-            ./build-study.sh v1 $ENVIRONMENT . <<parameters.study_key>> <<parameters.study_guid>> --config
+            ./build-study.sh v1 $ENVIRONMENT . <<parameters.study_key>> "$STUDY_GUID" --config
             cd ddp-workspace
```
MocanaAtBroad, line 125: They can go away because we are figuring out STUDY_GUID up front and sticking in in an env variable.

.circleci/config.yml
```diff
@@ -193,13 +189,34 @@ commands:
...
+              STUDY_KEYS=(<<parameters.study_keys>>)
+              STUDY_GUIDS=(<<parameters.study_guids>>)
+              for i in ${!STUDY_KEYS[@]}; do
+                  currentKey=${STUDY_KEYS[$i]}
+                  if [[ $currentKey == '<<parameters.study_key>>' ]]; then
+                      STUDY_GUID="${STUDY_GUIDS[$i]}"
+                      echo "export STUDY_GUID=${STUDY_GUID}" >> $BASH_ENV
+                      break;
+                  fi
+              done;
```
MocanaAtBroad, line 208: This is where we figure out the STUDY_GUID. Bash makes it a bit of a pain to work with arrays.

.circleci/config.yml
```diff
@@ -278,15 +293,13 @@ commands:
             set +x
             # CircleCi does not support arrays as params, so here is my workaround
             STUDY_KEYS=(<< parameters.study_keys >>)
-            STUDY_GUIDS=(<< parameters.study_guids >>)
```
MocanaAtBroad, line 291: Not needed!

.circleci/config.yml
```diff
@@ -297,7 +310,6 @@ commands:
                                   \"branch\": \"<< pipeline.git.branch >>\",
                                   \"parameters\": {
                                       \"study_key\": \"$STUDY_KEY\",
-                                      \"study_guid\": \"$STUDY_GUID\",
                                       \"do-builds\": true
```
MocanaAtBroad, line 308: Don't need to pass the GUID to call. Config will figure it out based on key.

.circleci/config.yml
```diff
@@ -382,65 +394,15 @@ jobs:
...
-          <<: *angio-params
-
-  brain-build-and-deploy:
-    executor:
-      name: commit-executor
-    steps:
-      - app-build-and-deploy:
-          <<: *brain-params
-
-  mbc-build-and-deploy:
```
MocanaAtBroad, line 438: After refactor the specialized jobs that only differ by what key and guid is used can be removed.

.circleci/config.yml
```diff
@@ -382,65 +394,15 @@ jobs:
...
+    parameters:
+      study_key:
+        type: string
     executor:
       name: commit-executor
     steps:
-      - app-build-and-deploy:
-          <<: *mbc-params
+      - deploy-stored-build:
+          study_key: << parameters.study_key >>
```
MocanaAtBroad, line 401: The more generic job. All that we need.

build-utils/Dockerfile-BuildDeploy
```diff
@@ -1,63 +1,120 @@
 # To make our config generation "compatible" with official dsdetoolbox, extract what is needed from it (this is not the base image)
 FROM broadinstitute/dsde-toolbox:consul-template-20-0.0.1 AS dsdetoolbox
 
-FROM google/cloud-sdk:279.0.0
+FROM google/cloud-sdk:alpine
+
```
MocanaAtBroad, line 5: The key to the smaller Docker image.

build-utils/findmodifiedprojects.sh
```diff
@@ -11,10 +11,10 @@ do
 done
 NG_PROJECT_PATH_PREFIX='ddp-workspace\/projects\/'
 SHARED='_SHARED_'
-if [[ "$OSTYPE" == "linux-gnu" ]]; then
-  SED_CMD='sed -E'
```
MocanaAtBroad, line 15: OSTYPE not set in Alpine Linux. No matter.

fat23cat (approved):

PR #83, opened 2020-04-13
DDP-4536: Textfield disappears if name data is entered but not year data

fat23cat:


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityCompositeAnswer.component.ts
```diff
@@ -118,6 +116,7 @@ export class ActivityCompositeAnswer implements OnChanges {
         newQuestionBlock.validators.forEach((validator) => validator.question = newQuestionBlock);
         newQuestionBlock.answer = answerContainer === null ? null : answerContainer.value;
         newQuestionBlock.shown = shown;
+        newQuestionBlock.question = '';
```
fat23cat, line 119: These changes related to this bug:
image
On the frontend, we shouldn't show 'Date' question, but the mobile client requires that field.
I see two solutions:
1. I looked through all composite blocks on our websites and look like composite blocks show only the main question, sub-questions don't show his own question, so we can just clear question text(like in my PR)
2. We can add `hidePrompt` field in the question to control it

What do you think @zyme, @MocanaAtBroad, @yufengwng?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityCompositeAnswer.component.ts
```diff
@@ -109,11 +120,14 @@ export class ActivityCompositeAnswer implements OnChanges {
...
             newQuestionBlock.placeholder = newQuestionBlock.question;
-            newQuestionBlock.question = null;
+            newQuestionBlock.question = '';
             if (childQuestionBlock.isRequired) {
                 newQuestionBlock.placeholder += ' *';
             }
         }
+        if (this.block.childOrientation === ChildOrientation.Horizontal) {
+            newQuestionBlock.question = '';
+        }
```
MocanaAtBroad, line 134: This looks correct. Don't think we ever want to show the question prompt in horizontal.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityCompositeAnswer.component.ts
```diff
@@ -84,14 +84,25 @@ export class ActivityCompositeAnswer implements OnChanges {
...
+                    const blankRow: ActivityQuestionBlock<any>[] = this.block.children.map((questionBlock: ActivityQuestionBlock<any>) =>
+                        this.buildBlockForChildQuestion(questionBlock, null, this.block.shown));
+
+                    this.childQuestionBlocks = questionsRows.map((currentRow: ActivityQuestionBlock<any>[]) => {
+                        if (currentRow.length !== blankRow.length) {
+                            const missedQuestions = blankRow.slice(currentRow.length);
+                            currentRow.push(...missedQuestions);
+                        }
+                        return currentRow;
+                    });
```
MocanaAtBroad, line 109: Please add some comments here. Not sure what you are shooting for here.

MocanaAtBroad (commented):
On the layout: just wanted to make sure that we did not break changes that we made a while ago to support mbc stuff.
Specifically, our odd duck question is MBC, about you question 6:
Screen Shot 2020-04-13 at 1 31 42 PM
The is the PR (from old repo) that enabled this:
https://github.com/broadinstitute/ddp/pull/1078
These new changes do not appear to interfere with it so that is good.
Please add some comments are what you are trying to achieve with the blank rows.

fat23cat:
@MocanaAtBroad, I've added a comment and checked MBC composite question - looks good!

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityCompositeAnswer.component.ts
MocanaAtBroad: OK. I see the problem you are dealing with now. Thanks for the comments.
Consider updating the newAnswers.map block to deal with the blank row. Might simplify things:
e.g.:
```typescript
   const questionsRows: ActivityQuestionBlock<any>[][] = newAnswers.map((rowOfAnswers: ActivityQuestionBlock<any>[]) => {
                        return this.block.children.map((questionBlock, index) => {
                            const answerValue = index < rowOfAnswers.length ? rowOfAnswers[index] : null;
                            return this.buildBlockForChildQuestion(questionBlock, answerValue , questionBlock.shown);
                        });
                    });

I don't understand the use case as well as you do, so do check if this indeed deals with this situation!

yufengwng:
Tried it locally and fix looks good to me!
Screen Shot 2020-04-13 at 3 17 55 PM

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityCompositeAnswer.component.ts
fat23cat: I tried to use your implementation, but it works differently. Can we leave it as-is?

MocanaAtBroad (approved):

PR #88, opened 2020-04-15

fat23cat:


ddp-workspace/projects/toolkit/src/lib/components/dialogs/sessionWillExpire.component.ts
```diff
@@ -68,13 +68,11 @@ export class SessionWillExpireComponent implements OnInit, OnDestroy {
...
     public signOut(): void {
         this.renewNotifier.hideSessionExpirationNotifications();
-        this.session.disableTokenExpiredProcess();
         this.auth0.logout();
     }
 
     public renewSession(): void {
         this.renewNotifier.hideSessionExpirationNotifications();
-        this.session.disableTokenExpiredProcess();
         this.auth0.auth0RenewToken();
```
fat23cat: Perhaps this caused an error... since we have additional time, we don't need to disableTokenExpiredProcess.
But honestly, I couldn't reproduce an error, I spent hours trying to reproduce locally and on osteo dev. Moreover, it really weird that only osteo affected by this issue, because angio, brain, mbc and osteo use the same component!
Do we have some special config for auth0/backend for the app which configures the possibility to renew the session? The apps use this method https://github.com/broadinstitute/ddp-angular/blob/develop/ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts#L203
cc @yufengwng, @zyme, @MocanaAtBroad

ddp-workspace/projects/toolkit/src/lib/components/dialogs/sessionWillExpire.component.ts
```diff
@@ -68,13 +72,28 @@ export class SessionWillExpireComponent implements OnInit, OnDestroy {
...
-        this.session.disableTokenExpiredProcess();
         this.auth0.logout();
     }
 
     public renewSession(): void {
-        this.renewNotifier.hideSessionExpirationNotifications();
-        this.session.disableTokenExpiredProcess();
+        // Blocks UI to prevent interference in the session renewing process
+        this.isRenewing = true;
+        this.dialogRef.disableClose = true;
```
yufengwng: Nice! I think I noticed one time that I was able to dismiss the dialog. Glad to see you have added the safeguard here!

MocanaAtBroad (commented):
Could you add some text explaining your understanding of what the problem was and your approach to fixing it.
Sounds like you believe the UI was blocking the renewal of the token? Because this is a modal dialog?
Could you also explain the introduction of the participantGuid as well. I see that this was an instance variable but was not been set before. Was this also causing a problem?

yufengwng:
@MocanaAtBroad I can add some context:

Kiara and I noticed an issue with token renewal in Osteo (and ONLY Osteo). We narrowed it down to the parental consent workflows. Token renewal is fine for self-consent workflow. While debugging, I noticed that token renewal is successful, but after that we're using the wrong user guid when talking to the API (we're using the parent guid instead of the child guid). I believe we already have this `participantGuid` thing in the frontend, and I suspected that we're not updating it properly in the proxy/child scenarios. (See screenshots I posted in Slack.)

Re: the modal dialog. I believe this is a separate issue. It appears that you can dismiss the dialog, so I suspect this PR is adding some safeguards so we don't accidentally dismiss the popup.

fat23cat:
Thanks, @yufengwng.

Yes, when the app calls `auth0RenewToken()` auth0 returns a new session without `participantGuid` because the App gets `participantGuid` from API endpoint on the login-landing page, my approach allows save `participantGuid` if it existed and reuse `participantGuid` in the renewed session.

Re: the modal dialog. Yes, I observed that sometimes I can accidentally dismiss the dialog and we don't have an obvious way to reopen it(you can reload the page, but IMO it isn't obvious for users). Now, users can close the dialog (x button), sign out, continue the session, they can't close the dialog by clicking on the background. Moreover, if a user clicked the continue button, while session is renewing, the dialog's buttons are blocked to prevent closing the dialog and clicking on the form.

ddp-workspace/projects/toolkit/src/lib/components/dialogs/sessionWillExpire.component.ts
```diff
@@ -68,13 +72,20 @@ export class SessionWillExpireComponent implements OnInit, OnDestroy {
...
+        const session = this.session.sessionObservable.pipe(
+            // Because the session is BehaviorSubject, we should skip a current session,
+            // we are only interested in the renewed session
+            skip(1)
+        ).subscribe(() => {
+            this.renewNotifier.hideSessionExpirationNotifications();
+        });
+        this.anchor.add(session);
         this.auth0.auth0RenewToken();
     }
```
MocanaAtBroad, line 81: I am surprised that flushing explicitly the previous session object is necessary! If this is indeed the case, can we push down this functionality into out of the dialog and down into one of the services?

fat23cat: Idea - get the last session object after session renewing. Why we should move it to another service? I guess this action is specific only for this case. Or did I misunderstand you?

MocanaAtBroad: Was asking for this to be moved down to a service because this looks like token renewal implementation/back-end plumbing on session management rather than a high-level API call that would be visible to a UI component.

fat23cat: Sorry, but I'm not sure that we are talking about the same thing... Because we just watching for the session. Did you pull the last changes? (I'm sorry if you did, I'm a bit confused)
image

MocanaAtBroad: I think we are talking about same thing. To be more explicit, my expectation would be that this component simply calls a `renewSession()` on one of our services associated with session management. e.g., `Auth0AdapterService` and whatever needs to happen to this correctly happens there instead of in the sessionwillexpire dialog.

fat23cat: Okay, thanks for the explanation.

MocanaAtBroad (approved):

PR #98, opened 2020-04-28
DDP-4479 and a couple of other fixes

MocanaAtBroad:
We had a couple of builds "fail" because previous versions of GAE Angular apps could not be removed.
Copied approach from pepper-apis that allows for deletion to fail a few times (out of our control).
Also noticed that our npm caching was not working. I have updated it and confirms that it now works.
Also took opportunity to update the information we write to deployment spreadsheet. See comments for details.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,6 +19,7 @@ executors:
     docker:
       - image: broadinstitute/study-website-build:04-09-2020A
     working_directory: ~/repo/ddp-workspace
+    resource_class: medium+
 
 commands:
   app-build-and-store:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,11 @@ version: 2.1
 
 references:
   npm_cache_key: &npm_cache_key
-                   v1-dependency-npm-{{ checksum "package-lock.json" }}
+                   v1-dependency-npm1-{{ checksum "package-lock.json" }}
+  npm_cache_restore_key: &npm_cache_restore_key
+    keys:
+      - *npm_cache_key
+      - v1-dependency-npm1
   # Note: important that values of guids and keys have matching order!"
   study_keys: &study_keys
                 "osteo brain angio mbc"
@@ -233,14 +237,10 @@ commands:
     steps:
       - checkout:
           path: ~/repo
-      - restore_cache:
-          keys:
-            - *npm_cache_key
-            - v1-dependency-npm-
+      - restore_cache: *npm_cache_restore_key
       - run:
           name: npm install
           command: npm install
-          working_directory: ~/repo/ddp-workspace
       - save_cache:
           key: *npm_cache_key
           paths:
@@ -357,6 +357,7 @@ commands:
 
 jobs:
   build-and-deploy-job:
+    working_directory: ~/repo/ddp-workspace
     parameters:
       study_key:
         type: string
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -253,6 +253,7 @@ commands:
           name: ng build
           command: |
             set -u
+            export NODE_OPTIONS='--max_old_space_size=4096'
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -2,7 +2,11 @@ version: 2.1
 
 references:
   npm_cache_key: &npm_cache_key
-                   v1-dependency-npm-{{ checksum "package-lock.json" }}
+                   v1-dependency-npm1-{{ checksum "package-lock.json" }}
+  npm_cache_restore_key: &npm_cache_restore_key
+    keys:
+      - *npm_cache_key
+      - v1-dependency-npm1
   # Note: important that values of guids and keys have matching order!"
```
MocanaAtBroad, line 10: The caching appeared to be broken: we were downloading all dependencies on each npm install.

.circleci/config.yml
```diff
@@ -14,7 +18,8 @@ executors:
   commit-executor:
     docker:
       - image: broadinstitute/study-website-build:04-09-2020A
-    working_directory: ~/repo
+    working_directory: ~/repo/ddp-workspace
+    resource_class: medium+
```
MocanaAtBroad, line 23: Made ddp-workspace the default directory. In only one command we need a different working_directory. So removing it from all places where setting it is unnecessary.

MocanaAtBroad: Also using a bigger instance. (Seems to make very little difference though).

.circleci/config.yml
```diff
@@ -171,11 +173,18 @@ commands:
...
+              set +e
+            fi
+
             for VERSION_TO_DELETE in $VERSIONS_TO_DELETE; do
               echo "Deleting version $VERSION_TO_DELETE of service $SERVICE"
               gcloud app versions delete --quiet --service "$SERVICE" "$VERSION_TO_DELETE" --project "broad-ddp-${ENVIRONMENT}"
             done;
-          working_directory: ~/repo/ddp-workspace
+            exit 0;
+
```
MocanaAtBroad, line 187: Transferred this change from the pepper-apis : sometimes deletion of previous services fails. We are going to accept that this is OK until we have more than 3 versions. Then we fail so we can intervene by hand.

.circleci/config.yml
```diff
@@ -225,14 +234,10 @@ commands:
     steps:
       - checkout:
           path: ~/repo
-      - restore_cache:
-          keys:
-            - *npm_cache_key
-            - v1-dependency-npm-
+      - restore_cache: *npm_cache_restore_key
       - run:
```
MocanaAtBroad, line 238: Refactored cache keys to references at top of file.

build-utils/reportdeploy.sh
```diff
@@ -51,4 +53,4 @@ OTHER_BRANCHES=$(other_branches_same_commit | xargs)
 AUTHOR=$(commit_author)
 # Passing to sheetappend the GCP credentials we are expecting to be piped into this script
 cat /dev/stdin |
-   sheetappend $RELEASE_SHEET_ID $DATE $TIME "$COMPONENT" $ENVIRONMENT $BRANCH "$TAG" $SHORT_GIT_SHA "$AUTHOR" $CIRCLE_BUILD_NUM "$CIRCLE_BUILD_URL" "$OTHER_BRANCHES"
+   sheetappend $RELEASE_SHEET_ID $DATE $TIME "$COMPONENT" $ENVIRONMENT $BRANCH "$TAG" $SHORT_GIT_SHA "$AUTHOR" "$CIRCLE_BUILD_HYPERLINK" "" "$OTHER_BRANCHES"
```
MocanaAtBroad, line 56: Instead of having a column for the build number and another for the build url, we are just going to create a hyperlink in the spreadsheet. This allows us to see an important column that is currently hiding (build branches). Leaving the URL blank column in place for now.

fat23cat (approved):
I've tested updated config on osteo, caching, and removing old services works well.

PR #100, opened 2020-04-29
DDP-4423: Modify build of Angular builds to include sourcemaps in dev and test

fat23cat:
Looks like source maps don't affect app performance:
Without source maps:
Screenshot_2
With source maps:
Screenshot_1

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,10 +49,10 @@ commands:
       - unless:
           condition: << parameters.skip_build >>
           steps:
-          - set-deployment-environment
-          - build-app
-          - deploy-exploded-build:
-              study_key: << parameters.study_key >>
+            - set-deployment-environment
+            - build-app
+            - deploy-exploded-build:
+                study_key: << parameters.study_key >>
 
   set-deployment-environment:
     description: "Determine working environment"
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -121,7 +126,6 @@ commands:
       study_key:
         type: string
     steps:
-      - set-deployment-environment
```
fat23cat, line 124: Looks like when circle ci executed `deploy-stored-build`, `set-deployment-environment` was called twice: in the `deploy-stored-build` and then in the `deploy-exploded-build`.

.circleci/config.yml
```diff
@@ -49,6 +49,7 @@ commands:
       - unless:
           condition: << parameters.skip_build >>
           steps:
+          - set-deployment-environment
```
fat23cat: We should determine working environment before building an app.

MocanaAtBroad (approved):
Was able to set breakpoint and bring up debugger on dev. Woo hoo!

MocanaAtBroad:
@fat23cat : was looking at your performance measurement screenshot. Looked like we can save a second (or more) by turning on text compression. Do you know what they are referring to and how we could go about it? A GAE static resources config thing?

fat23cat:
> @fat23cat : was looking at your performance measurement screenshot. Looked like we can save a second (or more) by turning on text compression. Do you know what they are referring to and how we could go about it? A GAE static resources config thing?

I think this feature should be enabled on the tcell side. https://web.dev/uses-text-compression/?utm_source=lighthouse&utm_medium=devtools

PR #101, opened 2020-04-30
Use npm ci instead of npm install in Circle CI

fat23cat:
https://docs.npmjs.com/cli/ci.html#description

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 
 references:
   npm_cache_key: &npm_cache_key
-                   v1-dependency-npm1-{{ checksum "package.json" }}-{{ checksum "package-lock.json" }}
+                   v1-dependency-npm1-{{ checksum "package-lock.json" }}
   npm_cache_restore_key: &npm_cache_restore_key
     keys:
       - *npm_cache_key
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -236,7 +236,7 @@ commands:
       - restore_cache: *npm_cache_restore_key
       - run:
           name: Install dependencies
-          command: dir
+          command: "[ ! -d ./node_modules ] && npm ci"
       - save_cache:
           key: *npm_cache_key
           paths:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,6 @@ references:
   npm_cache_restore_key: &npm_cache_restore_key
     keys:
       - *npm_cache_key
-      - v1-dependency-npm1
   # Note: important that values of guids and keys have matching order!"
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -2,11 +2,10 @@ version: 2.1
 
 references:
   npm_cache_key: &npm_cache_key
-                   v1-dependency-npm1-{{ checksum "package-lock.json" }}
+                   v1-dependency-npm1-{{ checksum "package.json" }}-{{ checksum "package-lock.json" }}
```
fat23cat, line 5: I add `package.json`'s checksum to prevent restoring the wrong cache if someone updated `package.json` and didn't commit `package-lock.json`

.circleci/config.yml
```diff
@@ -236,8 +235,13 @@ commands:
...
       - run:
-          name: npm install
-          command: npm install
+          name: Install dependencies
+          command: |
+            if [[ ! -d node_modules ]]; then
+              npm ci
+            else
+              echo "package.json and package-lock.json unchanged, using cache"
+            fi
```
fat23cat, line 248: If someone forgot to commit new `package-lock.json` build will be failed https://circleci.com/gh/broadinstitute/ddp-angular/1453

.circleci/config.yml
```diff
@@ -2,11 +2,10 @@ version: 2.1
 
 references:
   npm_cache_key: &npm_cache_key
-                   v1-dependency-npm1-{{ checksum "package-lock.json" }}
+                   v1-dependency-npm1-{{ checksum "package.json" }}-{{ checksum "package-lock.json" }}
   npm_cache_restore_key: &npm_cache_restore_key
     keys:
       - *npm_cache_key
-      - v1-dependency-npm1
```
fat23cat, line 9: I guess, we only need `npm_cache_key`, isn't it?

MocanaAtBroad: The way the CircleCi cache works is it tries to do a string match in the order listed.
The first one is meant to be an exact match and presumably would avoid npm having to download anything. All the packages are in the cache.
The next entry in the list of keys is meant for an inexact match. npm will have to download stuff, but hopefully not every single package. 
With the ci flag, the second entrywill take a bit of thought. Perhaps it is all or nothing with the ci flag?

fat23cat: Thanks for the explanation!
I think it would be better if we restrict the cache version to be sure that we use correct packages versions. Moreover, we don't change packages versions very often and `npm ci` works fast, so we can install deps if we didn't find a cache.

MocanaAtBroad (dismissed):

MocanaAtBroad (approved):

PR #107, opened 2020-05-06
Fix circleci repo paths, etc.

MocanaAtBroad:
Discovered that after a recent tweak paths were messed. Main symptom was that are build-utils scripts could not be found because PATH was not set properly.
Then found that we were getting errors when trying to rebuild the Angular node modules. Needed to include some additional programs installed in supporting Docker image.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,6 +78,9 @@ commands:
               staging)
                 DEPLOY_ENV=staging
               ;;
+              deletemedummy:
+                DEPLOY_ENV=dev
+              ;;
               production)
                 DEPLOY_ENV=prod
               ;;
@@ -407,7 +410,8 @@ jobs:
     executor:
       name: build-executor
     steps:
-      - checkout
+      - checkout:
+          path: *repo_path
       - conditionally-launch-build-and-store:
           study_key: osteo
       - conditionally-launch-build-and-store:
@@ -494,6 +498,7 @@ workflows:
                 - test
                 - staging
                 - production
+                - deletemedummy
           study_key: osteo
 
       - deploy-stored-build-job:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -254,6 +254,7 @@ commands:
       - run:
           name: Install dependencies
           command: |
+            echo "the PWD is $PWD"
             if [[ ! -d node_modules ]]; then
               npm ci
             else
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,9 +20,9 @@ references:
 
 # Container for all the builds
 executors:
-  commit-executor:
+  build-executor:
     docker:
-      - image: broadinstitute/study-website-build:04-09-2020A
+      - image: broadinstitute/study-website-build:05-06-2020A
     working_directory: *ng_workspace_path
     resource_class: medium+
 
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -1,8 +1,12 @@
 version: 2.1
 
 references:
+  repo_path: &repo_path
+      /home/circleci/repo
+  ng_workspace_path: &ng_workspace_path
+      /home/circleci/repo/ddp-workspace
```
MocanaAtBroad, line 7: Setting these up as " constants" is what we are doing in the ddp-study-server. Can be referenced wherever needed.

.circleci/config.yml
```diff
@@ -1,8 +1,12 @@
...
   npm_cache_key: &npm_cache_key
-    v1-dependency-npm1-{{ checksum "package.json" }}-{{ checksum "package-lock.json" }}
+    v1-dependency-npm2-{{ checksum "package.json" }}-{{ checksum "package-lock.json" }}
   npm_cache_restore_key: &npm_cache_restore_key
```
MocanaAtBroad, line 10: node_modules will now live in different absolute path, so old caches will not help us. So tweaked base key so we can refer to caches stored in new location.

.circleci/config.yml
```diff
@@ -16,10 +20,10 @@ references:
 
 # Container for all the builds
 executors:
-  commit-executor:
+  build-executor:
     docker:
-      - image: broadinstitute/study-website-build:04-09-2020A
-    working_directory: ~/repo/ddp-workspace
+      - image: broadinstitute/study-website-build:05-06-2020A
```
MocanaAtBroad, line 25: New image. Already pushed to Docker hub.

.circleci/config.yml
```diff
@@ -16,10 +20,10 @@ references:
...
+    working_directory: *ng_workspace_path
     resource_class: medium+
```
MocanaAtBroad, line 27: Using the new "constant" workspace path location. ddp-workspace is where most of the action occurs, so might as well make it the default.

.circleci/config.yml
```diff
@@ -30,6 +34,7 @@ commands:
         type: string
     steps:
       - setup-build-workspace
+      - set-deployment-environment
```
MocanaAtBroad, line 37: Turns out we need to set deployment environment as this is where we get the build params now that determine whether we include source code or not in builds.

.circleci/config.yml
```diff
@@ -114,7 +119,6 @@ commands:
             fi
             TAR_FILE_PATH="/tmp/$(basename -- $TAR_FILE_URL)"
             gsutil cp  $TAR_FILE_URL $TAR_FILE_PATH
-            cd ~/repo/ddp-workspace
```
MocanaAtBroad, line 117: It's the default already.

build-utils/Dockerfile-BuildDeploy
```diff
@@ -114,6 +114,8 @@ RUN apk  --no-cache add \
     ruby \
     ruby-bundler \
     ruby-json  \
+    make \
+    g++ \
     && rm -rf /var/cache/apk/*
```
MocanaAtBroad, line 119: npm install sometimes wants to build some binaries. These are the programs it wants to use to do so.

zyme (approved):

fat23cat (approved):

PR #108, opened 2020-05-07
TestBoston app workflow

fat23cat:
I've connected testboston frontend with backend, available features for now:
- testboston.dev.datadonationplatform.org is protected by irb password page
- you can register new user, sign in an existing account, logout
- you can go through workflow and see the dashboard


yufengwng (approved):
Nice!! LGTM

MocanaAtBroad (approved):

PR #119, opened 2020-05-23
Fix mailing address busy bug and highlight missing fields on load

MocanaAtBroad:
This is the bug we observed with Yufeng when walking through TestBoston.
One line fix and test updated to make sure does not happen again.
Also noticed that we did not highlight missing fields in address when loading a temporary or profile address. Have fix and test for that too.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -273,7 +273,7 @@ commands:
           name: ng test
           command: |
             mkdir -p /tmp/junit
-            ng test ddp-sdk --watch=false --browsers ChromeHeadlessNoSandbox
+            ng test ddp-sdk --watch=false --browsers ChromeHeadlessNoSandbox --reporters junit
           environment:
             JUNIT_REPORT_PATH: /tmp/junit/
             JUNIT_REPORT_NAME: test-results.xml
--- a/ddp-workspace/projects/ddp-sdk/karma.conf.js
+++ b/ddp-workspace/projects/ddp-sdk/karma.conf.js
@@ -21,7 +21,7 @@ module.exports = function (config) {
       reports: ['html', 'lcovonly'],
       fixWebpackSourcePaths: true
     },
-    reporters: ['junit'],
+    reporters: ['junit', 'kjhtml'],
     junitReporter: {
       outputDir: process.env.JUNIT_REPORT_PATH,
       outputFile: process.env.JUNIT_REPORT_NAME,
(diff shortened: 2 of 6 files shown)


.circleci/config.yml
```diff
@@ -273,7 +273,7 @@ commands:
           name: ng test
           command: |
             mkdir -p /tmp/junit
-            ng test ddp-sdk --watch=false --browsers ChromeHeadlessNoSandbox
+            ng test ddp-sdk --watch=false --browsers ChromeHeadlessNoSandbox --reporters junit
```
MocanaAtBroad, line 276: Explicitly saying which reporter to use now that we have multiple ones. More about this below.

ddp-workspace/projects/ddp-sdk/karma.conf.js
```diff
@@ -21,7 +21,7 @@ module.exports = function (config) {
       reports: ['html', 'lcovonly'],
       fixWebpackSourcePaths: true
     },
-    reporters: ['junit'],
+    reporters: ['junit', 'kjhtml'],
     junitReporter: {
```
MocanaAtBroad, line 25: Need the kjhtml reporter so we can see results when running tests locally in the browser.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressInput.service.ts
```diff
@@ -365,8 +365,8 @@ export class AddressInputService implements OnDestroy {
         const newFieldValue = address[propName];
         if (formControl && newFieldValue !== formControl.value) {
           formControl.patchValue(newFieldValue, { emitEvent: emitValueChange });
-          markFieldsAsTouched && formControl.markAsTouched();
         }
+        markFieldsAsTouched && formControl.markAsTouched();
```
MocanaAtBroad, line 369: Found another issue that when we load a saved address (from profile or a temporary one) we would not highlight in the form missing values (e.g., the  NAME field was missing). To make sure formgroup validations are visible, we touch all the fields in the form when load an address.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressInput.service.spec.ts
```diff
@@ -150,4 +150,26 @@ describe('AddressInputService', () => {
...
+      state : 'AK',
+      zip: '90120',
+      phone: '867-5309',
+      city: 'Fairbanks',
+      guid: '123'
+    };
+    tick(2);
+    ais.inputAddress$.next(new Address(addressData));
+    // check that when set an address, we ensure the controls are touched
+    expect(Object.values(ais.addressForm.controls).filter(control => !control.touched).length).toBe(0);
```
MocanaAtBroad, line 172: New test for the change below.

fat23cat (approved):
Works well.

yufengwng (approved):
👍 thanks for the fix!

PR #120, opened 2020-05-24
Add label to picklist question if one exists

MocanaAtBroad:
Took stab at trying to add a label to radio groups in composite questions. Label is there but not happy with it. Looking for input there. This is what it looks like with the change I introduced, right above "Positive":
Screen Shot 2020-05-24 at 1 15 13 PM

On client we create a label attribute that holds the child question text.
But radio groups are different than other Angular material input widgets. Labels are not part of mat-form-fields.
I added label following example at https://material.angular.io/components/radio/examples
But styling just looks off in COVID survey. And not sure that even if we had exact same styling as the label for the date, it would look right anyways.
Maybe not using mat-label for this question at all? Not sure how to do this for this question in testboston.
I don't have a good solution for this at this point.
A separate but related issue: one thing that would appear to improve appearance in the covid survey would be to layout the radio options horizontally rather than vertically. I this is what that would look like:

Screen Shot 2020-05-24 at 1 22 02 PM

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/radiobuttonsActivityPicklistQuestion.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/radiobuttonsActivityPicklistQuestion.component.ts
@@ -8,7 +8,8 @@ import { NGXTranslateService } from '../../../services/internationalization/ngxT
     selector: 'ddp-activity-radiobuttons-picklist-question',
     template: `
     <div>
-        <mat-radio-group class="ddp-radio-group-column">
+        <label id="radio-group-label" *ngIf="block.label" [innerHTML]="block.label"></label>
+        <mat-radio-group class="ddp-radio-group-column" aria-labelledby="radio-group-label">
             <div *ngFor="let option of block.picklistOptions" class="margin-5">
                 <mat-radio-button
                             [checked]="getOptionSelection(option.stableId)"


fat23cat (approved):
In my opinion, at this stage, the appearance is acceptable. When we will have accurate activities mockups, we can change labels font, color, size as we want using css.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/radiobuttonsActivityPicklistQuestion.component.ts
```diff
@@ -8,7 +8,8 @@ import { NGXTranslateService } from '../../../services/internationalization/ngxT
     selector: 'ddp-activity-radiobuttons-picklist-question',
     template: `
     <div>
-        <mat-radio-group class="ddp-radio-group-column">
+        <label id="radio-group-label" *ngIf="block.label" [innerHTML]="block.label"></label>
```
yufengwng, line 11: Is there anything we can do in `ActivityCompositeAnswer` component? From a cursory reading of the code, `shouldSetLabelToBeQuestionText()` returns `true` because we do have a "date picklist" child, which makes `convertQuestionToLabels` to be `true` and causing question prompt to be blanked out / converted to `label`?

MocanaAtBroad: I think first we just need to decide how we want to present this question and work our solution back from that. I actually like that the we are producing a label element for the radio buttons, just don't like how it looks with current styling and in contrast with the sibling date question.
Does not help that in material angular, the radio buttons, radio groups are treated differently that the other input widgets regarding labels.
The general solutions in composite answer are starting to get complicated!
I am inclined for now to follow @fat23cat suggestion and merge and come back to this as we add more styling. to the whole system. He is still refining them.

PR #121, opened 2020-05-25
Testboston: url and guid configuration for address activity

fat23cat:


ddp-workspace/projects/ddp-testboston/src/app/app-guids.ts
```diff
@@ -1,5 +1,6 @@
 export const AppGuids = {
     Consent: 'CONSENT',
     Dashboard: 'DASHBOARD',
-    Covid: 'COVID_SURVEY'
+    Covid: 'COVID_SURVEY',
```
MocanaAtBroad: Could we rename this to something like ActivityGuids ? AppGuid does not sound right to me.

MocanaAtBroad (dismissed):

MocanaAtBroad (approved):

PR #117, opened 2020-05-22
DDP-4588, DDP-4587, DDP-4570: Testboston homepage, footer, header

fat23cat:
Mockups are here https://projects.invisionapp.com/d/main#/projects/prototypes/19934732

PR #122, opened 2020-05-25
Testboston: agree/not agree buttons

fat23cat:
I didn't care about styles too much, because the styles of the buttons are in #117.
Appearance of the last section:
image
Mockup:
image

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-redesigned.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-redesigned.component.ts
@@ -5,7 +5,8 @@ import {
     Injector,
     OnDestroy,
     OnInit,
-    Renderer2
+    Renderer2,
+    Input
 } from '@angular/core';
 import { DOCUMENT } from '@angular/common';
 import { ActivityComponent } from './activity.component';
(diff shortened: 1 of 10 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-redesigned.component.ts
```diff
@@ -156,4 +171,8 @@ export class ActivityRedesignedComponent extends ActivityComponent implements On
         injector: Injector) {
         super(windowRef, renderer, submitService, analytics, document, injector);
     }
+
+    public isAgree(): boolean {
+        return this.model.activityCode === 'CONSENT';
```
fat23cat: @yufengwng, it looks like we must give the TestBoson consent activity a unique activityCode not to affect other studies.

yufengwng: Here's a wild idea:

* Make `useAgreeConsent: boolean` an `@Input` for `ActivityRedesignedComponent`
* Make `isAgreeConsent()` method check for `useAgreeConsent` and `model.formType === 'CONSENT'`
* Add optional `useAgreeConsent` to toolkit config
* Change `ActivityPageRedesignedComponent` in toolkit to read `useAgreeConsent` from config and pass flag accordingly to `ddp-activity-redesigned`

Will that work?

MocanaAtBroad (dismissed):
I am OK with this approach to deal with this (for now). Not happy about mentally having to process 5 different booleans flags to figure out under what conditions we will have show a button or not, but I think this class was already a candidate for major refactoring before this. There are certain parts of the functionality delivered that I am no longer able to understand. This change will get us over the hump. Then we should come back and refactor.

MocanaAtBroad (approved):

PR #125, opened 2020-05-26
Move recaptcha site key to vault and read value from pepperConfig.js

MocanaAtBroad:
To allow for different keys in different environments. Will be also renaming on server side as name I had in there that left out that it was server or client was confusing me.


yufengwng (approved):
Looks good from what I can tell. Good call on the rename!

fat23cat (approved):

PR #126, opened 2020-05-27
DDP-4592: Registration page

fat23cat:
- Added "Are you 18 years old or older?" question
- Texts moved into the translation file
- Improved styles and mobile version
- Added validation messages under fields
- Added loading spinner


ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/invitationServiceAgent.service.ts
```diff
@@ -44,6 +46,17 @@ export class InvitationServiceAgent extends NotAuthenticatedServiceAgent<any> {
...
-        return new DdpError(serverError.error ? serverError.error.message : '', serverError.error ? serverError.error.code : null);
+        return new DdpError(this.buildErrorMessage(serverError), this.buildErrorType(serverError));
+    }
+
+    private buildErrorMessage(serverError: HttpErrorResponse): string {
+        return serverError.error ? serverError.error.message : '';
+    }
+
+    private buildErrorType(serverError: HttpErrorResponse): ErrorType {
+        if (serverError.error && Object.values(ErrorType).includes(serverError.error.code)) {
```
MocanaAtBroad, line 57: Could you double-check this line? Under impression that enums don't exist in JavaScript so have not idea what using an enum for argument to Object.values will do. Maybe  switching to checking values of serverError.error.code would be safer?

fat23cat: How about transforming enum into an object?

fat23cat: Ah, it won't work as type :(

fat23cat: I think everything should be okay, `Object.values(ErrorType)` transfrorms into:
image

MocanaAtBroad: I did same experiment. Guess TypeScript has made some progress!

MocanaAtBroad (approved):

PR #130, opened 2020-05-28
DDP-4592: Registration page - error messages

fat23cat:
Added new error messages text and updated phone number
image


MocanaAtBroad (approved):
The code looks and works well, but the message we are displaying for bad zip codes is problematic. Suggests to "try again". If your home zip code is really not one of those supported, there is no point to try again unless you want to cheat and lie. Do merge, but I'll follow up with Jen and Erin on that one.

PR #131, opened 2020-05-28
DDP-4607: Handling for non-consent

fat23cat:


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-redesigned.component.ts
```diff
@@ -134,18 +134,19 @@ import { SubmissionManager } from '../../services/serviceAgents/submissionManage
                                     [innerHTML]="'SDK.CloseButton' | translate">
                             </button>
                         </ng-container>
-                        <ng-container *ngIf="(!isStepped || isLastStep) && isAgree() && !model.readonly && isLoaded">
-                            <button class="button button_medium button_primary"
+                        <ng-container *ngIf="(!isStepped || isLastStep) && isAgree() && isLoaded">
```
yufengwng, line 137: I see, so if it's read-only, we'll show the regular "close" button.

yufengwng (commented):
I think it will be nice if there's a bit more separation between `isAgree()` and `readonly` but seems fine for now for a bit more reuse of the button.

MocanaAtBroad (approved):

PR #141, opened 2020-06-02
DDP-4708: Session will expire dialog

fat23cat:
Just added dialog and styles for it.


ddp-workspace/projects/ddp-testboston/src/app/components/welcome/welcome.component.scss
```diff
@@ -100,6 +100,10 @@
     border-top: 1px solid mat-color($app-theme, 500);
 }
 
+.mat-expansion-panel-header {
+    transition: 0.3s;
+}
```
fat23cat, line 105: This change does not apply to the modal window. Just added a smooth hover animation for the expansion panel on the home page.

MocanaAtBroad (approved):

PR #147, opened 2020-06-04
DDP-4628 support admin auth

yufengwng:
Main pieces of admin auth:

* `Auth0AdapterService` provides method to login as admin. Sign up as admin is not supported.
* `AdminLoginLandingComponent` is where we handle the callback from Auth0.
* `Auth0CodeCallbackComponent` has been updated to handle local admin auth flow.
* `SessionMementoService` now stores a flag for `isAdmin`.
* `AdminAuthGuard` can be used to require admin auth for a path.
* SDK and toolkit configs now have config values for facilitating admin auth.

I have tried this locally and on `dev` and seems to work so far. Note that this is the auth portion and is mostly decoupled from the Okta SAML integration. These changes should facilitate admin auth regardless of what auth0 integration we actually use on the other end.


MocanaAtBroad (commented):
All this looks fine to me.
Was looking for a way to not have an almost-identical AuthGuard implementation. They differ only on on passing the client id. However the Angular dependency injection is kind of pain to use and when I tried to refactor to have the one implementation and two services that are configured with different Auth0 client ids, my solution was more complicated than actually rewriting it just like you did.
@fat23cat if you can come up with a nicer way to just just have one AuthGuard class and two services please do chime in.
Otherwise, I would say let's merge this in ASAP so we can start kicking tires.

fat23cat:
@MocanaAtBroad, I also tried to find the possibility to avoid writing the same code but found only one possible way - we can use `next` and see which URL a user wants to open and if it's `crc-dashboard`, pass adminCliedtID.
image
But I really don't like that idea, so, let's merge.

fat23cat (dismissed):

ddp-workspace/projects/toolkit/src/lib/components/login-landing/admin-login-landing.component.ts
```diff
@@ -0,0 +1,70 @@
...
+import {
+  SessionMementoService,
+  Auth0AdapterService,
+  ConfigurationService
+} from 'ddp-sdk';
+
+@Component({
+  selector: 'toolkit-admin-login-landing',
+  template: `
+      <toolkit-common-landing></toolkit-common-landing>
```
fat23cat:
```suggestion
      <toolkit-common-landing-redesigned></toolkit-common-landing-redesigned>
```

ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
```diff
@@ -198,31 +244,42 @@ export class Auth0AdapterService implements OnDestroy {
...
+            this.adminWebAuth.logout({
+                returnTo: `${baseUrl}${baseUrl.endsWith('/') ? '' : '/'}${returnToUrl}`,
+                clientID: this.configuration.adminClientId
+            });
+        } else {
+            this.webAuth.logout({
+                returnTo: `${baseUrl}${baseUrl.endsWith('/') ? '' : '/'}${returnToUrl}`,
+                clientID: this.configuration.auth0ClientId
+            });
+        }
```
fat23cat:
```suggestion
        const returnTo = `${baseUrl}${baseUrl.endsWith('/') ? '' : '/'}${returnToUrl}`;
        if (wasAdmin) {
            this.adminWebAuth.logout({
                returnTo,
                clientID: this.configuration.adminClientId
            });
        } else {
            this.webAuth.logout({
                returnTo,
                clientID: this.configuration.auth0ClientId
            });
        }
```

fat23cat (approved):

PR #148, opened 2020-06-05
DDP-4604: Dashboard for CRCs

fat23cat:
It's a draft PR because we can't login as study admin.

**08/06/2020 update:**
- Implemented code review suggestions
- Updated UI behavior 
- Updated styling


ddp-workspace/projects/ddp-testboston/src/app/components/crc-dashboard/crc-dashboard.component.ts
```diff
@@ -0,0 +1,104 @@
...
+        this.subject = null;
+      }),
+      filter(form => this.searchForm.valid && !!form.invitationId),
+      tap(() => {
+        this.isLoading = true;
+        this.ref.detectChanges();
+      }),
+      map(form => form.invitationId),
+      // todo: remove delay
+      delay(3000),
```
fat23cat: Added delay to simulate requests.

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/subjectInvitationServiceAgent.service.ts
```diff
@@ -0,0 +1,65 @@
...
+                voidedAt: '',
+                verifiedAt: '',
+                acceptedAt: '',
+                userGuid: '',
+                userHruid: '',
+                userLoginEmail: '',
+                notes: ''
+            });
+        }
+        return of(null);
```
fat23cat: Just mocked two invitation codes.

ddp-workspace/projects/ddp-testboston/src/app/components/crc-dashboard/crc-dashboard.component.ts
```diff
@@ -0,0 +1,104 @@
...
+    this.sessionService.setParticipant(this.subject.userGuid);
+  }
+
+  public saveNotes(notes: string): void {
+    this.notesSubject.next(notes);
+  }
+
+  public enrollSubject(): void {
+    // TBD
+  }
```
fat23cat: It's still not clear what we should do here.

ddp-workspace/projects/ddp-testboston/src/app/components/crc-dashboard/crc-dashboard.component.ts
```diff
@@ -0,0 +1,104 @@
...
+      }),
+      filter(form => this.searchForm.valid && !!form.invitationId),
+      tap(() => {
+        this.isLoading = true;
+        this.ref.detectChanges();
+      }),
+      map(form => form.invitationId),
+      // todo: remove delay
+      delay(3000),
+      mergeMap(invitationId => this.subjectInvitation.lookupInvitation(invitationId)),
```
MocanaAtBroad: I favor switchMap for these situations. No point in continuing to handle an invitationId user is no longer interested in.

ddp-workspace/projects/ddp-testboston/src/app/components/crc-dashboard/crc-dashboard.component.ts
```diff
@@ -0,0 +1,104 @@
...
+        this.error = new DdpError('', ErrorType.InvitationNotFound);
+      }
+    });
+    this.anchor.addNew(form);
+  }
+
+  private initNotesListener(): void {
+    const notes = this.notesSubject.asObservable().pipe(
+      debounceTime(1000),
+      mergeMap(notes => this.subjectInvitation.updateInvitationDetails(this.subject.invitationId, notes))
```
MocanaAtBroad: concatMap might be a better choice here: you can end up with multiple mergeMap requests in flight and going to separate servers and no guarantee that the last one out from client will be the last one saved on the server.

MocanaAtBroad: You already have a formGroup. Perhaps just simpler to add `notes` as additional group controller subscribe to form or controller value changes?

MocanaAtBroad: Do we have to worry about losing a note if user leaves page during the 1 sec debounce period?

ddp-workspace/projects/ddp-sdk/src/lib/pipes/invitationFormatter.pipe.ts
```diff
@@ -0,0 +1,11 @@
...
+
+@Pipe({
+    name: 'invitation'
+})
+export class InvitationPipe implements PipeTransform {
+    transform(code: string, chunk: number = 4, separator: string = ' - '): string {
+        const regexp = new RegExp(`.{1,${chunk}}`, 'g');
+        return code.match(regexp).join(separator);
+    }
+}
```
MocanaAtBroad, line 11: Nice idea to have create a pipe!

ddp-workspace/projects/ddp-testboston/src/app/components/crc-dashboard/crc-dashboard.component.ts
```diff
@@ -0,0 +1,103 @@
...
+  private initSearchListener(): void {
+    const form = this.searchForm.valueChanges.pipe(
+      tap(() => {
+        this.error = null;
+        this.subject = null;
+      }),
+      filter(form => this.searchForm.valid && !!form.invitationId),
+      tap(() => {
+        this.isLoading = true;
+        this.ref.detectChanges();
```
MocanaAtBroad: Interesting. You had to detect changes?

ddp-workspace/projects/ddp-testboston/src/app/components/crc-dashboard/crc-dashboard.component.ts
```diff
@@ -0,0 +1,103 @@
...
+      }),
+      map(form => form.invitationId),
+      // todo: remove delay
```
MocanaAtBroad: I am getting a tslint warning here: Shadowed name

MocanaAtBroad (commented):
Had a few comments but looks really good! I really like how just filling out the invitation code does immediately does the lookup without a button click.

ddp-workspace/projects/ddp-testboston/src/app/components/crc-dashboard/crc-dashboard.component.ts
fat23cat: > You already have a formGroup. Perhaps just simpler to add notes as additional group controller subscribe to form or controller value changes?

I thought about it, but I don't want to make notes as part of formGroup because initially notes field is hidden. In my opinion, it will make the component more complex: the search field and notes are situated in different sections. 

> Do we have to worry about losing a note if user leaves page during the 1 sec debounce period?

I'm also worried about it, maybe we can make a debounce period 0.5 sec or less?

ddp-workspace/projects/ddp-testboston/src/app/components/crc-dashboard/crc-dashboard.component.ts
fat23cat: Yep, the search field was unhappy when I made it read-only without detect changes.

ddp-workspace/projects/ddp-testboston/src/app/components/crc-dashboard/crc-dashboard.component.ts
fat23cat: I think, we can leave mergeMap, because we block input while request in flight. And user can't send another request.

MocanaAtBroad: Right. But with switchMap idea is that you don't have to block/disable. If new id is entered while another one in flight, you cancel the inflight one and start working on the new one. Don't think a huge deal here as this is not like Google search string or something like that, but nice not having to block.

ddp-workspace/projects/ddp-testboston/src/app/components/crc-dashboard/crc-dashboard.component.ts
MocanaAtBroad: I guess another option here is to have the onDestroy wait for completion or delay for 1 sec or whatever. Pick what you think makes more sense.

ddp-workspace/projects/ddp-testboston/src/app/components/crc-dashboard/crc-dashboard.component.ts
fat23cat: Got rid of change detection.

ddp-workspace/projects/ddp-testboston/src/app/components/crc-dashboard/crc-dashboard.component.ts
```diff
@@ -0,0 +1,112 @@
...
+      filter(invitationId => this.searchForm.valid && !!invitationId),
+      tap(() => this.isLoading = true),
+      switchMap(invitationId => this.subjectInvitation.lookupInvitation(invitationId)),
+      tap(() => this.isLoading = false)
+    ).subscribe(response => {
+      if (response && this.areInvitationsEqual(response.invitationId)) {
+        this.studySubject = response;
+      } else if (response === null && this.isInvitationValid()) {
+        this.error = new DdpError('', ErrorType.InvitationNotFound);
+      }
```
fat23cat, line 94: Improved UI behavior: in case a CRC imputed valid invitation id and then edited and now invitation id is invalid, UI shouldn't show request result.

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/subjectInvitationServiceAgent.service.ts
```diff
@@ -0,0 +1,37 @@
...
+        logger: LoggingService) {
+        super(session, configuration, http, logger);
+    }
+
+    public updateInvitationDetails(invitationId: string, notes: string): Observable<any | null> {
+        return this.postObservable(`/studies/${this.configuration.studyGuid}/invitation-details`, { invitationId, notes });
+    }
+
+    public lookupInvitation(invitationId: string): Observable<StudySubject | null> {
+        return this.postObservable(`/studies/${this.configuration.studyGuid}/invitation-lookup`, { invitationId });
```
MocanaAtBroad: I think we might be returning response object and not just the body in all these methods.

fat23cat: Oh, good catch, thanks!

MocanaAtBroad (dismissed):

yufengwng (approved):

PR #153, opened 2020-06-15
DDP-4709: Implement Testboston acitivities CSS styling

fat23cat:
- Added new consent's stepper component
- Added questions styles


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-redesigned.component.ts
```diff
@@ -177,6 +195,6 @@ export class ActivityRedesignedComponent extends ActivityComponent implements On
     }
 
     public isAgree(): boolean {
-        return this.model.activityCode === 'CONSENT' && this.agreeConsent && !this.model.readonly;
+        return this.model.activityCode === 'CONSENT' && this.agreeConsent;
```
MocanaAtBroad: Not sure if we should hard-code an activity code into SDK. If we want to provide special treatment in SDK for consent activities (which this looks like) perhaps we we need to make these configurable? As we saw in Osteo, we can have more than one consent in a study.
Further: are we sure that this behavior is desirable or applicable for consent only? Wonder if it this is applicable for any multipart activity that has an agreement question.

MocanaAtBroad (commented):
Also, was noticing that we use a different font in the consent text than we do for the questions in the activities. Is this by design?

fat23cat:
> Also, was noticing that we use a different font in the consent text than we do for the questions in the activities. Is this by design?

Yes, consent section here: https://projects.invisionapp.com/d/main/#/projects/prototypes/19934732

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-redesigned.component.ts
fat23cat: I don't like this approach too, but I thought we've already discussed that https://github.com/broadinstitute/ddp-angular/pull/122

MocanaAtBroad: I think @yufengwng was suggesting we check the formType attribute. https://pepper-dev.datadonationplatform.org/docs#operation/getActivity CONSENT is one of three possible values and would deal with my hard-coding concern.

fat23cat: Oh, right, my bad.

MocanaAtBroad: I am confused about the `agreeConsent` @Input param. I can't figure out what is the meaning of it. Are you sure we need that? If you look at the list of input params for  the activity component, that param stands out as looking very different than the others. Appears to be only set to true via configuration for testboston.

fat23cat: We need `agreeConsent` **@input** param to show agree buttons only for testboston study.
Otherwise, we will show agree buttons in other studies too.

MocanaAtBroad (approved):

PR #155, opened 2020-06-15
Ddp 4719 language selector

charlottedye:
This pull request covers DDP-4719.  It adds a language selector component to the header for studies with more than one configured language.  This does not cover the work to read the user's language from the profile or write the user's language to the profile or the work to pass the language to the Auth0 universal login (or to display the language selector on the Auth0 universal login page).  This does cover the work to do the translations for an unauthenticated user.

### DIFF
--- a/ddp-workspace/projects/ddp-angio/src/styles.scss
+++ b/ddp-workspace/projects/ddp-angio/src/styles.scss
@@ -310,6 +310,7 @@ a:focus {
     stroke: mat-color($app-theme, 100);
     fill: transparent;
     vertical-align: middle;
+    transition: stroke 0.3s ease;
 }
 
 .SimpleButton--Scrolled .ddp-globe {
--- a/ddp-workspace/projects/ddp-angio/src/styles.scss
+++ b/ddp-workspace/projects/ddp-angio/src/styles.scss
@@ -306,6 +306,35 @@ a:focus {
     color: mat-color($app-theme, 600);
 }
 
+.SimpleButton .ddp-globe {
+    stroke: mat-color($app-theme, 100);
+    fill: transparent;
+    vertical-align: middle;
+}
+
+.SimpleButton--Scrolled .ddp-globe {
+    stroke: mat-color($app-theme, 600);
+    fill: transparent;
+}
+
+.SimpleButton:hover .ddp-globe {
+    stroke: mat-color($app-theme, 700);
+    fill: transparent;
+}
+
+.SimpleButton:active .ddp-globe {
+    stroke: mat-color($app-theme, 600);
+    fill: transparent;
+}
+
+.ddp-current-language {
+    vertical-align: middle;
+}
+
+mat-icon.ddp-dropdown-arrow {
+    vertical-align: middle;
+}
+
 .LastUpdatedText {
     text-align: right;
     font-weight: 300;
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
@@ -1,0 +1,97 @@
+import { Component, Input, OnInit } from "@angular/core";
+import { StudyLanguage } from "../models/studyLanguage";
+import { LanguageServiceAgent } from "../services/serviceAgents/languageServiceAgent.service";
+import { LanguageService } from "ddp-sdk";
+
+@Component({
+  selector: 'ddp-language-selector',
+  template: `
+    <div [hidden]="!loaded">
+      <button class="SimpleButton" [ngClass]="{'SimpleButton--Scrolled': isScrolled}" *ngIf="currentLanguage !== null && currentLanguage !== undefined" [matMenuTriggerFor]="menu">
+        <svg class="ddp-globe">
+          <use href="assets/images/globe.svg#Language-Selector-3" height="24px" width="24px"></use>
+        </svg>
+        <span class="ddp-current-language">{{currentLanguage.displayName}}</span>
+        <mat-icon class="ddp-dropdown-arrow">arrow_drop_down</mat-icon>
+      </button>
+      
+      <mat-menu #menu="matMenu">
+        <button mat-menu-item *ngFor="let lang of getUnselectedLanguages()" (click)="changeLanguage(lang)">{{lang.displayName}}</button>
+      </mat-menu>
+    </div>
+  `
+})
+export class LanguageSelectorComponent implements OnInit {
+
+  //TODO: Figure out if all of these fields are necessary
+  @Input() studyGuid: string;
+  @Input() isScrolled: boolean;
+  public loaded: boolean;
+  private studyLanguages: Array<StudyLanguage>;
+  private defaultLanguage: StudyLanguage;
+  private currentLanguage: StudyLanguage;
+  public isLoadingError = false;
+  public isNoDefaultFoundError = false;
+
+  constructor (
+    private serviceAgent: LanguageServiceAgent
+  ) { }
+
+  public ngOnInit(): void {
+    //TODO: Figure out if this can be simplified
+    //TODO: Do something with the error messages
+    //TODO: Handle case where exactly one language is configured
+    //TODO: Make sure we handle the case where no languages have been configured cleanly
+    //TODO: Handle case where user is logged in and we can get the language from the profile
+    //TODO: Store current language in session storage and check session storage in case we are returning from Auth0 or something
+    //TODO: Test navigating to different pages
+    this.serviceAgent.getConfiguredLanguages(this.studyGuid).subscribe(x => {
+      if (x) {
+        console.log('got configured languages: ' + JSON.stringify(x));
+        this.isLoadingError = false;
+        this.studyLanguages = x;
+        this.findDefaultLanguage();
+        if (this.currentLanguage == null) {
+          this.currentLanguage = this.defaultLanguage;
+        }
+        if (this.studyLanguages === null) {
+          this.isLoadingError = true;
+        }
+        else {
+          this.loaded = true;
+        }
+      } else {
+        this.isLoadingError = true;
+      }
+    });
+  }
+
+  private findDefaultLanguage(): void {
+    //TODO: Check if this can be simplified
+    this.isNoDefaultFoundError = false;
+    let lang: StudyLanguage = this.studyLanguages.find(element => element.isDefault === true);
+    if (lang !== undefined && lang !== null) {
+      this.defaultLanguage = lang;
+    } else {
+      this.isNoDefaultFoundError = true;
+    }
+  }
+
+  private getUnselectedLanguages(): Array<StudyLanguage> {
+    //TODO: Check if this can be simplified
+    if (this.studyLanguages !== null  && this.studyLanguages !== undefined) {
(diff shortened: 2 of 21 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
```diff
@@ -0,0 +1,102 @@
...
+      </mat-menu>
+    </div>
+  `
+})
+export class LanguageSelectorComponent implements OnInit {
+  @Input() studyGuid: string;
+  @Input() isScrolled: boolean;
+  public loaded: boolean;
+  private studyLanguages: Array<StudyLanguage>;
+  private currentLanguage: StudyLanguage;
```
fat23cat: Since we use `currentLanguage` in the template, it should be public property.

ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
```diff
@@ -0,0 +1,102 @@
...
+      </button>
+      
+      <mat-menu #menu="matMenu">
+        <button mat-menu-item *ngFor="let lang of getUnselectedLanguages()" (click)="changeLanguage(lang)">{{lang.displayName}}</button>
+      </mat-menu>
+    </div>
+  `
+})
+export class LanguageSelectorComponent implements OnInit {
+  @Input() studyGuid: string;
```
fat23cat: I think it is not necessary to provide `studyGuid` as Input, we can just inject ConfigurationService in the constructor like here: https://github.com/broadinstitute/ddp-angular/blob/DDP-4719_LanguageSelector/ddp-workspace/projects/ddp-sdk/src/lib/components/user/activities/userActivities.component.ts#L200

ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
```diff
@@ -0,0 +1,102 @@
...
+            this.loaded = true;
+          }
+        }
+      } else {
+        console.error('Error: no configured language list was returned.');
+      }
+    });
+  }
+
+  private getUnselectedLanguages(): Array<StudyLanguage> {
```
fat23cat: The same, since we use `getUnselectedLanguages` in the template, it should be a public method.

ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
```diff
@@ -0,0 +1,102 @@
...
+    if (this.studyLanguages !== null  && this.studyLanguages !== undefined) {
+      return this.studyLanguages.filter(elem => elem !== this.currentLanguage);
+    }
+   return null;
+  }
+
+  private changeLanguage(lang: StudyLanguage): void {
```
fat23cat: The same, since we use `changeLanguage` in the template, it should be a public method.

ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
```diff
@@ -0,0 +1,102 @@
...
+        console.error('Error: no configured language list was returned.');
+      }
+    });
+  }
+
+  private getUnselectedLanguages(): Array<StudyLanguage> {
+    if (this.studyLanguages !== null  && this.studyLanguages !== undefined) {
+      return this.studyLanguages.filter(elem => elem !== this.currentLanguage);
+    }
+   return null;
```
fat23cat, line 68: Maybe it would be better to return empty array?

ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  private changeLanguage(lang: StudyLanguage): void {
+    this.currentLanguage = lang;
+    this.translateService.use(lang.languageCode);
+    sessionStorage.setItem('studyLanguage', lang.languageCode);
```
fat23cat: We have `LanguageService` in the SDK(`ddp-angular\ddp-workspace\projects\ddp-sdk\src\lib\services\languageService.service.ts`), can we consider moving all code which change and save a language in the sessionStorage, in that service?
And one more thing - maybe it would be better to use localStorage instead of sessionStorage? It will allow us to save preferred language in case a user closed a browser.

ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
```diff
@@ -0,0 +1,102 @@
...
+import { StudyLanguage } from "../models/studyLanguage";
+import { LanguageServiceAgent } from "../services/serviceAgents/languageServiceAgent.service";
+import { isNullOrUndefined } from "util";
+import { TranslateService } from "@ngx-translate/core";
+
+@Component({
+  selector: 'ddp-language-selector',
+  template: `
+    <div [hidden]="!loaded">
+      <button class="SimpleButton" [ngClass]="{'SimpleButton--Scrolled': isScrolled}" *ngIf="currentLanguage !== null && currentLanguage !== undefined" [matMenuTriggerFor]="menu">
```
fat23cat, line 13: Can we use `isNullOrUndefined` function here?
```suggestion
      <button class="SimpleButton" [ngClass]="{'SimpleButton--Scrolled': isScrolled}" *ngIf="!isNullOrUndefined(currentLanguage)" [matMenuTriggerFor]="menu">
```

ddp-workspace/projects/ddp-angio/src/styles.scss
```diff
@@ -306,6 +306,35 @@ a:focus {
     color: mat-color($app-theme, 600);
 }
 
+.SimpleButton .ddp-globe {
+    stroke: mat-color($app-theme, 100);
+    fill: transparent;
+    vertical-align: middle;
+}
```
fat23cat: Can we add transition here and to other projects? Now animation looks a bit weird: text has smooth animation, but glode doesn't.
```suggestion
.SimpleButton .ddp-globe {
    stroke: mat-color($app-theme, 100);
    fill: transparent;
    vertical-align: middle;
    transition: stroke 0.3s ease;
}
```

ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
```diff
@@ -0,0 +1,102 @@
...
+        <svg class="ddp-globe" height="24px" width="24px" aria-live="title">
+          <title id="title" [lang]="currentLanguage.languageCode" translate>Toolkit.Header.LanguageSelection</title>
+          <use href="assets/images/globe.svg#Language-Selector-3"></use>
```
fat23cat: I think we can make icon URL configurable, we can add a new property in the ConfigurationService with the default value. And if someone wants to change the icon URL, he can do it under his project.

fat23cat (changes requested):

ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
charlottedye: I actually considered making this change.  I went back and forth on whether to make it, mostly because I was worried there was a case I had thought of earlier and that was why I had planned on doing session storage.  But I think local storage is the right call here.  In particular, apparently session storage, at least in Chrome, is tab-specific.

ddp-workspace/projects/ddp-angio/src/styles.scss
charlottedye: Thanks for identifying this!  I could tell there was something different but I thought the SVG just changed colors faster.  This makes more sense!

ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
charlottedye: Good point on all of these!  I think at some point I did some refactoring and forgot to go back and make sure the access modifiers were still correct.

MocanaAtBroad (commented):
Please check tslint warnings.

ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
charlottedye: Turns out Angular doesn't like that for some reason, but I found a different place where I was checking manually rather than calling the method and changed that location at least.

ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  constructor (
+    private serviceAgent: LanguageServiceAgent,
+    private language: LanguageService,
+    @Inject('ddp.config') private config: ConfigurationService
+  ) { }
+
+  public ngOnInit(): void {
+    this.iconURL = this.config.languageSelectorIconURL ? this.config.languageSelectorIconURL : "assets/images/globe.svg#Language-Selector-3";
+    this.serviceAgent.getConfiguredLanguages(this.config.studyGuid).subscribe(x => {
```
fat23cat: Please, take care of Observable subscription one of the following methods:
1. take(1) - https://github.com/broadinstitute/ddp-angular/blob/DDP-4719_LanguageSelector/ddp-workspace/projects/ddp-osteo/src/app/components/workflow-progress/workflow-progress.component.ts#L46
2. Save subscription and unsubscribe on destroying life cycle hook - https://github.com/broadinstitute/ddp-angular/blob/DDP-4719_LanguageSelector/ddp-workspace/projects/basil-app/src/app/loginLanding.component.ts#L37

fat23cat (dismissed):

fat23cat (approved):

ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
```diff
@@ -0,0 +1,111 @@
...
+import { isNullOrUndefined } from "util";
+import { Subscription } from "rxjs";
+
+@Component({
+  selector: 'ddp-language-selector',
+  template: `
+    <div [hidden]="!loaded">
+      <button class="SimpleButton" [ngClass]="{'SimpleButton--Scrolled': isScrolled}" *ngIf="currentLanguage !== null && currentLanguage !== undefined" [matMenuTriggerFor]="menu">
+        <svg class="ddp-globe" height="24px" width="24px">
+          <title id="title" [lang]="currentLanguage.languageCode" translate>Toolkit.Header.LanguageSelection</title>
```
fat23cat: One small issue: translate for SDK components we should store in SDK section in the en.json

fat23cat (changes requested):

ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
fat23cat: Sorry, I've noticed it too late.

charlottedye: Fixed it!

fat23cat (approved):

PR #157, opened 2020-06-16
DDP-4744 pass user timezone during signup

yufengwng:
When signing up new user, pass along their timezone to Auth0.

See https://github.com/broadinstitute/ddp-study-server/pull/262.

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
@@ -140,6 +140,7 @@ export class Auth0AdapterService implements OnDestroy {
             ...(additionalParams && {
                 ...additionalParams
             }),
+            time_zone: Intl.DateTimeFormat().resolvedOptions().timeZone || null,
             language: this.language.getCurrentLanguage(),
             // @todo : hack delete when done
             serverUrl: this.configuration.backendUrl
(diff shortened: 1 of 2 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
```diff
@@ -140,6 +140,7 @@ export class Auth0AdapterService implements OnDestroy {
             ...(additionalParams && {
                 ...additionalParams
             }),
+            time_zone: Intl.DateTimeFormat().resolvedOptions().timeZone || null,
```
yufengwng, line 143: Plumbing is in place to pass along timezone, but need help with how to actually get it...

fat23cat: I think we can use `Intl.DateTimeFormat`, it has good browsers support - 94%
https://caniuse.com/#search=Intl.DateTimeFormat
But if we want to cover all browsers, we can use moment timezone - https://momentjs.com/timezone/docs/#/using-timezones/guessing-user-timezone/, they try to guess timezone if a browser doesn't support `Intl.DateTimeFormat`

fat23cat (approved):

ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
MocanaAtBroad: Another option is to use a polyfill:
https://github.com/formatjs/date-time-format-timezone
Might be a lighter weight solution if we can figure out how to load it conditionally (which I believe Angular 8 makes possible, but more research needed). So only the (very few) browsers that don't support it will load it.
But I am inclined to punt on momentjs or polyfill and consider timezone "unknown" or default to NY time or something like that until we have a way for user to explicitly set it via UI

PR #171, opened 2020-06-23
Fix compilation problem in sandbox

MocanaAtBroad:
Missing tooltip property problem was showing up in compiler. Just removing that problem.


fat23cat (approved):

fat23cat:
Good catch!

PR #172, opened 2020-06-24
Added dispatch.yaml entries for dsm running on GAE

zyme:
As part of [DSM migration to GAE](https://docs.google.com/document/d/1FcM6eL8tYxJbUxVJP2PczgP22nXzBJg0VA1FzEzpYX8), we needed  to update our `dispatch.yaml`.

### DIFF
--- a/config/deployment/prod/dispatch.yaml
+++ b/config/deployment/prod/dispatch.yaml
@@ -13,3 +13,7 @@ dispatch:
     service: pepper-backend
   - url: "pepper.datadonationplatform.org/*"
     service: pepper-backend
+  - url: "dsm.datadonationplatform.org/api/*"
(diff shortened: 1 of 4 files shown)


config/deployment/prod/dispatch.yaml
```diff
@@ -13,3 +13,7 @@ dispatch:
     service: pepper-backend
   - url: "pepper.datadonationplatform.org/*"
     service: pepper-backend
+  - url: "dsm.datadonationplatform.org/api/*"
```
zyme, line 16: These dispatch files can be safely deployed even while DSM is on docker.  These won't take effect until the app engine instance is created and custom domains are configured.

MocanaAtBroad (approved):

PR #173, opened 2020-06-25
DDP-4736: Add participant invitation code to the dashboard

fat23cat:


ddp-workspace/projects/ddp-testboston/src/app/app-routes.ts
```diff
@@ -11,7 +11,7 @@ export const AppRoutes = {
     Consent: 'consent',
     CovidSurvey: 'covid-survey',
     Dashboard: 'dashboard',
-    CrcDashboard: 'crc-dashboard',
+    CrcDashboard: 'prism',
```
fat23cat, line 14: Per Jen's request:
> Can you make it /prism?

ddp-workspace/projects/ddp-testboston/src/app/components/crc-dashboard/crc-dashboard.component.ts
```diff
@@ -56,6 +57,7 @@ export class CrcDashboardComponent implements OnInit, OnDestroy {
 
   public setSelectedSubject(): void {
     this.sessionService.setParticipant(this.studySubject.userGuid);
+    this.config.invitationId = this.studySubject.invitationId;
```
fat23cat: New mockups assume that activity will show the selected subject's invitation id - https://drive.google.com/drive/u/1/folders/10QOZmiT1BYv2OI7F6Gpm19tEw_jeNT57
for this reason, we should store selected invitation id in the config.

yufengwng: I feel like the invite code should belong in the `session` instead of the config...

fat23cat: I also had this idea but declined it, so, since we both thought about it, I will implement that :)

MocanaAtBroad (approved):

PR #174, opened 2020-06-25
testboston: turn kit instructions download into link

yufengwng:


ddp-workspace/projects/ddp-testboston/src/app/components/footer/footer.component.html
```diff
@@ -55,7 +55,7 @@
                             routerLinkActive="footer-link_active" translate>Navigation.Invitation</a>
                     </li>
                     <li>
-                        <a class="Link footer-link" href="kit-instructions.pdf" download translate>
+                        <a class="Link footer-link" href="kit-instructions.pdf" translate>
```
yufengwng: Do we need a leading `/` in the `href`?

fat23cat: No, we don't need that.

I think it would be better if we show pdf in the new tab:
```suggestion
                        <a class="Link footer-link" href="kit-instructions.pdf" target="_blank" translate>
```

fat23cat (approved):

PR #186, opened 2020-07-08
DDP-4771: Completed activities have Continue button instead of Edit

fat23cat:


ddp-workspace/projects/ddp-testboston/src/assets/i18n/en.json
```diff
@@ -333,6 +333,7 @@
         "PreviousButton": "Back",
         "SavingButton": "Saving",
         "EditButton": "Continue",
+        "CompleteButton": "Edit",
         "ReviewButton": "View",
```
MocanaAtBroad, line 337: Looks like when the status of the activity is COMPLETE we show the "Edit" as the button label. My reading of ticket was that it would say Continue. Can you double-check?

fat23cat: Hm, desired behavior section says: 
`
Change button text from Continue to Edit once activities are completed/submitted but still able to be modified.
`

MocanaAtBroad: OK. I misread thinking it was a feature request rather than a bug report. Looks good then!

MocanaAtBroad (approved):

PR #189, opened 2020-07-10
DDP-4798: Setup new pediatric BrainCancerProject Angular app

fat23cat:
Erin presented new BrainCancerProject design - https://projects.invisionapp.com/d/main/#/projects/prototypes/20209715
The new design is radically different from the old one, so, IMO, it would be easier to create new Angular app and new study, because the study also contains styles and new design assumes new count-me-in page. This will make for easier development, and keep open the possibility for concurrent changes to what we currently have for the brain.
And when redesign done, we can just remove the old app and the study, and then rename and use the new app and the study.

I added many reviewers, I'm not expecting real code review, I just want to discuss the idea, maybe someone can suggest something else.


config/pepperConfig.js.ctmpl
```diff
@@ -95,6 +95,22 @@ var DDP_ENV = {
     {{end}}
 {{end}}
 
+{{if eq $study_guid "cmi-brain-v2"}}
```
fat23cat: New study guid.

ddp-workspace/angular.json
```diff
@@ -1155,6 +1155,148 @@
           }
         }
       }
+    },
+    "ddp-brain-v2": {
```
fat23cat, line 1159: New Angular app. `ddp-brain` is still alive.

config/pepperConfig.js.ctmpl
yufengwng: Creating new app on frontend is a good idea. Creating new study on backend will be bad... A lot of backend configuration, users, and data is tied to the `cmi-brain` study. Moving it to a new study `cmi-brain-v2` is like doing a whole study migration...

fat23cat: Hmm, can we create custom activities in the current study scope and return them only for specific test users?

yufengwng: Might be a bit tricky but certainly do-able.

MocanaAtBroad (approved):
Let's get it into develop!

PR #194, opened 2020-07-20
Prion: DDP-4806, persist selected language when participant signs up …

charlottedye:
…or logs in: Porting over changes from DDP-4436

This corresponds to DDP-4806.

This ticket does the following:

- When language is changed using the language selector for an authenticated user, updates the language in the profile and reloads components that get text from the backend so that everything gets translated immediately
- When a user signs in, the language will be retrieved from their profile
- When the language selector loads, it will first check for a profile language if the user authenticated.  Previously, the plan was to first check for a locally stored language, but that caused issues when a user logged into an account that had a different language set.  This way, if an unauthenticated user changes the language and then logs into an account with a different language in the profile, the profile language will be restored.  

This required changing a number of SDK components to allow components to reload text retrieved from the backend when the language was changed.  There are some components that are not used by Prion.  I believe I handled this correctly for those components, but it is possible I missed some.

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
@@ -121,7 +121,7 @@ export class LanguageSelectorComponent implements OnInit, OnDestroy {
   }
 
   //Find the current language and return true if successful or false otherwise
-  public findCurrentLanguage(): Observable<Boolean> {
+  public findCurrentLanguage(): Observable<boolean> {
     //Check for a language in the profile
     let profileLangObservable: Observable<StudyLanguage> = this.getProfileLangObservable();
 
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
@@ -1,10 +1,15 @@
 import { Component, Inject, Input, OnDestroy, OnInit, Output, EventEmitter } from '@angular/core';
+import { CompositeDisposable} from '../compositeDisposable';
+import { UserProfile } from '../models/userProfile';
 import { StudyLanguage } from '../models/studyLanguage';
 import { ConfigurationService } from '../services/configuration.service';
 import { LanguageService } from '../services/languageService.service';
+import { SessionMementoService } from '../services/sessionMemento.service';
+import { UserProfileServiceAgent} from '../services/serviceAgents/userProfileServiceAgent.service';
 import { LanguageServiceAgent } from '../services/serviceAgents/languageServiceAgent.service';
 import { isNullOrUndefined } from 'util';
-import { Subscription } from 'rxjs';
+import { iif, Observable, of, Subscription } from 'rxjs';
+import { concatMap, flatMap, map, take } from 'rxjs/operators';
 
 @Component({
   selector: 'ddp-language-selector',
@@ -32,25 +37,29 @@ export class LanguageSelectorComponent implements OnInit, OnDestroy {
   public currentLanguage: StudyLanguage;
   public iconURL: string;
   private studyLanguages: StudyLanguage[];
-  private anchor: Subscription;
+  private anchor: CompositeDisposable;
 
   constructor(
     private serviceAgent: LanguageServiceAgent,
     private language: LanguageService,
-    @Inject('ddp.config') private config: ConfigurationService) { }
+    private profileServiceAgent: UserProfileServiceAgent,
+    @Inject('ddp.config') private config: ConfigurationService,
+    private session: SessionMementoService) { }
 
   public ngOnInit(): void {
+    this.anchor = new CompositeDisposable();
     this.iconURL = this.config.languageSelectorIconURL ? this.config.languageSelectorIconURL : "assets/images/globe.svg#Language-Selector-3";
-    this.anchor = this.serviceAgent.getConfiguredLanguages(this.config.studyGuid).subscribe(x => {
+    let sub: Subscription = this.serviceAgent.getConfiguredLanguages(this.config.studyGuid).subscribe(x => {
       if (x) {
         //Only use language selector if multiple languages are configured!
         if (x.length > 1) {
           this.studyLanguages = x;
           this.language.addLanguages(this.studyLanguages.map(x => x.languageCode));
-          if (this.findCurrentLanguage()) {
-            this.loaded = true;
-            this.isVisible.emit(true);
-          }
+          let sub2: Subscription = this.findCurrentLanguage().subscribe(x => {
+            this.loaded = x.valueOf();
+            this.isVisible.emit(x.valueOf());
+          });
+          this.anchor.addNew(sub2);
         } else {
           this.isVisible.emit(false);
         }
@@ -59,10 +68,13 @@ export class LanguageSelectorComponent implements OnInit, OnDestroy {
         this.isVisible.emit(false);
       }
     });
+    this.anchor.addNew(sub);
   }
 
   public ngOnDestroy(): void {
-    this.anchor.unsubscribe();
+    if (this.anchor) {
+      this.anchor.removeAll();
+    }
   }
 
   public getUnselectedLanguages(): Array<StudyLanguage> {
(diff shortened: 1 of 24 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
```diff
@@ -73,43 +85,134 @@ export class LanguageSelectorComponent implements OnInit, OnDestroy {
...
+  }
+
   //Find the current language and return true if successful or false otherwise
-  public findCurrentLanguage(): boolean {
-    //Check if we already have a current language
-    if (!isNullOrUndefined(this.currentLanguage)) {
-      this.changeLanguage(this.currentLanguage);
-      return true;
-    }
+  public findCurrentLanguage(): Observable<Boolean> {
```
fat23cat: We should use here type instead of class.
```suggestion
  public findCurrentLanguage(): Observable<boolean> {
```

ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
```diff
@@ -32,25 +37,29 @@ export class LanguageSelectorComponent implements OnInit, OnDestroy {
...
 
   constructor(
     private serviceAgent: LanguageServiceAgent,
     private language: LanguageService,
-    @Inject('ddp.config') private config: ConfigurationService) { }
+    private profileServiceAgent: UserProfileServiceAgent,
+    @Inject('ddp.config') private config: ConfigurationService,
+    private session: SessionMementoService) { }
 
   public ngOnInit(): void {
```
fat23cat, line 50: Please, consider refactoring here - subscribe inside of other subscribe is a bad practice. My suggestion here:
``` TS
  public ngOnInit(): void {
    this.anchor = new CompositeDisposable();
    this.iconURL = this.config.languageSelectorIconURL ? this.config.languageSelectorIconURL : "assets/images/globe.svg#Language-Selector-3";
    const sub = this.serviceAgent.getConfiguredLanguages(this.config.studyGuid).pipe(
      mergeMap(studyLanguages => {
        if (studyLanguages) {
          // Only use language selector if multiple languages are configured!
          if (studyLanguages.length > 1) {
            this.studyLanguages = studyLanguages;
            this.language.addLanguages(this.studyLanguages.map(language => language.languageCode));
            return this.findCurrentLanguage();
          } else {
            return of(false);
          }
        } else {
          console.error('Error: no configured language list was returned.');
          return of(false);
        }
      })
    ).subscribe(value => {
      this.loaded = value;
      this.isVisible.emit(value);
    });
    this.anchor.addNew(sub);
  }
```

ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
```diff
@@ -73,43 +85,134 @@ export class LanguageSelectorComponent implements OnInit, OnDestroy {
...
+    //Create an observable that will check each applicable option and return the first valid language found, if any
+    let langObservable: Observable<StudyLanguage> = profileLangObservable.pipe(
+      flatMap(x => {
+        return this.getNextObservable(x, currentStoredLangObservable);
+      })
+    ).pipe(
+      flatMap(y => {
+        return this.getNextObservable(y, defaultLangObservable);
+      })
+    );
```
fat23cat: Can we make it in one pipe operator?
```suggestion
    let langObservable: Observable<StudyLanguage> = profileLangObservable.pipe(
      flatMap(x => {
        return this.getNextObservable(x, currentStoredLangObservable);
      }),
      flatMap(y => {
        return this.getNextObservable(y, defaultLangObservable);
      })
    );
```

ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
```diff
@@ -59,10 +68,13 @@ export class LanguageSelectorComponent implements OnInit, OnDestroy {
...
       }
     });
+    this.anchor.addNew(sub);
   }
 
   public ngOnDestroy(): void {
-    this.anchor.unsubscribe();
+    if (this.anchor) {
+      this.anchor.removeAll();
+    }
```
fat23cat: If we use `CompositeDisposable` we don't need `if(...) {...}` here. This logic encapsulated in the `CompositeDisposable` class.

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/sessionServiceAgent.service.ts
```diff
@@ -47,4 +49,15 @@ export class SessionServiceAgent<TEntity> extends ServiceAgent<TEntity> {
...
+    protected getObservable(path: string, options: any = {}, unrecoverableStatuses: Array<number> = []): Observable<TEntity | null> {
+      if (this.language) {
+        return this.language.getProfileLanguageUpdateNotifier()
+          .pipe(switchMap(() =>
+            super.getObservable(path, options, unrecoverableStatuses)));
+      }
+      else {
+        return super.getObservable(path, options, unrecoverableStatuses);
+      }
+    }
```
fat23cat: Could you please clarify why do we need to override this method?

fat23cat (changes requested):
Overall looks good! Left some suggestions and questions.
Probably it's out of this ticket scope but poped up one issue. When we have two language selectors, they are not synchronized: if we changed one, another one stays with the previous language:
image

ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
charlottedye: Good point.  I forgot that TypeScript doesn't require you to use a class for generics like Java does...

ddp-workspace/projects/ddp-sdk/src/lib/components/languageSelector.component.ts
```diff
@@ -73,43 +85,133 @@ export class LanguageSelectorComponent implements OnInit, OnDestroy {
...
+  private updateProfileLanguage(): void {
+    // Create observable that gets the user's profile and creates a copy with updated preferred language
+    let getProfileObservable: Observable<UserProfile> = this.profileServiceAgent.profile
+      .pipe(map(x => {
+        return {...x.profile, preferredLanguage: this.currentLanguage.languageCode};
+      })).pipe(take(1));
+
+    // When we get the new profile value, save it
+    let updateProfileObservable: Observable<any> = getProfileObservable
+      .pipe(concatMap(profile => this.profileServiceAgent.saveProfile(false, profile)));
```
yufengwng: Since this `saveProfile()` call is really just doing a `PATCH` call, you can make API request with just the new language code. This way, you don't need to do a fetch beforehand.

charlottedye:
The code I just pushed has some bugs, so not ready for re-review yet...

charlottedye:
> Overall looks good! Left some suggestions and questions.
> Probably it's out of this ticket scope but poped up one issue. When we have two language selectors, they are not synchronized: if we changed one, another one stays with the previous language:
> image

Forgot to respond to this one: that's definitely a case I didn't test.  Is that new with this ticket or was that the case before, too?  That doesn't particularly surprise me when I think about it, though, because nothing would tell the language selector to update the displayed language.  I'd have to think about how to handle that without causing more problems.  Nothing comes to mind immediately as an obvious way of solving this issue.  The trick is that the notification should only happen when a language selector that is not the current language selector changed the language.

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/userProfileServiceAgent.service.ts
```diff
@@ -37,14 +38,27 @@ export class UserProfileServiceAgent extends UserServiceAgent<UserProfile> {
...
+          profile.preferredLanguage = 'en';
+        }
         if (isNew) {
             return this.postObservable('/profile', JSON.stringify(profile));
         } else {
             return this.patchObservable('/profile', JSON.stringify(profile));
         }
     }
 
+    public saveProfileAttributes(profile: UserProfile): Observable<any> {
```
fat23cat: Maybe call this something like `updateProfile`?
```suggestion
    public updateProfile(profile: UserProfile): Observable<any> {
```

fat23cat (approved):

PR #200, opened 2020-07-22
TestBoston: PDF i18n

fat23cat:
For PDF i18n support, we should store PDF in different folders.


ddp-workspace/projects/ddp-testboston/src/app/components/footer/footer.component.html
```diff
@@ -55,7 +55,8 @@
                             routerLinkActive="footer-link_active" translate>Navigation.Invitation</a>
                     </li>
                     <li>
-                        <a class="Link footer-link" href="kit-instructions.pdf" target="_blank" translate>
+                        <a class="Link footer-link" [href]="language + '/kit-instructions.pdf'" target="_blank"
```
fat23cat, line 58: Should we internationalize PDFs' names?

yufengwng: I think it's fine -- user can save it with whatever name they want.

yufengwng (approved):

PR #213, opened 2020-08-03
DDP-4834: [TestBoston] Allow CRC to create login account for participant post-enrollment

fat23cat:
Description - https://broadinstitute.atlassian.net/browse/DDP-4834

### DIFF
--- a/ddp-workspace/projects/ddp-testboston/src/app/components/enrollment/enrollment.component.ts
+++ b/ddp-workspace/projects/ddp-testboston/src/app/components/enrollment/enrollment.component.ts
@@ -5,12 +5,13 @@ import {
   SessionMementoService,
   SubjectInvitationServiceAgent,
   UserProfile,
+  UserProfileDecorator,
   UserProfileServiceAgent,
   WorkflowServiceAgent
 } from 'ddp-sdk';
 import { WorkflowBuilderService } from 'toolkit';
 import { AppRoutes } from '../../app-routes';
-import { of } from 'rxjs';
+import { of, throwError, Observable } from 'rxjs';
 import { take, tap, mergeMap } from 'rxjs/operators';
 
 @Component({
@@ -23,6 +24,7 @@ export class EnrollmentComponent implements OnInit {
   public accountForm: FormGroup;
   public isLoading = false;
   public showError = false;
+  private profile: UserProfile;
 
   constructor(
     private router: Router,
@@ -35,8 +37,7 @@ export class EnrollmentComponent implements OnInit {
 
   public ngOnInit(): void {
     this.initAccountForm();
-    this.showError = (this.session.session.participantGuid === null && this.session.session.invitationId === null) ||
-      (this.session.session.participantGuid !== null);
+    this.setupInitialState();
   }
 
   public onSubmit(): void {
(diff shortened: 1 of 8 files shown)


ddp-workspace/projects/ddp-testboston/src/app/components/header/header.component.html
```diff
@@ -59,7 +60,7 @@
     </div>
 </header>
 
-<div #overlay class="overlay">
+<div *ngIf="!isAdmin" #overlay class="overlay">
```
fat23cat, line 63: Found one issue - we also need to hide overlay panel for admins. Otherwise, it appeared in the DOM but was hidden by display: none.

ddp-workspace/projects/ddp-testboston/src/assets/i18n/es.json
```diff
@@ -241,7 +241,11 @@
   "Prism": {
     "ClearSearch": "Borrar",
     "Email": "DIRECCIÓN DE CORREO ELECTRÓNICO DEL PARTICIPANTE VOLUNTARIO",
-    "NoEmail": "NO APLICA",
+    "NoneListed": {
+      "Title": "None listed",
+      "Text": "If the participant wants to bo able to login to the TestBoston site, please enable their TestBoston account",
+      "Link": "here"
+    },
```
fat23cat, line 248: We don't need to translate it, because CRC UI should be in English constantly.

ddp-workspace/projects/ddp-testboston/src/app/components/enrollment/enrollment.component.ts
```diff
@@ -85,4 +87,50 @@ export class EnrollmentComponent implements OnInit {
...
+  private isProfileChanged(profile: UserProfile): boolean {
+    return this.profile.firstName !== profile.firstName || this.profile.lastName !== profile.lastName;
+  }
+
+  private setupInitialState(): void {
+    const invitationId$ = of(this.session.session.invitationId);
+
+    invitationId$.pipe(
+      take(1),
+      mergeMap(invitationId => {
```
MocanaAtBroad, line 106: Could this be simplified, dropping the creation of invitationId$ and just create the pipeline from `this.subjectInvitation(invitationId)`?

ddp-workspace/projects/ddp-testboston/src/assets/i18n/en.json
```diff
@@ -268,7 +272,7 @@
     }
   },
   "Enrollment": {
-    "Title": "Account setup",
+    "Title": "Account set-up",
```
MocanaAtBroad, line 275: The hyphen in `set-up` is unusual in American English as far I know. Original looks more correct to me.

MocanaAtBroad (commented):
Missing semicolon in  line 10 app-routing.module.ts

Still looking

ddp-workspace/projects/ddp-testboston/src/assets/i18n/en.json
fat23cat: I changed that per Jen's request:
image
image

ddp-workspace/projects/ddp-testboston/src/app/components/enrollment/enrollment.component.ts
fat23cat: Hm, no, I don't think so: I did it in this way, because CRCs can open testboston.org/enrolement directly, without selecting a study subject. So the component will try to get `invitationId` from the session if `invitationId` exists, will send a `lookupInvitation` request, otherwise, observable will throw an error to prevent the next requests and show the error message.

MocanaAtBroad (approved):

PR #215, opened 2020-08-05
DDP-4564: Manage cookie preferences

marat-mamytov:
Current PR adds ability to manage cookies preferences.

**AC1, AC2**
When participant clicks on “Manage Preferences” in the cookie banner 
image
the modal “Manage Cookie Preferences” is displayed
image

**AC3**
There are 3 tabs:  “Your Privacy”, “Functional Cookies”, “Analytical Cookies”
For example:
image

But any additional categories of cookies can be added to the `type CookiesTypes` and listed in the project configuration.

**AC4**
Clicking on the “Submit Preferences” button -> submits the accepted or rejected cookies -> user preferences set in the Local Storage under `prion_cookies_consent` cookie + dismisses the modal and the underlying cookie banner
image

**AC5**
Click on the “X” -> dismiss the modal, not changes apply

**AC6**
Click on the “Privacy Policy” link -> launch the Privacy Policy page in the modal
image

**AC7**
The cookie preferences selected should persist across sessions since set in the Local Storage

**AC8**
On mobile view the modal takes up the whole view port
image

**AC9**
Added  a button to launch this cookie preference panel in the Privacy Policy page
image

**AC10**
Added logger to the cookies consent update event


charlottedye:
Question on this: 

"For authenticated uses cookies preferences are saved into profile:
New field public cookiesPreferences: CookiesPreferences | null; was added to the UserProfile model.
Need backend support for storing this data."

  I thought we decided not to do this.

ddp-workspace/projects/ddp-sdk/src/lib/models/userProfile.ts
```diff
@@ -7,4 +9,5 @@ export class UserProfile {
     public preferredLanguage: string | null;
     public firstName: string | null;
     public lastName: string | null;
+    public cookiesPreferences: CookiesPreferences | null;
```
charlottedye: I thought we decided not to do the profile support.  If we are doing this, we probably need to fully implement it.

charlottedye (changes requested):
I ran the code locally to review the functionality before looking closely at the actual code and found a number of issues: go ahead and fix these and then re-test and check the acceptance criteria to make sure I didn't miss anything.  Also, I thought we decided not to put the cookie preferences in the profile, so we should make sure we're on the same page about the plan for that.

-_gat and _gat_platform sometimes get added before cookies are accepted or rejected
-When the user submits the cookie preferences banner without changing anything (I.e., when they select accept on analytics cookies), the _gid and _ga cookies are added immediately like they should be, but _gat_platform and _gat are added only after doing something like clicking on the header logo.
-Functional cookies and analytical categories are described differently.  Functional cookies are grouped by service and analytical cookies are not.  It would be better to either always group or never group.  At minimum, "Name" should probably mean the same thing for both categories.
-For functional cookies: It seems like we should include the actual name of the cookie here somewhere.  For example, PRION_cookies_consent, tcell_agent_policy_cache, tcell_agent_session_id, etc.  For example, maybe under name we could do something like: "TCell: tcell_agent_policy_cache, tcell_agent_session_id".
-For functional cookies: some groups have expiration listed as "various".  It seems like it would be better to specify the different expirations, either by specifying the expiration for each cookie in the category or just listing the possible expirations.  We could say things in local storage have an expiration of "none"; keep "session" for session storage; and list the actual time before expiration for actual cookies
-For functional cookies: Auth0, TCell, and Pepper platform should probably be capitalized 
-For functional cookies: the descriptions aren't really accurate for some of the categories.  User authorization makes sense for Auth0, although you could also describe it as user authentication or access control.  For Pepper platform, I don't recall what exactly the cookies are (pepper.PRION.irbsession will only be present in test environments so you don't need to include that.  You probably should include PRION_cookies_consent, and that would be for storing cookie preferences).  For TCell, description should say something about security, like website security or just security.
-For analytical cookies: It looks like there's also _gat_platform
-For analytical cookies: the description for _gat should probably indicate that this is for network calls to analytics services.  Maybe something like "throttling to increase the efficiency of analytics collection" (or something shorter if you can think of anything)
-On the welcome page, the join us button at the bottom looks different.  Seems like there was a change to the CSS for Button—primaryYellow.  If something using Button—primaryYellow needs to change, it should do that in a different CSS class

ddp-workspace/projects/ddp-sdk/src/lib/models/userProfile.ts
marat-mamytov: Removed saving to user's profile

charlottedye:
I thought I had mentioned this but don't see it in my comment, so just in case I left it out: there are at least two missing cookies not described here related to Google Analytics that appear after accepting or rejecting cookies and clicking on the Prion banner at the top.

marat-mamytov:
> I thought I had mentioned this but don't see it in my comment, so just in case I left it out: there are at least two missing cookies not described here related to Google Analytics that appear after accepting or rejecting cookies and clicking on the Prion banner at the top.

 @charlottedye 
+ I've added a few Pepper cookies stored in local storage: session_key, token, studyLanguage. They are set in SessionMementoService and LanguageService. 
+ Updated the cookies description and table in the cookie preferences modal.
+ Added manage preferences button to the policy page.
+ Fixed Button—primaryYellow styles

_Please check updates if you'll have a chance, how it looks like._

But unfortunately,  I'm not getting _gat and _gat_platform added before cookies are accepted or rejected -> I'll investigate it more. 
And I'm not getting any additional GA cookies (at least two as you mentioned). 

_Could you please remove all cookies and local storage data for prion sites and local storage and check one more time.
If they remain, write their names, I'll investigate it also._

charlottedye:
I recall at least one case where they showed up was after clicking accept
on the cookie banner and then clicking on the Prion logo in the header.
Try clearing cookies and storage and going to the page, clicking accept,
and then clicking on the Prion logo and see if those show up.  If not, I
can see if it's a configuration issue.  Then make sure things work and you
aren't seeing any cookies you aren't expecting when creating an account,
filling out a survey, logging out, and logging in, etc.  You could also try
searching source control for "_gat", "_gat_platform", and anything related
to analytics.  If you're still not seeing it after that, I can investigate
and that would probably mean either it's a configuration issue or it was
inadvertently fixed somehow.

marat-mamytov:
@charlottedye 
> Also, I thought we decided not to put the cookie preferences in the profile, so we should make sure we're on the same page about the plan for that.

Misunderstood, fixed, no saving into profile.

> -_gat and _gat_platform sometimes get added before cookies are accepted or rejected
> -When the user submits the cookie preferences banner without changing anything (I.e., when they select accept on analytics cookies), the _gid and _ga cookies are added immediately like they should be, but _gat_platform and _gat are added only after doing something like clicking on the header logo.

Fixed my local issue with _gat cookie. Thank you! =>  Now both _gat and _gat_platform are set **with** _ga and _gid on acceptance. And no _gat and _gat_platform cookies are added before consent.
image

> -On the welcome page, the join us button at the bottom looks different. Seems like there was a change to the CSS for Button—primaryYellow. If something using Button—primaryYellow needs to change, it should do that in a different CSS class

Fixed.

Also fixed cookies list and descriptions.

charlottedye (changes requested):
I tried to rebase this on the latest version of the Prion branch multiple times but it does not seem to be working.  It might be better if you do that or even start fresh with a new branch based on prion_frontend_migration_branch and put all your changes in one commit since this is all one ticket anyway.

Additionally, I ran the non-rebased version of this locally and there are some basic UI things in the cookie preferences window that appear to have been messed up at some point.  I'm assuming the extra space under the button and link at the bottom of the modal window for the your privacy tab was unintentional because it looks a little strange to me.  At at least one screen size (I noticed this at 1121px x 721px), the bottom section with the link and button starts at a different place on the modal on each tab.  Looks like I might have missed this on the initial review.  Of the three tab versions, I think the analytical cookies tab's version looks the best, personally, but pick something that looks reasonable and use it for all three.  

Nice work finding the additional cookies.

PR #221, opened 2020-08-11
DDP-4841: BrainCancerProject pediatric Frequently Asked Questions page

fat23cat:
Description - https://broadinstitute.atlassian.net/browse/DDP-4841


MocanaAtBroad (commented):
It looks good both on desktop and mobile displays!
A question about style that had not occurred to me before (though I think this is more of an Erin thing).
The **+** symbols that expand the answers to the questions. Isn't the more common convention to convert the **+** to **-** when expanded? (I am aware we do the same thing in Osteo, but first time I had thought about it).

MocanaAtBroad (approved):

PR #232, opened 2020-08-18
[TestBoston] Single kit instructions pdf for all languages

fat23cat:
I have not replaced pdf with the real one, because it's not the final version.


yufengwng (approved):

ddp-workspace/projects/ddp-testboston/src/style/session-expired.scss
```diff
@@ -1,5 +1,5 @@
 .session-content-section {
-    margin: 5rem 0 0 0;
+    margin: 5rem 0 15rem 0;
```
fat23cat, line 2: I bit updated the margin for the session expired page. I noticed, that content is too compressed

PR #236, opened 2020-08-19
DDP-4853 support marking phone required in address component

yufengwng:
We added new `requirePhone` flag, so this change is to support reading that flag and marking phone field as required. This ONLY marks it as required, but doesn't seem to check it is indeed filled in. I think it will be checked by backend once Marco makes his change related to that. Please let me know if I missed something here.

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressInput.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressInput.component.ts
@@ -1,4 +1,5 @@
 import {
+  ChangeDetectorRef,
   ChangeDetectionStrategy,
   Component,
   ElementRef,
@@ -20,6 +21,7 @@ import * as _ from 'underscore';
 import { mergeMap, take, takeUntil, tap } from 'rxjs/operators';
 import { AddressInputService } from '../address/addressInput.service';
 import { NGXTranslateService } from '../../services/internationalization/ngxTranslate.service';
+import { AddressService } from '../../services/address.service';
 
 @Component({
   selector: 'ddp-address-input',
@@ -147,7 +149,7 @@ import { NGXTranslateService } from '../../services/internationalization/ngxTran
       padding: 0;
       margin:0;
     }`],
-  providers: [AddressInputService],
+  providers: [],
   changeDetection: ChangeDetectionStrategy.OnPush
 })
 
@@ -182,7 +184,7 @@ export class AddressInputComponent implements OnInit, OnDestroy {
   }
 
   @Input()
-  phoneRequired: boolean = false;
+  phoneRequired = false;
 
   /**
    * Will emit event with address as it changes in form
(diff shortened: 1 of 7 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressInput.component.ts
```diff
@@ -179,6 +180,10 @@ export class AddressInputComponent implements OnInit, OnDestroy {
   set addressErrors(addressErrors: AddressError[]) {
     this.displayVerificationErrors(addressErrors);
   }
+
+  @Input()
+  phoneRequired: boolean = false;
```
MocanaAtBroad: tslint warning that type is redundant. the `false` assignment enough to know it is boolean.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressInput.component.ts
```diff
@@ -179,6 +180,10 @@ export class AddressInputComponent implements OnInit, OnDestroy {
   set addressErrors(addressErrors: AddressError[]) {
     this.displayVerificationErrors(addressErrors);
   }
+
```
MocanaAtBroad, line 186: To make the `required` flag stick you will need to go a bit further and add the required validation to the form group (referenced in line 27 in this file and instantiated here: https://github.com/broadinstitute/ddp-angular/blob/2512fef4e0e0cdb25d1445cefb260043d7b786ce/ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressInput.service.ts#L54

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/MailAddressBlock.ts
```diff
@@ -5,6 +5,8 @@ export class MailAddressBlock extends ActivityBlock {
     public displayNumber: number | null;
     public titleText: string | null;
     public subtitleText: string | null;
+    public requireVerified: boolean;
```
MocanaAtBroad, line 8: This is not doing anything, right?

yufengwng: We'll need to add it sooner or later right? I mean I can remove it here and someone can add it in later.

MocanaAtBroad: It's fine. Just checking that not an oversight or incomplete.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressInput.component.ts
```diff
@@ -146,7 +149,7 @@ import { NGXTranslateService } from '../../services/internationalization/ngxTran
       padding: 0;
       margin:0;
     }`],
-  providers: [AddressInputService],
+  providers: [],
```
MocanaAtBroad: You can just remove this now.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressInput.service.ts
```diff
@@ -334,7 +336,7 @@ export class AddressInputService implements OnDestroy {
       state: new FormControl(''),
       city: new FormControl('', Validators.required),
       phone: new FormControl(''),
-      guid: new FormControl('')
+      guid: phoneRequired ? new FormControl('', Validators.required) : new FormControl(''),
```
MocanaAtBroad: You might have applied the validation to wrong the field.

yufengwng: d'oh! But that points out something I don't understand: it seemed fine before even if I didn't push required down to this service.

MocanaAtBroad:
Also check the tests. Looks like we have compilation problem with one of them.
To run, from command line:
`ng test ddp-sdk --browsers Chrome`
Let me know if tests become difficult to deal with. They are a royal pain.

MocanaAtBroad (approved):

PR #239, opened 2020-08-19
DDP-4894 use different redirect item for admin login

yufengwng:
Fixes DDP-4894.

### Issue
Issue is that both regular login and admin login is using the `nextUrl` session storage item. So after navigating to `/prism` the session storage item is populated, then when you back out and do a regular login, `login-landing` component will pick up on `nextUrl` and redirect to `/prism`.

### Solution
We'll use a different session item call `adminNextUrl`, so admin login flow is more isolated.

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
@@ -114,6 +114,7 @@ export class Auth0AdapterService implements OnDestroy {
             ...(additionalAuth0QueryParams && additionalAuth0QueryParams)
         };
         if (this.configuration.doLocalRegistration) {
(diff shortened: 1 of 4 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
```diff
@@ -114,6 +114,7 @@ export class Auth0AdapterService implements OnDestroy {
             ...(additionalAuth0QueryParams && additionalAuth0QueryParams)
         };
         if (this.configuration.doLocalRegistration) {
+            sessionStorage.setItem('localAdminAuth', 'false');
```
yufengwng, line 117: This also helps to ensure in local registration we're doing clean logins instead of going down the admin path.

fat23cat (approved):

PR #245, opened 2020-08-21
DDP-4799: Third party vendor list page

marat-mamytov:
_Recreated branch and PR due to circular rebase errors_

Current PR adds third party vendor list page:
image


charlottedye (approved):
Looks good!

PR #248, opened 2020-08-21
testboston: add GA token

yufengwng:
See DDP-4882.

This adds the production GA Tracking ID for TestBoston. This is also saved in vault under `secret/pepper/prod/v1/testboston/conf`. Are there other places I need to update?


fat23cat (approved):
Everything is good! pepperConfig is only one place where we need to add projectGAToken.

PR #254, opened 2020-08-27
DDP-4898: [TestBoston] Help page

fat23cat:
Mockup:
image


ddp-workspace/projects/ddp-testboston/src/app/components/header/header.component.html
```diff
@@ -1,6 +1,6 @@
 <header *ngIf="!isAdmin" class="header" [ngClass]="{'header_shadow': isPageScrolled && !isMenuOpened}">
     <div class="header-content">
-        <a class="project-logo" routerLink="/">
+        <a class="project-logo" routerLink="/" (click)="closeMenu()">
```
MocanaAtBroad, line 3: Good idea!

MocanaAtBroad (commented):
A couple of suggestions:

MocanaAtBroad (commented):
I think would be better if this link opens in separate tab so our page can be used as reference:
Screen Shot 2020-09-08 at 5 24 46 PM

MocanaAtBroad (commented):
Would be better if the text "location type" was capitalized and bolded so it echoes the text in the picture.
Also would be nice if we had the search icon Screen Shot 2020-09-08 at 5 32 23 PM in the instructions text.


Screen Shot 2020-09-08 at 5 20 18 PM

fat23cat:
> I think would be better if this link opens in separate tab so our page can be used as reference:

Hm, I've double-checked, when I click "www.ups.com/dropoff" it opens in the separate tab.

> Would be better if the text "location type" was capitalized and bolded so it echoes the text in the picture.

Good catch, thanks!

> Also would be nice if we had the search icon Screen Shot 2020-09-08 at 5 32 23 PM in the instructions text.

Nice suggestion, but I have to discuss it with Jen and Erin.

fat23cat:
> Would be better if the text "location type" was capitalized and bolded so it echoes the text in the picture.

Good catch, thanks!

fat23cat:
> Also would be nice if we had the search icon Screen Shot 2020-09-08 at 5 32 23 PM in the instructions text.

Nice suggestion, but I need to discuss it with Jen and Erin.

MocanaAtBroad (approved):

PR #257, opened 2020-08-31
DDP-4850: BrainCancerProject pediatric Data page

fat23cat:
Mockups - https://projects.invisionapp.com/d/main?origin=v7#/console/20209715/423517818/preview?scrollOffset=0


MocanaAtBroad (commented):
Looks good! Only visible difference from invision design seems to be the text in the grey box (with the US map). Text seems like 1 or 2 points smaller to my eye. This is nitpicking as it looks very nice as is.

fat23cat:
> Looks good! Only visible difference from invision design seems to be the text in the grey box (with the US map). Text seems like 1 or 2 points smaller to my eye. This is nitpicking as it looks very nice as is.

I've double-checked font sizes - everything seems to be correct. 
Title: 28px -> 1.75rem
Subtitle: 18px -> 1.125rem
I used this converter - https://www.ninjaunits.com/converters/pixels/pixels-rem/
Default pixel size: 16px

MocanaAtBroad (approved):

PR #266, opened 2020-09-03
Ddp 4917 sync recaptcha language to application language

MocanaAtBroad:
Deals with problem identified by translators and explained in ticket. Solution described in PR code comments.


ddp-workspace/projects/ddp-sdk/src/lib/services/languageService.service.ts
```diff
@@ -52,20 +52,19 @@ export class LanguageService {
...
     public changeLanguage(languageCode: string): boolean {
-      if (this.canUseLanguage(languageCode)) {
-        this.translate.use(languageCode);
-        localStorage.setItem('studyLanguage', languageCode);
-        return true;
-      }
+        if (this.canUseLanguage(languageCode)) {
+            localStorage.setItem('studyLanguage', languageCode);
+            this.translate.use(languageCode);
+            return true;
```
MocanaAtBroad, line 65: This is the only functional change in this file. Everything else just clears tslint warnings.

Important that we set the language in localStorage first. Setting this value is synchronous. The translate current language happens async. We are going to depend on localstorage value to tell us what is is currently selected globally.

ddp-workspace/projects/ddp-testboston/src/app/app.module.ts
```diff
@@ -102,17 +102,16 @@ sdkConfig.tooltipIconUrl = 'assets/images/info.png';
...
-            translate.setDefaultLang(locale);
-            translate.use(locale).subscribe(() => {
-                console.log(`Successfully initialized '${locale}' language as default.`);
+            translate.setDefaultLang(initialLanguageCode);
+            translate.use(initialLanguageCode).subscribe(() => {
+                console.log(`Successfully initialized '${initialLanguageCode}' language as default.`);
             }, () => {
-                console.error(`Problem with '${locale}' language initialization.`);
+                console.error(`Problem with '${initialLanguageCode}' language initialization.`);
             }, () => {
```
MocanaAtBroad, line 115: Ensuring that we get the current language correctly right of the bat. Before, if Spanish was the language selected in localstorage, we would start of with english then switch to Spanish. Simplifies syncing recaptcha. More below.

ddp-workspace/projects/ddp-testboston/src/app/app.module.ts
```diff
@@ -167,6 +166,10 @@ export function translateFactory(translate: TranslateService, injector: Injector
                 Injector
             ],
             multi: true
+        },
+        {
+            provide: RECAPTCHA_LANGUAGE,
+            useValue: initialLanguageCode,
```
MocanaAtBroad, line 172: We only get to set the recaptcha language once, so here it is. To change it, we would need to reload the Angular app. We getting to that part...

ddp-workspace/projects/ddp-testboston/src/app/components/user-registration-prequal/user-registration-prequal.component.ts
```diff
@@ -22,13 +23,25 @@ export class UserRegistrationPrequalComponent implements OnInit, OnDestroy {
...
+    // recaptcha language can only be initialized only once globally
+    // if there is a language mismatch or the language is changed we need to force
+    // a reload of application to initialize recaptcha with the same language as the translation service
+    this.subscription = merge(
+        of(this.translateService.currentLang).pipe(filter(currentLang => currentLang !== this.recaptchaLanguage)),
+        this.translateService.onLangChange.asObservable()
+    ).subscribe(() => {
+      window.location.reload();
+    });
     this.phone = this.config.phone;
```
MocanaAtBroad, line 44: The code comment explains idea. We force the application to reload if the translate service language does not match recaptcha language or the language is modified.
On reload, app.module.ts will ensures translate language are synced up again.

charlottedye (dismissed):
Looks good!  The "as void" definitely seems cleaner--I don't think I knew about that TypeScript feature.

MocanaAtBroad (commented):
> Looks good! The "as void" definitely seems cleaner--I don't think I knew about that TypeScript feature.

Do you have tslint checks on? Highly recommend it!

ddp-workspace/projects/ddp-testboston/src/app/components/user-registration-prequal/user-registration-prequal.component.ts
```diff
@@ -22,13 +23,25 @@ export class UserRegistrationPrequalComponent implements OnInit, OnDestroy {
   public isLoading = false;
   private anchor: Subscription;
   @ViewChild('captcha', { static: false }) private captcha: RecaptchaComponent;
+  private subscription: Subscription;
```
fat23cat: Could we create Subscription class instance:
`private anchor = new Subscription();`
and add subscriptions using:
`this.anchor.add(sub);`
instead of creating two subscription management properties in the component.

fat23cat (commented):
Looks really nice! 
I have only one suggestion, please, consider it.

ddp-workspace/projects/ddp-testboston/src/app/components/user-registration-prequal/user-registration-prequal.component.ts
MocanaAtBroad: 0f778a11cd7c9b457ca69e85c3f8b2f212e9b994

fat23cat (approved):

PR #269, opened 2020-09-05
DDP-4919 address states list not shown for testboston

MocanaAtBroad:
There was a bug when we tried to set a default country for mailing address. The dependent state list was not updated properly.
This problem was only showing up on testboston; the only study where we confine country to be US.

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
@@ -72,7 +72,7 @@ interface AddressSuggestion {
                 [address]="inputAddress$ | async"
                 [addressErrors]="verifyFieldErrors$ | async"
                 [readonly]="isReadOnly$ | async"
-                [country]="staticCountry$ | async"
+                [country]="country"
                 [phoneRequired]="block.requirePhone"
                 (componentBusy)="isInputComponentBusy$.next($event)"></ddp-address-input>
         <ddp-validation-message
@@ -193,7 +193,6 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
   public isReadOnly$: Observable<boolean>;
   public inputComponentAddress$: Subject<Address | null> = new BehaviorSubject(null);
   public generateTaggedAddress = generateTaggedAddress;
-  public staticCountry$: Observable<string | null>;
 
 
   private ngUnsubscribe = new Subject();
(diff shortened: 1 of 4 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -72,7 +72,7 @@ interface AddressSuggestion {
                 [address]="inputAddress$ | async"
                 [addressErrors]="verifyFieldErrors$ | async"
                 [readonly]="isReadOnly$ | async"
-                [country]="staticCountry$ | async"
+                [country]="country"
```
MocanaAtBroad, line 75: More straightforward.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressInput.component.ts
```diff
@@ -172,7 +172,12 @@ export class AddressInputComponent implements OnInit, OnDestroy {
     this.ais.inputIsReadOnly$.next(val);
   }
   @Input()
-  country: string | null;
+  set country(countryCode: string | null) {
+    this.ais.defaultCountryCode$.next(countryCode);
+  }
+  get country(): string | null {
+    return this.ais.defaultCountryCode$.getValue();
+  }
```
MocanaAtBroad, line 180: Going to try to keep the state in the service instead of keeping a field variable here.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressInput.service.ts
```diff
@@ -145,6 +148,13 @@ export class AddressInputService implements OnDestroy {
         distinctUntilChanged(),
         map(val => ({ isReadOnly: val }))),
 
+      this.defaultCountryCode$.pipe(
+          cachingCountryInfoOp,
+          map(countryInfo => ({ country: (countryInfo ? countryInfo.code : ''), countryInfo })),
+          map(countryInfoState => ({formData: new Address({country: countryInfoState.country}),
+            ...countryInfoState, formDataSource: 'INPUT' }))
+      ),
+
```
MocanaAtBroad, line 157: Setting the defaultCountry is more or less the same as getting an address passed in, but with just the country.

ddp-workspace/projects/ddp-sdk/src/lib/services/addressCountry.service.ts
```diff
@@ -65,7 +64,6 @@ export class CountryService extends SessionServiceAgent<CountryAddressInfo | Cou
                     return [];
                 }
             }),
-          tap((data) => console.log('initializeAllCountryInfosums got allcedss' + data.length))
         ).subscribe(this.allCountryInfoSummariesSubject$);
```
MocanaAtBroad, line 67: Annoying/unnecessary logging,

fat23cat (approved):

PR #270, opened 2020-09-08
Fix test with address list

MocanaAtBroad:
Wait: there is one more test failing!
Fixed the test failing after merging the original PR.
I think it was a mistake to  use the same branch, but it is what is for now. I have commented the changes that actually fix the test.
And this the one new commit:
249b10e09151aa66995d9896a1efa556d105b5af

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.spec.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.spec.ts
@@ -30,7 +30,7 @@ class FakeAddressInputComponent {
   private _readonly = false;
   @Output()valueChanged = new EventEmitter();
   @Input()addressErrors;
-  @Input()country;
+  @Input()country = null;
   @Input()phoneRequired;
   @Input()
   set address(val: Address | null) {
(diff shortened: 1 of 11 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.spec.ts
```diff
@@ -30,7 +30,7 @@ class FakeAddressInputComponent {
   private _readonly = false;
   @Output()valueChanged = new EventEmitter();
   @Input()addressErrors;
-  @Input()country;
+  @Input()country = null;
```
MocanaAtBroad, line 33: This is one change for the test. Test was expecting null but we were getting undefined.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.spec.ts
```diff
@@ -476,14 +476,7 @@ describe('AddressEmbeddedComponent', () => {
...
 
-    expect(childComponent.country).toBeNull();
-    component.activityGuid = '123';
-    component.country = 'CA';
-    tempAddress = new Address({country: 'CA'});
-    addressServiceSpy.getTempAddress.and.returnValue(of(tempAddress));
-    component.ngOnInit();
-    fixture.detectChanges();
-    expect(childComponent.country).toBe('CA');
+    expect(childComponent.country).toBe('US');
```
MocanaAtBroad, line 479: I changed behavior. I was originally allowing for country in input form to show up if we get a full address that had a country that was different from the default country. For the sake of simplicity, I dropped that capability, so this chunk of test was no longer valid.

fat23cat:
Tip: to prevent accidental merge, you can convert the PR to draft in addition to the 'Draft' label:
image

MocanaAtBroad:
> Tip: to prevent accidental merge, you can convert the PR to draft in addition to the 'Draft' label:
> image

I can never find that link! Thanks! (At any rate, I think it is good now).

fat23cat (approved):

PR #271, opened 2020-09-08
DDP-4921 added /ddp for GP kit receipt in BSP

zyme:
DDP-4921 restores the `/ddp` route so that GP can receive kits.

Will put up a separate hotfix PR for prod.

### DIFF
--- a/config/deployment/test/dispatch.yaml
+++ b/config/deployment/test/dispatch.yaml
@@ -19,7 +19,7 @@ dispatch:
     service: pepper-api-spec
   - url: "pepper-test.datadonationplatform.org/*"
     service: pepper-backend
-  - url: "dsm-dev.datadonationplatform.org/ddp/*"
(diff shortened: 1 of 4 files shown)


config/deployment/test/dispatch.yaml
```diff
@@ -19,6 +19,8 @@ dispatch:
     service: pepper-api-spec
   - url: "pepper-test.datadonationplatform.org/*"
     service: pepper-backend
+  - url: "dsm-dev.datadonationplatform.org/ddp/*"
+    service: study-manager-backend
```
MocanaAtBroad, line 23: You sure you want it to be dsm-dev in test, staging and dev?

zyme: OMG

MocanaAtBroad (approved):

PR #282, opened 2020-09-16
DDP-4936: Add [projecturl].org/MailingList route in CMI sites

fat23cat:
For all existing cmi sites on pepper, add a /MailingList route to angular that opens the mailing list modal so that the CMI team can send users directly to the mailing list without them having to click on anything.

### DIFF
--- a/ddp-workspace/projects/toolkit/src/lib/components/app/app-redesigned-base.component.ts
+++ b/ddp-workspace/projects/toolkit/src/lib/components/app/app-redesigned-base.component.ts
@@ -1,0 +1,62 @@
+import { OnInit, OnDestroy, Inject } from '@angular/core';
+import { MatDialog } from '@angular/material/dialog';
+import { NoopScrollStrategy } from '@angular/cdk/overlay';
+import { NavigationEnd, Router } from '@angular/router';
+import { CommunicationService } from '../../services/communication.service';
+import { ToolkitConfigurationService } from '../../services/toolkitConfiguration.service';
+import { JoinMailingListComponent } from '../dialogs/joinMailingList.component';
+import { SessionWillExpireComponent } from '../dialogs/sessionWillExpire.component';
+import { CompositeDisposable, RenewSessionNotifier } from 'ddp-sdk';
+
+export class AppRedesignedBaseComponent implements OnInit, OnDestroy {
+    private anchor = new CompositeDisposable();
+    private readonly DIALOG_BASE_SETTINGS = {
+        width: '740px',
+        position: { top: '30px' },
+        data: {},
+        autoFocus: false,
+        scrollStrategy: new NoopScrollStrategy()
+    };
+
+    constructor(
+        private communicationService: CommunicationService,
+        private dialog: MatDialog,
+        private renewNotifier: RenewSessionNotifier,
+        private router: Router,
+        @Inject('toolkit.toolkitConfig') private config: ToolkitConfigurationService) { }
+
+    public ngOnInit(): void {
+        this.initMailingListDialogListener();
+        this.initSessionExpiredDialogListener();
+        this.initRouterListener();
+    }
+
+    public ngOnDestroy(): void {
+        this.anchor.removeAll();
+    }
+
+    private initMailingListDialogListener(): void {
+        const modalOpen = this.communicationService.openJoinDialog$.subscribe(() => {
+            this.dialog.open(JoinMailingListComponent, this.DIALOG_BASE_SETTINGS);
+        });
+        this.anchor.addNew(modalOpen);
+    }
+
+    private initSessionExpiredDialogListener(): void {
+        const modalOpen = this.renewNotifier.openDialogEvents.subscribe(() => {
+            this.dialog.open(SessionWillExpireComponent, { ...this.DIALOG_BASE_SETTINGS, disableClose: true });
+        });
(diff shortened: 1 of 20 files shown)


ddp-workspace/projects/toolkit/src/lib/components/app/app-redesigned-base.component.ts
```diff
@@ -0,0 +1,62 @@
...
+import { MatDialog } from '@angular/material/dialog';
+import { NoopScrollStrategy } from '@angular/cdk/overlay';
+import { NavigationEnd, Router } from '@angular/router';
+import { CommunicationService } from '../../services/communication.service';
+import { ToolkitConfigurationService } from '../../services/toolkitConfiguration.service';
+import { JoinMailingListComponent } from '../dialogs/joinMailingList.component';
+import { SessionWillExpireComponent } from '../dialogs/sessionWillExpire.component';
+import { CompositeDisposable, RenewSessionNotifier } from 'ddp-sdk';
+
+export class AppRedesignedBaseComponent implements OnInit, OnDestroy {
```
fat23cat, line 11: Common logic from osteo and brain-v2 app components was extracted in the base class.

ddp-workspace/projects/toolkit/src/lib/components/app/app.component.ts
```diff
@@ -178,7 +178,7 @@ export class AppComponent implements OnInit, AfterViewInit, OnDestroy {
                 of({ joinDialogOpen: false }) :
                 merge(
                     dialog.afterClosed().pipe(map(() => ({ joinDialogOpen: false }))),
-                    dialog.afterOpen().pipe(map(() => ({ joinDialogOpen: true })))
+                    dialog.afterOpened().pipe(map(() => ({ joinDialogOpen: true })))
```
fat23cat, line 181: `afterOpen` method is deprecated

ddp-workspace/projects/toolkit/src/lib/components/dialogs/joinMailingList.component.ts
```diff
@@ -101,6 +101,8 @@ export class JoinMailingListComponent implements OnInit, OnDestroy {
             if (isSubmitted) {
                 this.communicationService.closeSidePanel();
                 this.router.navigateByUrl(this.stayInformedUrl);
+            } else if (this.router.url.includes(this.toolkitConfiguration.mailingListDialogUrl)) {
+                this.router.navigateByUrl('/');
```
fat23cat, line 105: Return a user to the root URL if the modal window was closed.

ddp-workspace/projects/toolkit/src/lib/components/app/app-redesigned-base.component.ts
```diff
@@ -0,0 +1,62 @@
...
+        this.initRouterListener();
+    }
+
+    public ngOnDestroy(): void {
+        this.anchor.removeAll();
+    }
+
+    private initMailingListDialogListener(): void {
+        const modalOpen = this.communicationService.openJoinDialog$.subscribe(() => {
+            this.dialog.open(JoinMailingListComponent, this.DIALOG_BASE_SETTINGS);
```
MocanaAtBroad, line 40: Does the app*component have to be the one that handles the dialog? Already A LOT going on here. Perhaps this can be setup in a separate component?
Also, we should try to avoid an explicit reference to JoinMailingListComponent from this component; this component is meant to be generic/high-level/reusable and JoinMailingListComponent is specific/different level of abstraction.
Maybe the dialog class could be injected into a different component?

fat23cat: Hm, AppRedesignedBaseComponent is only 63 lines and 5 methods, I don't think that there is going on a lot :) 
It's a bit unclear for me why we should avoid JoinMailingListComponent here: AppRedesignedBaseComponent will be a universal component for all redesigned CMI studies, and at least 2 studies already use this component - osteo and brain-v2. And all CMI studies use JoinMailingListComponent.
And the app component the most appropriate place to init global listeners, because it exists during the whole app lifecycle.

fat23cat: But maybe we can do this base component more universal - we can make constructor parameters optional using @Optional() decorator, so then it allows disabling initMailingListDialogListener or initSessionExpiredDialogListener if we don't need some of them.

MocanaAtBroad: Actually, I guess before I suggest anything, I am not sure if I understand the purpose of this component. You also have an app.component.ts that appears to wire all the the functionality of the app together, but "redesigned" is not connect to it.
So is this meant to be the updated base component for the newer style apps (e.g., Osteo) and we would use app.component.ts just for the older apps (e.g., Angio). I am getting this right?

fat23cat: Yes, angio, brain, mbc, mpc use app.component.ts
osteo, brain-v2 use redesigned as a base component, which inits global listeners

MocanaAtBroad: OK. No problem. Let's move forward with it.

MocanaAtBroad (approved):

PR #297, opened 2020-09-22
DDP-5002: MailingList borked for temp users

fat23cat:
Bug description: https://broadinstitute.atlassian.net/browse/DDP-5002


ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/userProfileServiceAgent.service.ts
```diff
@@ -29,7 +29,7 @@ export class UserProfileServiceAgent extends UserServiceAgent<UserProfile> {
                 return new UserProfileDecorator(x);
             }),
             catchError(e => {
-                if (e.error && e.error.errorCode && e.error.errorCode === 'MISSING_PROFILE') {
+                if (e.error && e.error.code && e.error.code === 'MISSING_PROFILE') {
```
fat23cat, line 32: Only one real code change here: it looks like API changed and now we use `code` instead of `errorCode`. Due to this issue method threw an error instead of the new `UserProfileDecorator`.

MocanaAtBroad (approved):

PR #303, opened 2020-09-23
Uppercase UPS route appears to break when running in AOT

MocanaAtBroad:
testboston.dev.datadonatioplatform.org was coming up with a blank page.
This is a screenshot of the console:
Screen Shot 2020-09-23 at 2 57 05 PM

My interpretation of problem is that the AOT compiled app did not like including the toUppercase() function in the route definitions.

I deployed to dev environment the fix. It appears to deal with problem.


fat23cat (approved):
Oh, my bad. Thanks!

PR #313, opened 2020-10-06
DDP-4303: Angular logging service

fat23cat:
- Logging service works
- All console. replaced with logging service
- Other improvements, details will be provided in the comments


config/pepperConfig.js.ctmpl
```diff
@@ -14,6 +14,17 @@ var DDP_ENV = {
...
+#Log levels: 1 - debug; 2 - info; 3 - warning; 4 - error. E.g. 1 includes debug level and higher, 2 includes info and higher, etc.
+{{if eq $environment "dev"}}
+    DDP_ENV['logLevel'] = 1;
+{{else if eq $environment "test"}}
+    DDP_ENV['logLevel'] = 1;
+{{else if eq $environment "staging"}}
+    DDP_ENV['logLevel'] = 1;
+{{else if eq $environment "prod"}}
+    DDP_ENV['logLevel'] = 2;
+{{end}}
```
fat23cat, line 26: Decided to setup LogLevel for all apps in one place. If you have another opinion about LogLevel for some env, please, let me know.
cc @zyme

ddp-workspace/projects/basil-app/src/app/app.module.ts
```diff
@@ -53,7 +53,7 @@ config.backendUrl = DDP_ENV.basePepperUrl;
 config.auth0Domain = DDP_ENV.auth0Domain;
 config.auth0ClientId = DDP_ENV.auth0ClientId;
 config.studyGuid = 'TESTSTUDY1';
-config.logLevel = LogLevel.Info;
+config.logLevel = LogLevel.Debug;
```
fat23cat, line 56: Since basil-app is not going to go to the prod, we may hardcode Debug LogLevel.

ddp-workspace/projects/ddp-angio/src/app/app.module.ts
```diff
@@ -72,7 +72,7 @@ config.backendUrl = DDP_ENV.basePepperUrl;
 config.auth0Domain = DDP_ENV.auth0Domain;
 config.auth0ClientId = DDP_ENV.auth0ClientId;
 config.studyGuid = DDP_ENV.studyGuid;
-config.logLevel = LogLevel.Info;
+config.logLevel = DDP_ENV.logLevel;
```
fat23cat, line 75: Now we can setup LogLevel using environment config.

ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
```diff
@@ -1,40 +1,39 @@
...
+export class LoggingService implements OnDestroy {
     private subjectSource = new ReplaySubject<LogEvent>();
     private anchor: Subscription;
 
     constructor(
-        @Inject('ddp.config') private config: ConfigurationService,
-        private dispatcher: ExceptionDispatcher) {
-    }
-
-    public ngOnInit(): void {
```
fat23cat, line 20: The problem was that Angular services don't have ngOnInit lifecycle hook. Their initialization should be in the constructor.

ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
```diff
@@ -44,16 +43,23 @@ export class LoggingService implements OnInit, OnDestroy {
...
         );
     }
 
+    private initEventsListener(): void {
+        this.anchor = this.getSubscription().subscribe(event => {
+            switch (event.level) {
+                case LogLevel.Debug: console.debug(`${event.message}::${event.context}`); break;
+                case LogLevel.Info: console.log(`${event.message}::${event.context}`); break;
+                case LogLevel.Warning: console.warn(`${event.message}::${event.context}`); break;
+                case LogLevel.Error: console.error(`${event.message}::${event.context}`); break;
```
MocanaAtBroad: Getting tslint warning  that we don't have a default case for this `switch`

MocanaAtBroad:
The code makes sense. Would like to see if we are able to maintain in log output the line number and file source file of the original statement that calls the logging service instead of always showing `logging.service.ts:56` 
Here is some sample code (what you are implementing is more complicated but I think the idea can be applied to your code as well). Have not checked out if maintaining the file name is possible.
If you run this html in your browser, you will notice that the line number for ` log("Foo", {bar: 1})()` is what shows up in the console and 

```html
<html>
<head>
    <style>

    </style>
</head>
<body>

<script type="text/javascript">
    log = function() {
        // Put your extension code here
        var args = Array.prototype.slice.call(arguments);
        console.log("The args are: %o", args);
        args.unshift(console);
        console.log("The args after unshift: %o" , args);
        return Function.prototype.bind.apply(console.log, args);
    }

    // Note the extra () to call the original console.log

    log("Foo", {bar: 1})();
</script>
</body>
</html>
```

MocanaAtBroad (approved):
Works really nice!

PR #314, opened 2020-10-06
Ddp 4940 dashboard activity loading

charlottedye:
See DDP-4940.  Prion's dashboard displays an additional line under the main table that gives information about the study listing page.  Because that text is specified in the language file, that piece of the table ends up loading quickly compared to the other activities, which take a while to load.  This code just delays loading the static activity until the other activities have been retrieved from the backend (in practice, the static activity still loads a little earlier, but not to the extent that it could lead people to believe the study loading page is the only available activity).


fat23cat:
Instead of creating additional service, could we emit a `loaded` boolean event using Angular EventEmitter from `ddp-user-activities` to `ddp-dashboard` and the from `ddp-dashboard` to outside and that handle it?

fat23cat (approved):

PR #317, opened 2020-10-10
DDP-5079 testboston: new languages

yufengwng:
See ticket for the ZIP that contains the translations.


ddp-workspace/projects/ddp-testboston/src/assets/auth0-login-ar.js
```diff
@@ -0,0 +1,121 @@
+Auth0.registerLanguageDictionary("ar", {
```
yufengwng, line 1: These two `auth0-login-ar.js` and `auth0-login.pt.js` files are needed because Auth0 does not have native support for Arabic or Portuguese, so we have to register our own dictionaries.

fat23cat (dismissed):

fat23cat (approved):

PR #318, opened 2020-10-12
DDP-5078: Better SEO elements for TestBoston

fat23cat:
Requirements are here https://broadinstitute.atlassian.net/browse/DDP-5078


ddp-workspace/projects/ddp-testboston/src/assets/i18n/en.json
```diff
@@ -500,7 +500,10 @@
     },
     "Title": {
       "Default": "TestBoston",
-      "Alert": "Your session will time-out soon"
+      "Alert": "Your session will time-out soon",
+      "Dashboard": "TestBoston Participant Dashboard",
+      "UPS": "Covid at-home test and shipping instructions for TestBoston",
```
MocanaAtBroad: I think this should COVID instead of Covid, right?

ddp-workspace/projects/ddp-testboston/src/assets/i18n/fr.json
```diff
@@ -500,7 +500,10 @@
     },
     "Title": {
       "Default": "TestBoston",
-      "Alert": "Votre session va bientôt expirer"
+      "Alert": "Votre session va bientôt expirer",
+      "Dashboard": "TestBoston Participant Dashboard",
+      "UPS": "Covid at-home test and shipping instructions for TestBoston",
+      "Home": "TestBoston - The at-home COVID-19 testing study"
```
MocanaAtBroad, line 506: If we don't have a translation, let's figure out a way to track it . Maybe we do the same thing we do in back-end, addding the [FR] / [ES] prefix?

fat23cat: Will ask Jen.

MocanaAtBroad (commented):
Code LGTM, but still have questions about text/translations.

PR #320, opened 2020-10-14
DDP-5030: Brain pediatrics styling

fat23cat:
Mockups: https://projects.invisionapp.com/d/main?origin=v7#/console/20209715/423517836/preview?scrollOffset=1600

### DIFF
--- a/ddp-workspace/projects/ddp-brain-v2/src/assets/i18n/en.json
+++ b/ddp-workspace/projects/ddp-brain-v2/src/assets/i18n/en.json
@@ -563,6 +563,9 @@
         }
     },
     "Toolkit": {
+        "Dashboard": {
+            "Title": "My Dashboard"
+        },
         "Dialogs": {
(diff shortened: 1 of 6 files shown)


MocanaAtBroad:
I don't think this problem list bullets spacing  is caused by one of the styles you modified and you are still working on this, so approving so you are not blocked. But including this display bug just in case. Not asking you fix it right away, just pointing it out:
Screen Shot 2020-10-15 at 9 08 35 AM

MocanaAtBroad (approved):

fat23cat:
> I don't think this problem list bullets spacing is caused by one of the styles you modified and you are still working on this, so approving so you are not blocked. But including this display bug just in case. Not asking you fix it right away, just pointing it out:
> 
Thanks for the notice, I know about this bug. We need to update brain activities on the backend side to reflect new changes, css class name was added there, but dev contains old activities.

PR #322, opened 2020-10-15
DDP-4303: Angular logging service

fat23cat:
@MocanaAtBroad, this PR was inspired by your comment in the #313. What do you think about this logging implementation?

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
@@ -1,68 +1,20 @@
-import { Injectable, Inject, OnDestroy } from '@angular/core';
+import { Injectable, Inject } from '@angular/core';
 import { ConfigurationService } from './configuration.service';
-import { LogEvent } from '../models/logEvent';
 import { LogLevel } from '../models/logLevel';
-import { DdpException } from '../models/exceptions/ddpException';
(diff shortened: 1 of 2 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
```diff
@@ -1,68 +1,20 @@
...
-        this.anchor = this.getSubscription().subscribe(event => {
-            switch (event.level) {
-                case LogLevel.Debug: console.debug(`${event.message}::${event.context}`); break;
-                case LogLevel.Info: console.log(`${event.message}::${event.context}`); break;
-                case LogLevel.Warning: console.warn(`${event.message}::${event.context}`); break;
-                case LogLevel.Error: console.error(`${event.message}::${event.context}`); break;
-            }
-        });
-    }
+    public logError = this.showEvent(LogLevel.Error) ? console.warn.bind(window.console) : () => { };
```
MocanaAtBroad, line 15: I think you can use console.error instead of console.warn for this one.

MocanaAtBroad (commented):
This capture the original line number AND source file name. Awesome!

One small comment.

ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
fat23cat: Good catch, thanks!

MocanaAtBroad (approved):

PR #333, opened 2020-10-26
DDP-5179: Support right-to-left languages

fat23cat:
All styles should be mirrored.
Phone numbers should be shown in the same ways as for other languages.


ddp-workspace/projects/ddp-testboston/src/assets/i18n/ar.json
```diff
@@ -33,9 +33,9 @@
   "MailingList": {
     "Title": "نموذج القائمة البريدية",
     "Questions": {
-      "Adult": "1. هل تبلغ من العمر 18 عامًا أو أكثر؟",
-      "Patient": "2. هل أنت أحد مرضى مستشفى بريجهام والنساء؟",
-      "Email": "3. ما هو عنوان البريد الإلكتروني الأفضل لك؟"
+      "Adult": "١. هل تبلغ من العمر 18 عامًا أو أكثر؟",
+      "Patient": "٢. هل أنت أحد مرضى مستشفى بريجهام والنساء؟",
+      "Email": "٣. ما هو عنوان البريد الإلكتروني الأفضل لك؟"
```
yufengwng, line 38: I think it should be fine to keep numeric digits as digits instead of Arabic symbols?

ddp-workspace/projects/ddp-testboston/src/style/right-to-left.scss
```diff
@@ -0,0 +1,280 @@
+html[dir='rtl'] {
+
```
MocanaAtBroad, line 2: Do we need to set this attribute in the DOM ourselves? (I am not sure. This is more of question).
I tried setting it in Chrome after switching to Arabic using the language selector.
It does set the text alignment to the right and the cursor for input moves right to left as you enter text.
If we are not setting it, when does this style get applied?

MocanaAtBroad: Also home page looks better when setting this attribute. Obviously you counting on this as CSS selector. How do we set it?

ddp-workspace/projects/ddp-testboston/src/assets/i18n/ar.json
fat23cat: I had a chat with Pegah, she said that it would be nicer to use Arabic symbols instead of numeric digits.

ddp-workspace/projects/ddp-testboston/src/style/right-to-left.scss
fat23cat: > Do we need to set this attribute in the DOM ourselves? (I am not sure. This is more of question).
I tried setting it in Chrome after switching to Arabic using the language selector.
It does set the text alignment to the right and the cursor for input moves right to left as you enter text.
If we are not setting it, when does this style get applied?

We absolutely need to set up this attribute, because ltr languages should be mirrored on the screen. 

>Also home page looks better when setting this attribute. Obviously you counting on this as CSS selector. How do we set it?

We set i18n attributes here: https://github.com/broadinstitute/ddp-angular/blob/DDP-5179_rtl/ddp-workspace/projects/ddp-sdk/src/lib/components/internationalization/languageSelector.component.ts#L252

ddp-workspace/projects/ddp-sdk/src/lib/components/internationalization/languageSelector.component.ts
```diff
@@ -234,4 +237,21 @@ export class LanguageSelectorComponent implements OnInit, OnDestroy {
...
+      this.changeDocumentLanguage(direction, languageCode);
+    });
+    this.anchor.addNew(sub);
+  }
+
+  private changeDocumentLanguage(direction: string, languageCode: string): void {
+    const html = this.document.querySelector('html');
+    html.setAttribute('dir', direction);
+    html.setAttribute('lang', languageCode);
+  }
```
MocanaAtBroad, line 256: Fixed it for me! I don't understand what was wrong with your original code; looked good to me. But this definitely works now.

MocanaAtBroad (approved):

PR #339, opened 2020-11-05
DDP-5251 Setting a max height on text area appears to fix sudden scroll problem

MocanaAtBroad:
Change sets a maximum height on textarea entry. Any text beyond that height will trigger a scrollbar showing up for field.
The sudden jumping observed does not appear to occur no matter how large entry is.

### DIFF
--- a/ddp-workspace/projects/ddp-brain/src/styles/activities.scss
+++ b/ddp-workspace/projects/ddp-brain/src/styles/activities.scss
@@ -463,6 +463,10 @@ select:disabled {
     }
 }
 
+.ddp-textarea {
+    max-height: 10rem !important;
+}
+
 // nested picklist
 .ddp-nested-picklist {
     margin-left: 3rem;


ssettipalli (approved):

ddp-workspace/projects/ddp-brain/src/styles/activities.scss
```diff
@@ -463,6 +463,10 @@ select:disabled {
     }
 }
 
+.ddp-textarea {
```
zyme, line 466: Is there a `overflow-y:scroll` in play somewhere here?

MocanaAtBroad: Did not see one, but text area appears to be doing the right thing (or at least, what I would want it to do: stretch out to 10 lines and then show scrollbar if more text.
I doubt 10 lines would ever be needed (even in mobile it is a lot). But seems like reasonable behavior.

PR #340, opened 2020-11-05
DDP-5254 testboston: red message

yufengwng:
Add CSS for red message. The shade of red is the same as for `ErrorMessage`.


yufengwng:
Not doing this to avoid a round of Angular deploys.

PR #348, opened 2020-11-12
[TestBoston] Stepper for single section consent 

fat23cat:
We need to allow a user to navigate between different Consent section even Consent is a single page.
Here is a code snippet, which should be added in the content block to show stepper for 1-section consent:
``` HTML
<div class="activity-steps">
    <div class="activity-step active completed scrollable step1">
        <span class="activity-step__number scrollable step1">1</span>
        <span class="activity-step__text scrollable step1">Introduction</span>
    </div>
    <div class="activity-steps__divider"></div>
    <div class="activity-step active completed scrollable step2">
        <span class="activity-step__number scrollable step2">2</span>
        <span class="activity-step__text scrollable step2">Key Information</span>
    </div>
    <div class="activity-steps__divider"></div>
    <div class="activity-step active completed scrollable step3">
        <span class="activity-step__number scrollable step3">3</span>
        <span class="activity-step__text scrollable step3">Full Form</span>
    </div>
</div>
```
Desired section should have id="step1" or id="step2" etc.

### DIFF
--- a/ddp-workspace/projects/ddp-testboston/src/app/components/app/app.component.ts
+++ b/ddp-workspace/projects/ddp-testboston/src/app/components/app/app.component.ts
@@ -31,6 +31,20 @@ export class AppComponent implements OnInit, OnDestroy {
     this.anchor.removeAll();
   }
 
+  public scrollToStep(event: MouseEvent): void {
+    const element = event.target as HTMLElement;
+    if (element.className.includes('scrollable')) {
+      const id = element.classList[element.classList.length - 1];
+      const step = document.getElementById(id);
+      const headerHeight = 7.5 * parseFloat(getComputedStyle(document.documentElement).fontSize);
+      const top = step.getBoundingClientRect().top + window.scrollY - headerHeight;
+      window.scrollTo({
+        top,
+        behavior: 'smooth'
+      });
+    }
+  }
+
   private initSessionExpiredDialogListener(): void {
     const modalOpen = this.renewNotifier.openDialogEvents.subscribe(() => {
       this.dialog.open(SessionWillExpireComponent, { ...this.DIALOG_BASE_SETTINGS, disableClose: true });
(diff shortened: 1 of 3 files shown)


ddp-workspace/projects/ddp-testboston/src/app/components/app/app.component.ts
```diff
@@ -31,6 +31,23 @@ export class AppComponent implements OnInit, OnDestroy {
     this.anchor.removeAll();
   }
 
+  public scrollToStep(event: MouseEvent): void {
+    const element = event.target as HTMLElement;
+    if (element.className.includes('scrollable')) {
```
MocanaAtBroad, line 36: This is very clever!

MocanaAtBroad (dismissed):
Can't really test it. Does not appear we have the updated consent available. Approving so we can try it out more easily.

ddp-workspace/projects/ddp-testboston/src/app/components/app/app.component.ts
```diff
@@ -31,6 +31,23 @@ export class AppComponent implements OnInit, OnDestroy {
...
+      const id = element.classList[element.classList.length - 1];
+      const step = document.getElementById(id);
```
yufengwng: Unfortunately, it doesn't look like we can use the `id` attribute. If I put something like this in the backend:
```
bodyTemplate: {
   ...
   "template": "... <div id="step1"></div> ..."
}
```

Then what gets rendered on Angular side is this:
```
<div></div>
```

In other words, the `id` attribute gets stripped out by Angular. Can we configure Angular NOT to strip it out? Maybe we match on a CSS `class` instead? (something like `step1-marker` for `step1`?)

yufengwng (changes requested):

ddp-workspace/projects/ddp-testboston/src/app/components/app/app.component.ts
fat23cat: This is an interesting one, thanks, will fix it.

fat23cat: Fixed. Target section sould have `class="anchor-step1"` css class.

yufengwng: Sounds good! I'll put that in the activity configuration.

yufengwng (approved):

yufengwng:
I've tried this locally and it works well. Let's merge and push forward on this.

PR #364, opened 2020-11-20
Bugfix: update session locale on language change

aleksey-kuznetsov:
When an unauthenticated user changes the language it is not being reflected in the session object. Because of this behaviour `Accept-Language` header contains default `en` value.
These changes will make sure that session locale will be updated no matter what (logged in or not) and `Accept-Language` header will have appropriate value. This changes will not affect authenticated users since backend prioritizes `preferredLanguage` field in the user profile to identify user language.


ddp-workspace/projects/ddp-sdk/src/lib/services/internationalization/languageService.service.ts
```diff
@@ -67,4 +72,11 @@ export class LanguageService {
     const obs = this.changeLanguageObservable(languageCode);
     return obs !== null;
   }
+
+  public getAppLanguageCode(): string {
```
aleksey-kuznetsov, line 82: This method is useful for the `translateFactory` to init the application with appropriate language.

ddp-workspace/projects/ddp-sdk/src/lib/services/sessionMemento.service.ts
```diff
@@ -110,6 +116,19 @@ export class SessionMementoService implements OnDestroy {
...
+      if (this.session === null) {
+        return;
+      }
+
+      const updatedSession = Object.assign({}, this.session, { locale: languageCode });
+
+      localStorage.setItem(this.SESSION_KEY, JSON.stringify(updatedSession));
+      localStorage.setItem(this.TOKEN_KEY, updatedSession.idToken);
+
+      this.sessionSubject.next(updatedSession);
```
fat23cat: You can update session using existing method:
```suggestion
      this.updateSession(updatedSession);
```

aleksey-kuznetsov: Yep, removed `updateSessionLocale` and switched to `updateSession`.

ddp-workspace/projects/ddp-sdk/src/lib/services/internationalization/languageService.service.ts
```diff
@@ -67,4 +72,11 @@ export class LanguageService {
...
+    return this.translate.currentLang ||
+      (this.session.session && this.session.session.locale) ||
```
fat23cat, line 84: I think we shouldn't rely on the session's locale because here is a bug(I'm not sure where: backend, study config, or auth0): if TestBoston user logs in, chooses a language other then English, logs out and then logs in again, session info which stores in the localStorage will contain `locale: en` instead of preferred user's language.

ddp-workspace/projects/ddp-sdk/src/lib/services/internationalization/languageService.service.ts
```diff
@@ -67,4 +78,11 @@ export class LanguageService {
...
+      localStorage.getItem('studyLanguage') ||
+      this.config.defaultLanguageCode;
```
fat23cat: Can we also add `|| 'en';` to make 'en' default?

fat23cat (approved):

PR #368, opened 2020-11-23
DDP-5343: [TestBoston] The /join page constantly reloads when a user changes the language 

fat23cat:


ddp-workspace/projects/ddp-testboston/src/app/app.module.ts
```diff
@@ -105,16 +106,18 @@ sdkConfig.tooltipIconUrl = 'assets/images/info.png';
...
+export function translateFactory(translate: TranslateService,
+    injector: Injector,
+    logger: LoggingService,
+    language: LanguageService): () => Promise<any> {
     return () => new Promise<any>((resolve: any) => {
         const LOG_SOURCE = 'AppModule';
         const locationInitialized = injector.get(LOCATION_INITIALIZED, Promise.resolve(null));
         locationInitialized.then(() => {
-            const locale = 'en';
+            const locale = language.getAppLanguageCode();
```
fat23cat, line 120: The bug occurred due to race conditions in the language setup. We should not hardcode 'en' language.

MocanaAtBroad (approved):

PR #377, opened 2020-12-04
DDP-5365: Implement ActivityDateNavyValidationRule i18n

fat23cat:
@aleksey-kuznetsov, @charlottedye, please note that I've updated AT and Prion i18n files.


charlottedye (dismissed):
It looks like the only studies this affects are studies that are only in English, but it seems like this is the sort of thing where it would ideally be translated before being added to language files.

MocanaAtBroad (dismissed):

MocanaAtBroad (approved):

PR #380, opened 2020-12-07
Add statistics service

aleksey-kuznetsov:
This change adds new statistics service for gathering study specific statistics.

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/models/statistic.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/models/statistic.ts
@@ -1,0 +1,14 @@
+export interface Statistic {
+  configuration: {
+    type: string;
+    questionStableId: string | null;
+    answerValue: string | null;
+  };
+  statistics: {
+    name: string;
+    data: {
+      count: string;
+      optionDetails?: any;
+    };
+  }[];
+}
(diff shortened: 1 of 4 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/models/statistic.ts
```diff
@@ -0,0 +1,14 @@
+export interface Statistic {
+  configuration: {
+    type: string;
```
fat23cat: Probably it would be better to create an enum with acceptable values: `PARTICIPANTS` `DISTRIBUTION` `MAILING_LIST` `SPECIFIC_ANSWER` `KITS` instead of string type.

ddp-workspace/projects/ddp-sdk/src/lib/models/statistic.ts
```diff
@@ -0,0 +1,14 @@
...
+    questionStableId: string | null;
+    answerValue: string | null;
+  };
+  statistics: {
+    name: string;
+    data: {
+      count: string;
+      optionDetails?: any;
```
fat23cat: We can use `ActivityPicklistOption` interface instead of any.

fat23cat (changes requested):

ddp-workspace/projects/ddp-sdk/src/lib/models/statistic.ts
aleksey-kuznetsov: Fixed. This type fully depends on the study configuration so this has to be generic.

ddp-workspace/projects/ddp-sdk/src/lib/models/statistic.ts
aleksey-kuznetsov: Fixed.

fat23cat (approved):

PR #388, opened 2020-12-11
DDP-5403: [Sessions - Edge Case] Session popup can freeze if expired in inactive tab

fat23cat:


ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
```diff
@@ -351,7 +352,7 @@ export class Auth0AdapterService implements OnDestroy {
     }
 
     private getSessionExpiredUrl(): string {
-        return this.session.session.isAdmin ?
```
fat23cat, line 347: The problem occurred due to race conditions. When on the one tab the `handleExpiredSession()` method was executed and the session was cleared, on the other tab when trying to execute this code, the session will be equal null, for this reason, we need to store isAdmin value in the property to correctly redirect a user when the session was expired.

MocanaAtBroad (approved):

PR #392, opened 2020-12-11
DDP-5396 better handling of failed admin logins

yufengwng:
This PR improves our handling of failed admin login attempts. With the changes in https://github.com/broadinstitute/ddp-study-server/pull/563, we will have fine control of which google account can login to TestBoston Prism. If a google account is not pre-authorized, it will error now. I discovered that when login fails, Auth0 caches some session information in their layer (see [here](https://auth0.com/docs/logout) for the different layers). So when you go back in to retry login, it reuses the cached information and fail you again. This means once someone fails the first login, they will have difficulty logging in even after we authorized them to login. This will be problematic for TestBoston admin logins.

Basic idea here is to call Auth0's `logout` API after a failed login attempt. This should clear an cached Auth0 information. I have verified this to work for local registration as well as in `dev` environment.

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/login/auth0-code-callback.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/login/auth0-code-callback.component.ts
@@ -64,7 +64,12 @@ export class Auth0CodeCallbackComponent implements OnInit, OnDestroy {
                 this.windowRef.nativeWindow.location.href = nextUrl;
             });
         } else {
-            throw new Error('No auth code present in url.');
+            const error = this.route.snapshot.queryParams['error'];
+            const errorDescription = this.route.snapshot.queryParams['error_description'];
+            this.log.logError(this.LOG_SOURCE, 'No auth code present in url');
+            this.log.logError(this.LOG_SOURCE, `Error: ${error}`);
+            this.log.logError(this.LOG_SOURCE, `Error Description: ${errorDescription}`);
+            this.adapter.logout();
         }
     }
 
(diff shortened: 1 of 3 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
```diff
@@ -263,7 +256,7 @@ export class Auth0AdapterService implements OnDestroy {
 
     public logout(returnToUrl: string = ''): void {
         const baseUrl = this.configuration.baseUrl;
-        const wasAdmin = this.session.session.isAdmin;
+        const wasAdmin = this.session.session && this.session.session.isAdmin;
```
yufengwng, line 259: We might not have a session object if we failed login, so adding a safeguard here.

ddp-workspace/projects/ddp-sdk/src/lib/components/login/auth0-code-callback.component.ts
```diff
@@ -64,7 +64,12 @@ export class Auth0CodeCallbackComponent implements OnInit, OnDestroy {
                 this.windowRef.nativeWindow.location.href = nextUrl;
             });
         } else {
-            throw new Error('No auth code present in url.');
+            const error = this.route.snapshot.queryParams['error'];
+            const errorDescription = this.route.snapshot.queryParams['error_description'];
+            this.log.logError(this.LOG_SOURCE, 'No auth code present in url');
+            this.log.logError(this.LOG_SOURCE, `Error: ${error}`);
+            this.log.logError(this.LOG_SOURCE, `Error Description: ${errorDescription}`);
+            this.adapter.logout();
```
yufengwng: Since this component is in the SDK and not toolkit, I don't think we have a general SDK "error page", so the logout will currently redirect to homepage. That might be confusing but I think its okay for local development.

fat23cat: We have error page URL in the SDK Configuration service.
```suggestion
            this.adapter.logout(this.configuration.errorPageUrl);
```

fat23cat:
Looks good overall, only one suggestion.

ddp-workspace/projects/ddp-sdk/src/lib/components/login/auth0-code-callback.component.ts
yufengwng: Sweet, thanks!

fat23cat (dismissed):

yufengwng:
Sorry, accidentally pushed a commit from another branch. I have reverted it. Can I get another checkmark?

fat23cat (approved):

PR #396, opened 2020-12-21
Fix IrbGuard and IrbPasswordService issue

fat23cat:


ddp-workspace/projects/ddp-sdk/src/lib/services/irbPassword.service.ts
```diff
@@ -34,7 +34,7 @@ export class IrbPasswordService extends NotAuthenticatedServiceAgent<boolean> {
                 }
                 return loggedIn;
             }),
-            catchError(this.handleError));
+            catchError(this.handleError.bind(this)));
```
fat23cat, line 37: We pass `this.handleError` as a callback, so, we have to save context, to make this.log work properly.
Without this, we will get an error in the console if the backend returns an error:
image

ddp-workspace/projects/ddp-sdk/src/lib/guards/irb.guard.ts
```diff
@@ -1,25 +1,32 @@
...
         } else {
             return this.irbPassword.requiresIrbAuthentication().pipe(
-                tap(passwordRequired => passwordRequired && this.router.navigate(['/password'])),
-                map(passwordRequired => !passwordRequired));
+                tap(passwordRequired => passwordRequired && this.router.navigateByUrl(this.config.passwordPageUrl)),
+                map(passwordRequired => !passwordRequired),
+                catchError(() => {
+                    this.router.navigateByUrl(this.config.errorPageUrl);
+                    return of(false);
+                })
```
fat23cat, line 28: If backend returns errors, a user will get stuck on this page:
image
Let's redirect him to the error page.

fat23cat:
I've found one more inaccuracy, PR was converted to draft.

MocanaAtBroad (approved):

PR #399, opened 2020-12-25
Angular 10 minor fixes

fat23cat:


ddp-workspace/projects/ddp-sdk/src/lib/components/validationMessage.component.ts
```diff
@@ -11,7 +11,7 @@ import * as _ from 'underscore';
             </li>
         </ul>
         <ng-template #singleMessage>
-            <span [innerHTML]="(isArrayOfMessages() ? message[0] : message) | translate" routeTransformer></span>
```
fat23cat, line 14: The compiler was not happy with this line: the translate pipe can handle only strings, but the compiler assumes that `message` can be a string or array of strings.

ddp-workspace/projects/toolkit/tsconfig.lib.json
```diff
@@ -15,15 +15,7 @@
...
       "es2018"
-    ],
-    "paths": {
-      "ddp-sdk": [
-        "dist/ddp-sdk"
-      ],
-      "ddp-sdk/*": [
-        "dist/ddp-sdk/*"
-      ]
-    }
```
fat23cat, line 26: We don't need it, `paths` were added in the `ddp-workspace/tsconfig.base.json`

MocanaAtBroad (approved):

PR #405, opened 2021-01-08
DDP-5508 fix issue where picklist dropdown native select is not prepopulated

yufengwng:
Fixes DDP-5508.


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/dropdownActivityPicklistQuestion.component.ts
```diff
@@ -29,7 +29,7 @@ import { NGXTranslateService } from '../../../services/internationalization/ngxT
     </ng-template>
 
     <ng-template #nativeSelect>
-        <select [value]="setNativeSelected()"
+        <select [(ngModel)]="nativeSelectedValue"
```
yufengwng, line 32: My solution here is derived from this [SO question](https://stackoverflow.com/questions/62869182/unable-to-set-a-option-selected-by-default-in-angular-10). Not sure if the issue is related to Angular 10, but this approach seems to fix it.

fat23cat (approved):

PR #406, opened 2021-01-08
DDP-5504 hide question errors if block is hidden

yufengwng:
Fixes DDP-5504. The bug is caused by not hiding the validation messages if the block is hidden, so we add a check to the component template.

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityQuestion.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityQuestion.component.ts
@@ -18,12 +18,12 @@ import { delay, filter, map, shareReplay, startWith, takeUntil, tap } from 'rxjs
                   [validationRequested]="validationRequested$ | async"
                   (valueChanged)="enteredValue$.next($event)">
           </ddp-activity-answer>
-          <div *ngIf="block.shown">
+          <ng-container *ngIf="block.shown">
             <div class="ddp-activity-validation" *ngIf="errorMessage$ | async as errorMsg">
                 <ddp-validation-message [message]="errorMsg">
                 </ddp-validation-message>
             </div>
-          </div>
+          </ng-container>
       </div>`
 })
 export class ActivityQuestionComponent implements OnInit, OnDestroy {
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityQuestion.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityQuestion.component.ts
@@ -18,9 +18,11 @@ import { delay, filter, map, shareReplay, startWith, takeUntil, tap } from 'rxjs
                   [validationRequested]="validationRequested$ | async"
                   (valueChanged)="enteredValue$.next($event)">
           </ddp-activity-answer>
-          <div class="ddp-activity-validation" *ngIf="errorMessage$ | async as errorMsg">
-              <ddp-validation-message [message]="errorMsg">
-              </ddp-validation-message>
+          <div *ngIf="block.shown">
+            <div class="ddp-activity-validation" *ngIf="errorMessage$ | async as errorMsg">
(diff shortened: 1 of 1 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityQuestion.component.ts
```diff
@@ -18,9 +18,11 @@ import { delay, filter, map, shareReplay, startWith, takeUntil, tap } from 'rxjs
                   [validationRequested]="validationRequested$ | async"
                   (valueChanged)="enteredValue$.next($event)">
           </ddp-activity-answer>
-          <div class="ddp-activity-validation" *ngIf="errorMessage$ | async as errorMsg">
-              <ddp-validation-message [message]="errorMsg">
-              </ddp-validation-message>
+          <div *ngIf="block.shown">
```
yufengwng: The actual question itself checks `block.shown` in the `ActivityAnswerComponent`. So easiest fix seems to be adding a check here for the validation messages.

Checking whether the block is actually valid or not seems to be handled already by the `ActivityGroupBlock` model, which is used by the `ConditionalBlock` model. (The bug appears for the conditional question in TestBoston.)

fat23cat (dismissed):

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityQuestion.component.ts
```diff
@@ -18,9 +18,11 @@ import { delay, filter, map, shareReplay, startWith, takeUntil, tap } from 'rxjs
...
+            <div class="ddp-activity-validation" *ngIf="errorMessage$ | async as errorMsg">
+                <ddp-validation-message [message]="errorMsg">
+                </ddp-validation-message>
+            </div>
           </div>
```
MocanaAtBroad: I gave you wrong information during PRTy time. The Angular syntax won't just let you parenthesize the existing ngIf expression (I think issue here is the creation of the errorMsg variable).
Suggestion below gets rid of new `<div>` element. ng-container add nothing to DOM, which I think is the desired behavior.
```suggestion
          <ng-container *ngIf="block.shown">
            <div class="ddp-activity-validation" *ngIf="errorMessage$ | async as errorMsg">
                <ddp-validation-message [message]="errorMsg">
                </ddp-validation-message>
            </div>
        </ng-container>
```

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityQuestion.component.ts
```diff
@@ -18,9 +18,11 @@ import { delay, filter, map, shareReplay, startWith, takeUntil, tap } from 'rxjs
                   [validationRequested]="validationRequested$ | async"
                   (valueChanged)="enteredValue$.next($event)">
           </ddp-activity-answer>
-          <div class="ddp-activity-validation" *ngIf="errorMessage$ | async as errorMsg">
-              <ddp-validation-message [message]="errorMsg">
-              </ddp-validation-message>
+          <div *ngIf="block.shown">
+            <div class="ddp-activity-validation" *ngIf="errorMessage$ | async as errorMsg">
```
yufengwng, line 22: @MocanaAtBroad I tried your suggestion of putting `block.shown` in here but I'm getting TypeScript compile errors. I lack the experience to debug this. So I think I'm going to proceed with my current approach as-is.

Screen Shot 2021-01-11 at 12 27 24 PM

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityQuestion.component.ts
yufengwng: Oops, just saw this now. I'll try this.

yufengwng: Done. `ng-container` did the trick!

MocanaAtBroad (approved):

PR #410, opened 2021-01-14
Ddp 5094 auth

charlottedye:
See DDP-5094 for the ticket.  However, this branch is less obviously related.  This branch sets up the language selector and a preliminary version of the frontend Spanish "translations".  The reason I did this as part of this ticket is that it makes it possible to confirm that the authentication process and emails will be translated correctly.


MocanaAtBroad (commented):
Looks good!

MocanaAtBroad (approved):

PR #411, opened 2021-01-20
User deletion

aleksey-kuznetsov:
These changes introduce service for deleting users. Should be merged once [this](https://github.com/broadinstitute/ddp-study-server/pull/599) is approved & merged.


ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/userDeletionServiceAgent.service.ts
```diff
@@ -0,0 +1,18 @@
...
+@Injectable()
+export class UserDeletionServiceAgent extends UserServiceAgent<void> {
+  public deleteUser(userGuid: string): Observable<void> {
+    const url = `${this.configuration.backendUrl}/pepper/v1/user/${userGuid}`;
+
+    return this.deleteObservable(url).pipe(take(1));
+  }
+
+  protected getBackendUrl(): string {
+    return '';
```
MocanaAtBroad: This method necessary? I think we might be inheriting implementation already?

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/userDeletionServiceAgent.service.ts
```diff
@@ -0,0 +1,18 @@
+import { Injectable } from '@angular/core';
+import { Observable } from 'rxjs';
+import { take } from 'rxjs/operators';
+
+import { UserServiceAgent } from './userServiceAgent.service';
+
+@Injectable()
+export class UserDeletionServiceAgent extends UserServiceAgent<void> {
+  public deleteUser(userGuid: string): Observable<void> {
+    const url = `${this.configuration.backendUrl}/pepper/v1/user/${userGuid}`;
```
MocanaAtBroad: I believe deleteObservable() is already including part of this path. You might just need `/user/${userGuid}` , but take closer look.

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/userDeletionServiceAgent.service.ts
aleksey-kuznetsov: Since `deleteObservable()` uses this method we have to overwrite it to return an empty string.

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/userDeletionServiceAgent.service.ts
aleksey-kuznetsov: The way how user deletion works is that only proxy user is allowed to delete it's managed users. Therefore we have to manually pass the `userGuid` since underlying `UserServiceAgent` dynamically substitutes `userGuid` part in the URL with either from proxy user or a participant if one is set.
I just thought that it would be more logical to extend the `UserServiceAgent` since we are dealing with user entity.

MocanaAtBroad: I see your point.
I think what you want is to inherit from ServiceAgent . Will set credentials for http call and you can inherit the correct base url; without the `/user` part of the path.
Also maybe we should rename this class so that name is not exclusively about deletion? Unfortunately we are using UserServiceAgent for sort of another purpose. Maybe call this something like UserManagementService ?

P.S. I don't know how why we started using the ServiceAgent suffix . Service would have sufficed.

aleksey-kuznetsov: Sure, this sounds reasonable.

MocanaAtBroad (approved):

PR #422, opened 2021-01-26
DD-5566 fix language code related errors

MocanaAtBroad:
In addition to error reported in ticket, also tried to address error reported to console as they might be related.
Screen Shot 2021-01-26 at 1 09 44 PM


fat23cat (approved):
Works good!
And then we need to bring these changes into `develop`

yufengwng:
This bugfix has been released to production as part of `v1.9.9`. I didn't see this in the `develop` so I went ahead and cherry-picked it over to `develop`. This should be all set now.

PR #423, opened 2021-01-27
Fix linter warnings/errors for DDP-SDK

alextanlan:


MocanaAtBroad (approved):
LGTM

MocanaAtBroad (commented):
When I run `ng lint ddp-sdk` only thing that comes back now are some deprecation warnings.
Since you are already at it and there is only a handful of them, perhaps we should just clear those up too?

alextanlan:
> When I run ng lint ddp-sdk only thing that comes back now are some deprecation warnings.
> Since you are already at it and there is only a handful of them, perhaps we should just clear those up too?

At the beginning, I tried to fix all linter warnings/errors in one PR. Then I realized the PR was becoming an enormous one. Any next changes could affect previous ones, there were some failed tests as well. 

I decided to divide fixes into several PR's according to the rules:
- firstly, fix syntactic warnings/errors, which are not connected to any logic. One PR for one study (brain, angio, etc.)
- then fix logical warnings/errors (mostly connected with deprecated stuff, e.g. Rxjs operarors)
- finally, when the linter log is clear, add linter check into circleCI and as a pre-push hook locally. 

That way seems to me more clear and easier to code-review and test. Also it can be more convinient (in case of any possible merge-conflicts) to resolve smaller partial changes instead of a big common one.

PR #431, opened 2021-01-28
Fix linter warnings/errors for DDP-OSTEO

alextanlan:


ddp-workspace/projects/ddp-osteo/src/app/app.module.ts
```diff
@@ -117,15 +116,6 @@ export function translateFactory(translate: TranslateService, injector: Injector
...
 
-export class MyHammerConfig extends HammerGestureConfig {
-  public buildHammer(element: HTMLElement): Hammer {
-    const hammer = new Hammer(element, {
-      touchAction: 'pan-y'
-    });
-    return hammer;
-  }
-}
-
```
MocanaAtBroad, line 128: We sure about about this one? I don't recall where it gets used. Maybe @fat23cat recalls?

fat23cat: We use it for supporting swipe gestures on mobile devices for gallery widget on the osteo homepage. Aleksei just moved this class into separate file.

alextanlan: Yes, it's correct. I only moved `class MyHammerConfig ` from `app.module.ts` as to have it in a separate file.

fat23cat (approved):

PR #442, opened 2021-02-03
Datepicker required validation

aleksey-kuznetsov:
These changes make sure that once user clears out date picker value manually the `null` value is emitted and validation is triggered.


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activityDateAnswer.component.ts
```diff
@@ -33,14 +33,16 @@ export class ActivityDateAnswer {
     constructor(private logger: LoggingService) { }
 
     public handleChange(value: DatePickerValue | null): void {
-        if (value == null) {
-            return;
+        let dateValue: DatePickerValue | null = value;
```
MocanaAtBroad: This assignment is a bit odd. Intent clearer perhaps if just:
```suggestion
        let dateValue: DatePickerValue | null = null;
```

MocanaAtBroad (dismissed):

fat23cat (approved):

PR #444, opened 2021-02-04
RGP | Add COVID banner

aleksey-kuznetsov:


charlottedye (changes requested):
The banner itself looks good but in some places, the main header portion is too short.  This is what it looks like in Gen2: 
orig.  This is what it looks like in this pull request: 
pr.  (If you give me your GitHub name, I can also give you access to the Gen2 repos if that would be helpful.  The best way to compare to Gen2 is to go to dev.datadonationplatform.org).

aleksey-kuznetsov:
I'll refactor header component a bit since we have some strange behavior when it is resized and also address this issue that you've mentioned.

aleksey-kuznetsov:
These new changes include header refactoring (bug fixing & better responsiveness).
Besides, I've moved language selector to the header as well, as described in DDP-5506 (sorry for making those changes in this branch, but changed parts are connected with each other and it'd be easier to make them in one go).

charlottedye (approved):
This is much better.  The only thing I noticed was that the "for your physician" link in the header appears not to work, but I checked and this is also the case in dev.  I'll go ahead and make a ticket for that.  Nice work fitting the language selector in the header, too!

PR #445, opened 2021-02-08
DDP-5590: Add new activity block ('ACTIVITY') handling

alextanlan:


ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityActivityBlock.ts
```diff
@@ -0,0 +1,20 @@
...
+    Modal = 'modal'
+}
+
+export class ActivityActivityBlock extends ActivityBlock {
+    public title: string;
+    public renderHint: ChildActivityRendering;
+    public activityCode: string;
+    public allowMultiple: boolean;
+    public addButtonText: string;
+    public instances: any[];
```
softcat23, line 15:
```suggestion
    public instances: ActivityInstance[];
```
Also, pay attention that you need to update `ActivityInstance` interface, it contains a new `canDelete` property.
https://github.com/broadinstitute/ddp-study-server/blob/develop/pepper-apis/docs/specification/src/components/schemas.yml#L193

fat23cat:
```suggestion
    public instances: ActivityInstance[];
```
Also, pay attention that you need to update `ActivityInstance` interface, it contains a new `canDelete` property.
https://github.com/broadinstitute/ddp-study-server/blob/develop/pepper-apis/docs/specification/src/components/schemas.yml#L193

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/activityConverter.service.ts
```diff
@@ -150,24 +155,49 @@ export class ActivityConverter {
...
+          {
+              type: BlockType.Content.toUpperCase(),
+              func: (input) => this.convertContentBlock(input)
+          },
+          {
+              type: BlockType.Question.toUpperCase(),
+              func: (input) => this.questionConverter.buildQuestionBlock(input.question, input.displayNumber)
+          },
+          {
+              type: 'COMPONENT', // TODO: check - where is this type enum ?
```
fat23cat: I think you can update `BlockType` enum and add this block type there.

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/activityQuestionConverter.service.ts
```diff
@@ -210,4 +111,123 @@ export class ActivityQuestionConverter {
...
+
+    private getCompositeBlock(questionJson: any): ActivityCompositeQuestionBlock {
+        const newBlock = new ActivityCompositeQuestionBlock();
+        newBlock.addButtonText = questionJson.addButtonText;
+        newBlock.allowMultiple = questionJson.allowMultiple;
+        newBlock.childOrientation = questionJson.childOrientation;
+        newBlock.allowMultiple && (newBlock.additionalItemText = questionJson.additionalItemText);
+        newBlock.children = (questionJson.children as ActivityQuestionBlock<any>[])
+            .map(childInputBlock => this.buildQuestionBlock(childInputBlock, null))
+            .filter(block => block != null);
```
fat23cat:
```suggestion
            .filter(block => block !== null);
```

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/activityQuestionConverter.service.ts
```diff
@@ -210,4 +111,123 @@ export class ActivityQuestionConverter {
             return convertedGroup;
         });
     }
+
+    private initQuestionBuilders(): void {
+      this.questionBuilders = [
+          {
+              type: QuestionType.Boolean.toUpperCase(),
```
MocanaAtBroad: If we are going to refactor, perhaps we should just rename the enum values to their uppercase values? The values we get from server are uppercase and does not help us to use them in lowercase. E.g.:
Screen Shot 2021-02-08 at 11 19 03 AM

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityActivityBlock.ts
```diff
@@ -0,0 +1,20 @@
+import { ActivityBlock } from './activityBlock';
+import { BlockType } from './blockType';
+
+enum ChildActivityRendering {
+    Embedded = 'embedded',
+    Modal = 'modal'
```
MocanaAtBroad: Enum values are incorrect:
Screen Shot 2021-02-08 at 11 04 45 AM

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityActivityBlock.ts
```diff
@@ -0,0 +1,20 @@
+import { ActivityBlock } from './activityBlock';
+import { BlockType } from './blockType';
+
+enum ChildActivityRendering {
```
MocanaAtBroad: Name of enum seems to indicate it might be an action of some sort.
Would suggest renaming to "ActivityRenderHintType" or something else like that.

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityActivityBlock.ts
alextanlan: Thank you! 
I was just confused that ActivityInstance and Activity.Summary are a bit different things.

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/activityConverter.service.ts
alextanlan: Good point, thanks!

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/activityQuestionConverter.service.ts
alextanlan: Agree, fixed.

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/activityQuestionConverter.service.ts
alextanlan: You are right.
Fixed in 36842b840f002035e924d49ec926a43804a46e98 (for blockType and questionType)

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityActivityBlock.ts
alextanlan: My bad, fixed.

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityActivityBlock.ts
alextanlan: renamed.

fat23cat (commented):
Could you please check `ActivitySectionComponent`? I've observed some TS issues:
image
Please, consider `GroupBlockList` refactoring, I guess, we can use enum here, instead of strings:
image

alextanlan:
> Please, consider GroupBlockList refactoring, I guess, we can use enum here, instead of strings:

Good point! 
It's the case when we have to avoid 'magic' words (and use contants/enums instead of). 
Fixed in 5ed9c3989bc6f529f2235626bdfd42437c42579a

alextanlan:
> Could you please check ActivitySectionComponent? I've observed some TS issues:

Checked. It is alright.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/groupBlockList.component.ts
```diff
@@ -77,6 +78,7 @@ export class GroupBlockList {
     @Output() valueChanged: EventEmitter<AnswerValue> = new EventEmitter();
 
     public readonly LIST_STYLE = ListStyleHint;
+    public readonly BLOCK_TYPE = BlockType;
```
MocanaAtBroad, line 81: This looked weird to me at first and now I see this is a nice idea!

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityActivityBlock.ts
```diff
@@ -0,0 +1,21 @@
+import { ActivityBlock } from './activityBlock';
+import { BlockType } from './blockType';
+import { ActivityInstance } from 'ddp-sdk';
```
MocanaAtBroad: Importing library **_might_** be OK, but safer/more consistent to do a relative part import like the others.

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityActivityBlock.ts
```diff
@@ -0,0 +1,21 @@
...
+
+enum ActivityRenderHintType {
+    Embedded = 'EMBEDDED',
+    Modal = 'MODAL'
+}
```
MocanaAtBroad: We have been giving this type of definitions their own files. We should keep with that approach.

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityActivityBlock.ts
alextanlan: My IDE decided it instead of me 🙂
Fixed in d076b6949bc6943d9b7bd86d671b35b5e37ee643.

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityActivityBlock.ts
alextanlan: Got it, fixed in a8003463771fa957fff2a8cefafc757b2b74ee3d

alextanlan: In addition, what are your thoughts about a structuring files into directories ? 
E.g. `ddp-sdk\src\lib\models\activity` and `ddp-sdk\src\lib\models`. Each of them has more than 50 files, it is quite confusing.

My suggestion is to sort those files into some sub-folders, either '**_by features_**' or '**_by file types_**' (../enums, ../interfaces)

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityActivityBlock.ts
MocanaAtBroad: I have same problem with IDE (I use IntelliJ). I think you are using WebStorm which is pretty much the same thing.
If you can figure out how to stop the IDE from doing these incorrect imports please let me know!

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityActivityBlock.ts
MocanaAtBroad: I am personally not bothered with the large folder as I tend not to click/navigate through folders. I generally look for classes and files by typing their name or part of name, and would continue to do that regardless of folder arrangement.
But if you feel strongly that we should split "model" into more subfolders, sure. I am not opposed to idea.
Now, I do have a strong dislike for separating enums from interfaces (I see them as serving same purpose: defining the shape of objects and data), but if you can come up with a feature-based folder scheme that make sense, go for it.

MocanaAtBroad (approved):

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityActivityBlock.ts
alextanlan: Ok, let me think about it.

PR #452, opened 2021-02-15
DDP-5635: Progress indicator for family history

alextanlan:


MocanaAtBroad (approved):
Code looks good. And the test is awesome.
How did you go about seeing the code in action?
This question should not hold up merging as we will get opportunity to exercise code when we have more stuff integrated.

alextanlan:
I checked it on the landing page of ddp-osteo study in testing purpose.
And it works as planned.

PR #453, opened 2021-02-17
DDP-5538: Auth0 UI Improvements

aleksey-kuznetsov:
Introduced `EmailVerifiedComponent` which is used as a redirect page once an email is verified. In turn, this component will redirect user to login page if everything is alright and to an error page if there is something wrong.


charlottedye (approved):
This is great!

PR #455, opened 2021-02-22
DDP-5470: Loading Optimization

EdPokhvalin:
Unfortunately, the ticket has not been closed, but some of the changes can already be reviewed and used.
The main changes have affected the twitter widget, now it is completely lazy. I rolled back similar changes for instagram, because its widget is lazy by default. There is also an idea to bring the logic for tracking intersections to the service, but I don't see any opportunities to reuse it yet.

I'm still having problems with merge, so there's a bunch of extra commits here. My changes are in _twitter-timeline-widget.component.ts_, _index.html_ (in all applications) and _gallery.component.html_.

---
This change is [Reviewable](https://reviewable.io/reviews/broadinstitute/ddp-angular/455)


MocanaAtBroad:
Yikes! Lots of changed files!
I suggest creating a brand new branch off `develop` and use `git cherry-pick` to bring over the very specific commits that are relevant.

PR #458, opened 2021-02-24
Add support for confirmation field in text questions

aleksey-kuznetsov:
These changes include support for confirmation field for text questions if one is configured.

Changes:
- Created new `ActivityTextInput` component
- Moved previous functionality for text questions to newly created component
- Implemented support for `confirmEntry` field in text questions
- Implemented support for `confirmPrompt` field in text questions
- Moved tests for previous functionality related to text questions to newly created component
- Added relevant tests for new functionality

Note:
If a question is configured in a way so that `confirmEntry` is `true`, then a confirmation field will appear underneath original one. This confirmation field does not include any validation but the whole `FormGroup` is valid once all necessary validation for original field has passed and confirmation field is filled with the same value as the original field.


charlottedye (changes requested):
The behavior with the error messages is a little odd.  If errors are displayed and the main phone number field is valid but the confirm phone number field doesn't match, both error messages are displayed.  This ends up looking confusing.
Screen Shot 2021-02-25 at 1 22 39 PM

aleksey-kuznetsov:
In the current implementation a value for the answer is updated once user enters a valid value in both fields, so whatever user enters in the main field is not propagated until confirmation field is filled with the same valid value.

What would you suggest?

aleksey-kuznetsov:
@charlottedye any suggestions? We could probably tweak question's validation message and remove `mismatchMessage`.

charlottedye:
Could we maybe do something with CSS to hide the "Please enter a valid..." message when the mismatch message is shown?  I haven't totally thought this through but maybe we add a CSS class to the main validation message with display: none if the control is invalid.  Not 100% sure that approach would work but it seems like something along those lines should.

aleksey-kuznetsov:
@charlottedye to be honest, I don't think this will work out. Currently, behavior of this component (in terms of validation) is the same as for `ActivityEmailInput` component.

Since we are going to use it in RGP for phone number verification & validation we could leave `mismatchMessage` field for this question empty and rephrase validation message for `required` rule like "Please enter and re-enter a phone number with exactly and only 10 digits" or something similar. This way error message underneath confirmation field won't be visible.

charlottedye:
That seems reasonable.  Let's ask in the RGP channel what the message text should be.

PR #464, opened 2021-03-04
AT deployment fixes

aleksey-kuznetsov:
Previously, I made some changes to CircleCI config where I'd rely on `pipeline.parameters.study_key` in order to determine whether an additional command should be executed (needed for AT). I switched to passing the `study_key` as a parameter to the `build_app` command since we don't always have `pipeline.parameters.study_key` set (e.g. in `nightly` workflow).


aleksey-kuznetsov:
@MocanaAtBroad can you please take a look?

MocanaAtBroad (approved):
Looks good to me, Aleksey!

PR #470, opened 2021-03-10
DDP-5645: Enrollment survey UI improvements

aleksey-kuznetsov:


charlottedye (changes requested):
I noticed a number of issues.  I think correcting some of these isn't feasible at the moment (and I just reviewed a backend ticket involving the survey so it's possible that ticket will address some of these.  If not, there may be ways to address some of these that involve changes to the study builder.  So this might involve some coordination with Yuriy) but see how it goes and let me know which ones aren't practical to fix.  Some of these comments will make more sense if you look at the Pepper implementation and Gen2 (dev.raregenomes.org) side by side.

General:
-The validation messages are a little far down from the question.  There also isn't much space between the validation message and the next question in some cases. (I know the validation stuff is hard to work with.  Do whatever makes things easiest to read.  It will probably still not look great but that's fine for now)
-The detail input for checklists has an indicator that displays how many characters remain.  Ideally we wouldn't display this.

Page 1:
-"Your telephone number *" is bold and is prompt-style rather than placeholder-style 
-Dropdowns on this page look un-styled where Gen2 has them styled.  They look fine on other pages.
-There is hardly any spacing between "For each genetic diagnosis, please also provide information as to why the geneticist does not think these results fully explain the condition" and the last option of "Have any genetic diagnoses been made?".  Additionally, the "For each genetic diagnosis..." part isn't bold.

Page 2:
-For the test question, when certain options are selected, a number of related questions appear below the question (this part is fine).  The name of the section (e.g. "Chromosomal microarray") isn't bolded and there isn't much space between this and whatever is above.  The subquestion "Which other family members had this test?" is bolded and there is plenty of space between this question and the field above it.  This looks confusing.

Page 3:
-The current age question for siblings, children, and other affected relatives is a little too close to the question above.
-I think this might not be supported, in which case you can ignore this, but ideally the "Please describe" section would be larger.

Otherwise this looks great!

charlottedye:
One more comment: I'd suggest waiting until DDP-5646 is merged in before trying to fix any of these things.  That will probably make things simpler.

aleksey-kuznetsov:
@charlottedye thanks for the review!
Let's wait for [DDP-5646](https://broadinstitute.atlassian.net/browse/DDP-5646) and then see what's missing since we have cooperated with @YuriyPavlovBI to fix styling issues which are related to backend configuration.

aleksey-kuznetsov:
@charlottedye regarding the dropdowns: dropdown with single select is rendered as a native `select` tag. Shall we switch to the `mat-select` in this case? This can be a rather significant change for other studies.

aleksey-kuznetsov:
@charlottedye I've addressed most of the issues you've found.
1. Played around with validation errors a bit
2. Removed indicator which showed the count of left characters
3. Moved "Your Telephone Number *" from prompt template to placeholder template
4. Added spacings for "age" questions on the 3rd page
5. Adjusted height for textareas

It's better to take a look at these changes in conjunction with backend [PR](https://github.com/broadinstitute/ddp-study-server/pull/677).

charlottedye (approved):
Looks good!

PR #483, opened 2021-03-25
DDP-5720: Save the last visited step

EdPokhvalin:
This change is [Reviewable](https://reviewable.io/reviews/broadinstitute/ddp-angular/483)


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -204,6 +204,8 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
     // one entry per section (header, body, and footer respectively)
     private embeddedComponentsValidationStatus: boolean[] = new Array(3).fill(true);
     private readonly LOG_SOURCE = 'ActivityComponent';
+    private readonly COMPLETE = 'COMPLETE';
```
MocanaAtBroad: Please do some refactoring.
Appears `ddp-angular/ddp-workspace/projects/ddp-rarex/src/app/constants/activity-status-codes.ts`
contains an enum.
Please move enum, into SDK, rename file to use camel case and use this enum instead of type 'string' or readonly var.

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityForm.ts
```diff
@@ -16,6 +16,7 @@ export class ActivityForm {
     public lastUpdated?: Date;
     public validationState: ReplaySubject<boolean> = new ReplaySubject();
     public sectionIndex: number | null;
+    public statusCode: string;
```
MocanaAtBroad: As part of refactoring, please convert this to the ActivityStatusCode enum mentioned above.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -472,4 +476,25 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
...
+          take(1),
+          tap(() => this.shouldSaveLastStep = this.model.statusCode !== this.COMPLETE &&
+            this.config.usesVerticalStepper.includes(this.model.activityCode))
+        )
+        .subscribe(() => {
+          this.currentSectionIndex = this.shouldSaveLastStep
+            ? this.model.sectionIndex || 0
+            : 0;
+        });
+    }
```
MocanaAtBroad, line 503: A few suggestions here:
1. The prefix "get" in a method generally means that it is a property "getter". This is not the intent of this method and it's return type is `void`. I would suggest renaming to something like `initStepperState` or something to that effect.
2. The reference to this.COMPLETE can hopefully be replace with the enum mentioned earlier.
3. Looks like this method intent is to set 2 instance variables, but one is being set in a `tap` and the other in the `subscribe`. Might be clearer if `this.currentSectionIndex` is set in another `tap` and leave the `subscribe` without an argument.
4. Does not look to me that checking the value of `this.shouldSaveLastStep` to decide between and 0 and `model.sectionIndex` helps. We are either going to look at it or not? Consider simplifying to something like:
    `this.currentSectionIndex = this.model.sectionIndex === null ? 0 :  this.model.sectionIndex`

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -472,4 +476,25 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
             });
         }
     }
+
+    private saveLastVisitedSectionIndex(sectionIndex: number): void {
+      if (this.shouldSaveLastStep && sectionIndex > this.model.sectionIndex) {
```
alextanlan: It would be nice to have the same indent for new code lines like the whole file has.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -472,4 +476,25 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
...
+        this.serviceAgent.saveLastVisitedActivitySection(this.studyGuid, this.activityGuid, this.currentSectionIndex);
+      }
+    }
+
+    private getFirstSectionIndex(): void {
+      this.getIsLoaded$()
+        .pipe(
+          filter(value => !!value),
```
alextanlan: Or just `filter(Boolean)`.
Even more, I guess `filter(value => value)` would be quite enough as `getIsLoaded$` already `boolean`.
It is not critical, anyway.

EdPokhvalin (commented):
> *[Reviewable](https://reviewable.io/reviews/broadinstitute/ddp-angular/483)* status: 0 of 7 files reviewed, 5 unresolved discussions (waiting on @alextanlan and @MocanaAtBroad)

___
*[ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts, line 207 at r1](https://reviewable.io/reviews/broadinstitute/ddp-angular/483#-MWiKnzuKSaSf3Fd0yWW-r1-207:-MWiND9o7vlngBMNOjSy:bb6un1j) ([raw file](https://github.com/broadinstitute/ddp-angular/blob/a17ac9e3810a6f3a6c5f236faf8e660984589f12/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts#L207)):*
<details><summary><i>Previously, MocanaAtBroad wrote…</i></summary><blockquote>

Please do some refactoring.
Appears `ddp-angular/ddp-workspace/projects/ddp-rarex/src/app/constants/activity-status-codes.ts`
contains an enum.
Please move enum, into SDK, rename file to use camel case and use this enum instead of type 'string' or readonly var.
</blockquote></details>

Done
___
*[ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts, line 481 at r1](https://reviewable.io/reviews/broadinstitute/ddp-angular/483#-MWiKnzuKSaSf3Fd0yWW-r1-481:-MWiN3_W4EyWEtZFWwr-:b-tr9yxy) ([raw file](https://github.com/broadinstitute/ddp-angular/blob/a17ac9e3810a6f3a6c5f236faf8e660984589f12/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts#L481)):*
<details><summary><i>Previously, alextanlan (Alex Maratkanov) wrote…</i></summary><blockquote>

It would be nice to have the same indent for new code lines like the whole file has.
</blockquote></details>

My mistake, fixed it
___
*[ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts, line 489 at r1](https://reviewable.io/reviews/broadinstitute/ddp-angular/483#-MWiKnzuKSaSf3Fd0yWW-r1-489:-MWiM3nf1kOVlsh8LXGN:bw5qau9) ([raw file](https://github.com/broadinstitute/ddp-angular/blob/a17ac9e3810a6f3a6c5f236faf8e660984589f12/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts#L489)):*
<details><summary><i>Previously, alextanlan (Alex Maratkanov) wrote…</i></summary><blockquote>

Or just `filter(Boolean)`.
Even more, I guess `filter(value => value)` would be quite enough as `getIsLoaded$` already `boolean`.
It is not critical, anyway.
</blockquote></details>

Fixed too, but in my opinion double negation looks more obvious
___
*[ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts, line 499 at r1](https://reviewable.io/reviews/broadinstitute/ddp-angular/483#-MWiKnzuKSaSf3Fd0yWW-r1-499:-MWiPC7p-qzteG9abxb1:b-e4cnih) ([raw file](https://github.com/broadinstitute/ddp-angular/blob/a17ac9e3810a6f3a6c5f236faf8e660984589f12/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts#L499)):*
<details><summary><i>Previously, MocanaAtBroad wrote…</i></summary><blockquote>

A few suggestions here:
1. The prefix "get" in a method generally means that it is a property "getter". This is not the intent of this method and it's return type is `void`. I would suggest renaming to something like `initStepperState` or something to that effect.
2. The reference to this.COMPLETE can hopefully be replace with the enum mentioned earlier.
3. Looks like this method intent is to set 2 instance variables, but one is being set in a `tap` and the other in the `subscribe`. Might be clearer if `this.currentSectionIndex` is set in another `tap` and leave the `subscribe` without an argument.
4. Does not look to me that checking the value of `this.shouldSaveLastStep` to decide between and 0 and `model.sectionIndex` helps. We are either going to look at it or not? Consider simplifying to something like:
    `this.currentSectionIndex = this.model.sectionIndex === null ? 0 :  this.model.sectionIndex`


</blockquote></details>

1-3: fixed
As for the last suggestion, we need to check `this.shouldSaveLastStep` because `model.sectionIndex` may be set to non-zero value but activity status can be `COMPLETE`. In that case we should use 0 as the first step. And also we need to check that `sectionIndex` is set (it can be `null` or probably even `undefined`)
___
*[ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityForm.ts, line 19 at r1](https://reviewable.io/reviews/broadinstitute/ddp-angular/483#-MWiKnzvj0CTBL-F-9Wj-r1-19:-MWiM1-x6SH1IpuFX0Wu:bb6un1j) ([raw file](https://github.com/broadinstitute/ddp-angular/blob/a17ac9e3810a6f3a6c5f236faf8e660984589f12/ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityForm.ts#L19)):*
<details><summary><i>Previously, MocanaAtBroad wrote…</i></summary><blockquote>

As part of refactoring, please convert this to the ActivityStatusCode enum mentioned above.
</blockquote></details>

Done

alextanlan (commented):
> *[Reviewable](https://reviewable.io/reviews/broadinstitute/ddp-angular/483)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @EdPokhvalin and @MocanaAtBroad)

___
*[ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts, line 489 at r1](https://reviewable.io/reviews/broadinstitute/ddp-angular/483#-MWiKnzuKSaSf3Fd0yWW-r1-489:-MWiZyxL-Mtyy-m9hxiQ:b-ly8e71) ([raw file](https://github.com/broadinstitute/ddp-angular/blob/a17ac9e3810a6f3a6c5f236faf8e660984589f12/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts#L489)):*
<details><summary><i>Previously, EdPokhvalin wrote…</i></summary><blockquote>

Fixed too, but in my opinion double negation looks more obvious
</blockquote></details>

It was just a tip. Sometimes a code-review can be used as to share knowledge between developers.

MocanaAtBroad (changes requested):
> *[Reviewable](https://reviewable.io/reviews/broadinstitute/ddp-angular/483)* status: 0 of 7 files reviewed, 7 unresolved discussions (waiting on @alextanlan, @EdPokhvalin, and @MocanaAtBroad)

___
*[ddp-workspace/projects/ddp-rarex/src/app/components/dashboard/dashboard.component.ts, line 16 at r2](https://reviewable.io/reviews/broadinstitute/ddp-angular/483#-MWiye_p1daHFp7lq5Zx:-MWiye_p1daHFp7lq5Zy:b-p9arvf) ([raw file](https://github.com/broadinstitute/ddp-angular/blob/2553cf843de8d0ef42197136fd567f0602556fba/ddp-workspace/projects/ddp-rarex/src/app/components/dashboard/dashboard.component.ts#L16)):*
> ```XML
> 
> import { ActivityCodes } from '../../constants/activity-codes';
> import { ActivityStatusCodes } from '../../../../../ddp-sdk/src/lib/models/activity/activityStatusCodes';
> ```

Can't refer to code in ddp-sdk from other Angular projects using relative path. You probably have to export ActivityStatusCodes from ddp-sdk so you can end up with with an import that looks like:
```
import { ActivityStatusCodes } from 'ddp-sdk';
```
___
*[ddp-workspace/projects/ddp-rarex/src/app/components/rarex-activity-page/rarex-activity-page.component.ts, line 23 at r2](https://reviewable.io/reviews/broadinstitute/ddp-angular/483#-MWizP-qA5zg9HsiZWih:-MWizP-qA5zg9HsiZWii:br9sbew) ([raw file](https://github.com/broadinstitute/ddp-angular/blob/2553cf843de8d0ef42197136fd567f0602556fba/ddp-workspace/projects/ddp-rarex/src/app/components/rarex-activity-page/rarex-activity-page.component.ts#L23)):*
> ```XML
> import { CurrentActivityService } from '../../services/current-activity.service';
> import { ActivityCodes } from '../../constants/activity-codes';
> import { ActivityStatusCodes } from '../../../../../ddp-sdk/src/lib/models/activity/activityStatusCodes';
> ```

Sane comment from above applies here.
___
*[ddp-workspace/projects/ddp-rarex/src/app/components/workflow-progress/workflow-progress.component.ts, line 5 at r2](https://reviewable.io/reviews/broadinstitute/ddp-angular/483#-MWizTzWAdNfF3nBSGiE:-MWizTzWAdNfF3nBSGiF:bs4q9op) ([raw file](https://github.com/broadinstitute/ddp-angular/blob/2553cf843de8d0ef42197136fd567f0602556fba/ddp-workspace/projects/ddp-rarex/src/app/components/workflow-progress/workflow-progress.component.ts#L5)):*
> ```XML
> import { ActivityInstance } from 'ddp-sdk';
> 
> import { ActivityStatusCodes } from '../../../../../ddp-sdk/src/lib/models/activity/activityStatusCodes';
> ```

Ditto.
___
*[ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts, line 489 at r1](https://reviewable.io/reviews/broadinstitute/ddp-angular/483#-MWiKnzuKSaSf3Fd0yWW-r1-489:-MWih1cm28kw3C60xZeo:b-wqhdow) ([raw file](https://github.com/broadinstitute/ddp-angular/blob/a17ac9e3810a6f3a6c5f236faf8e660984589f12/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts#L489)):*
<details><summary><i>Previously, alextanlan (Alex Maratkanov) wrote…</i></summary><blockquote>

It was just a tip. Sometimes a code-review can be used as to share knowledge between developers.
</blockquote></details>

OK. This is very interesting way to write it out. It's good!
___
*[ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts, line 499 at r1](https://reviewable.io/reviews/broadinstitute/ddp-angular/483#-MWiKnzuKSaSf3Fd0yWW-r1-499:-MWihZXyB0dX3J6mPsII:biopwwc) ([raw file](https://github.com/broadinstitute/ddp-angular/blob/a17ac9e3810a6f3a6c5f236faf8e660984589f12/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts#L499)):*
<details><summary><i>Previously, EdPokhvalin wrote…</i></summary><blockquote>

1-3: fixed
As for the last suggestion, we need to check `this.shouldSaveLastStep` because `model.sectionIndex` may be set to non-zero value but activity status can be `COMPLETE`. In that case we should use 0 as the first step. And also we need to check that `sectionIndex` is set (it can be `null` or probably even `undefined`)
</blockquote></details>

Looks like sectionIndex value depends more directly on the status of the activity. Consider rewriting code to make that more explicit then (code suggests that it depends on value of `shouldSaveLastStep` which I think is why I was confused reading this even though functionally it is correct).

EdPokhvalin (commented):
> *[Reviewable](https://reviewable.io/reviews/broadinstitute/ddp-angular/483)* status: 0 of 7 files reviewed, 7 unresolved discussions (waiting on @alextanlan and @MocanaAtBroad)

___
*[ddp-workspace/projects/ddp-rarex/src/app/components/dashboard/dashboard.component.ts, line 16 at r2](https://reviewable.io/reviews/broadinstitute/ddp-angular/483#-MWiye_p1daHFp7lq5Zx:-MWy18s71Mtnr3FRVRjh:bb6un1j) ([raw file](https://github.com/broadinstitute/ddp-angular/blob/2553cf843de8d0ef42197136fd567f0602556fba/ddp-workspace/projects/ddp-rarex/src/app/components/dashboard/dashboard.component.ts#L16)):*
<details><summary><i>Previously, MocanaAtBroad wrote…</i></summary><blockquote>

Can't refer to code in ddp-sdk from other Angular projects using relative path. You probably have to export ActivityStatusCodes from ddp-sdk so you can end up with with an import that looks like:
```
import { ActivityStatusCodes } from 'ddp-sdk';
```
</blockquote></details>

Done
___
*[ddp-workspace/projects/ddp-rarex/src/app/components/rarex-activity-page/rarex-activity-page.component.ts, line 23 at r2](https://reviewable.io/reviews/broadinstitute/ddp-angular/483#-MWizP-qA5zg9HsiZWih:-MWy186Y54NKwgClUu0C:bb6un1j) ([raw file](https://github.com/broadinstitute/ddp-angular/blob/2553cf843de8d0ef42197136fd567f0602556fba/ddp-workspace/projects/ddp-rarex/src/app/components/rarex-activity-page/rarex-activity-page.component.ts#L23)):*
<details><summary><i>Previously, MocanaAtBroad wrote…</i></summary><blockquote>

Sane comment from above applies here.
</blockquote></details>

Done
___
*[ddp-workspace/projects/ddp-rarex/src/app/components/workflow-progress/workflow-progress.component.ts, line 5 at r2](https://reviewable.io/reviews/broadinstitute/ddp-angular/483#-MWizTzWAdNfF3nBSGiE:-MWy17T9-uBTuLcLnMeD:b-886gsx) ([raw file](https://github.com/broadinstitute/ddp-angular/blob/2553cf843de8d0ef42197136fd567f0602556fba/ddp-workspace/projects/ddp-rarex/src/app/components/workflow-progress/workflow-progress.component.ts#L5)):*
<details><summary><i>Previously, MocanaAtBroad wrote…</i></summary><blockquote>

Ditto.
</blockquote></details>

Fixed
___
*[ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts, line 499 at r1](https://reviewable.io/reviews/broadinstitute/ddp-angular/483#-MWiKnzuKSaSf3Fd0yWW-r1-499:-MWyCQLy36iua9Ab6xLw:b-yphdu2) ([raw file](https://github.com/broadinstitute/ddp-angular/blob/a17ac9e3810a6f3a6c5f236faf8e660984589f12/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts#L499)):*
<details><summary><i>Previously, MocanaAtBroad wrote…</i></summary><blockquote>

Looks like sectionIndex value depends more directly on the status of the activity. Consider rewriting code to make that more explicit then (code suggests that it depends on value of `shouldSaveLastStep` which I think is why I was confused reading this even though functionally it is correct).
</blockquote></details>

I tried to simplify this logic, hopefully it will be easier to understand now

MocanaAtBroad (changes requested):
> *[Reviewable](https://reviewable.io/reviews/broadinstitute/ddp-angular/483)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @alextanlan and @EdPokhvalin)

*[a discussion](https://reviewable.io/reviews/broadinstitute/ddp-angular/483#-MX3a3CF8X9SOvZT6UCB:-MX3a3CF8X9SOvZT6UCC:b-2phyhb) (no related file):*
Are you SURE we are saving the current section?

---

MocanaAtBroad (changes requested):
> *[Reviewable](https://reviewable.io/reviews/broadinstitute/ddp-angular/483)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @alextanlan and @EdPokhvalin)

*[a discussion](https://reviewable.io/reviews/broadinstitute/ddp-angular/483#-MX3aP4c4vv05Fnt8lou:-MX3aP4c4vv05Fnt8lov:bgamdmw) (no related file):*
The current section is not being saved. Please test it locally.

---

EdPokhvalin (commented):
> *[Reviewable](https://reviewable.io/reviews/broadinstitute/ddp-angular/483)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @alextanlan and @EdPokhvalin)

*[a discussion](https://reviewable.io/reviews/broadinstitute/ddp-angular/483#-MX3a3CF8X9SOvZT6UCB:-MX7V2UxCCKKVxF_aCqI:b9weshk) (no related file):*
<details><summary><i>Previously, MocanaAtBroad wrote…</i></summary><blockquote>

Are you SURE we are saving the current section?
</blockquote></details>

Yes, I realized that, first of all, I forgot we get an observable when calling a `saveLastVisitedActivitySection` method. In second, I want to check initialization of the index too. I think I also saw a problem there.

---

EdPokhvalin (commented):
> *[Reviewable](https://reviewable.io/reviews/broadinstitute/ddp-angular/483)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @alextanlan and @MocanaAtBroad)

*[a discussion](https://reviewable.io/reviews/broadinstitute/ddp-angular/483#-MX3aP4c4vv05Fnt8lou:-MX7YtD42tII1plHRrfn:b-snxf4k) (no related file):*
<details><summary><i>Previously, MocanaAtBroad wrote…</i></summary><blockquote>

The current section is not being saved. Please test it locally.
</blockquote></details>

I fixed that and also fixed a little problem with the vertical stepper (now it will be initialized when we initialize the section index too).

---

MocanaAtBroad (approved):
> Reviewed 1 of 3 files at r1, 2 of 6 files at r2, 4 of 5 files at r3, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/broadinstitute/ddp-angular/483)* status: all files reviewed, 3 unresolved discussions (waiting on @alextanlan, @EdPokhvalin, and @MocanaAtBroad)

*[a discussion](https://reviewable.io/reviews/broadinstitute/ddp-angular/483#-MX7xyiT82d2XmXFgAPS:-MX7xyiT82d2XmXFgAPT:bnt4x4l) (no related file):*
Looks good and works perfectly! Let's merge it!

---

PR #487, opened 2021-03-29
DDP-5708 rgp: styling for required questions

yufengwng:
I see that other projects use CSS for styling required questions. Doing the same for RGP so we remove the `*` from actual question prompt text in the backend configuration (see https://github.com/broadinstitute/ddp-study-server/pull/689).

### DIFF
--- a/ddp-workspace/projects/ddp-rgp/src/styles/_activity.scss
+++ b/ddp-workspace/projects/ddp-rgp/src/styles/_activity.scss
@@ -197,6 +197,10 @@
   }
 }
 
+.ddp-required-question-prompt:after {
+    content: ' *';
+}
+
 .ddp-helper {
   display: none;
 }
--- a/ddp-workspace/projects/ddp-rgp/src/styles/_activity.scss
+++ b/ddp-workspace/projects/ddp-rgp/src/styles/_activity.scss
@@ -202,7 +202,7 @@
 }
 
 .ddp-helper {
-  display: none;
+  display: none !important;
 }
 
 .ddp-textarea {


ddp-workspace/projects/ddp-rgp/src/styles/_activity.scss
```diff
@@ -197,6 +197,10 @@
   }
 }
 
+.ddp-required-question-prompt:after {
```
aleksey-kuznetsov, line 200: Just FYI, CSS Selector Level 3 spec now differentiates between pseudo elements using single-colon (`:hover`, `:focus`) and pseudo classes using double-colon (`::before`, `::after`, `::placeholder`).

aleksey-kuznetsov (commented):
Looks good!

aleksey-kuznetsov (dismissed):

ddp-workspace/projects/ddp-rgp/src/styles/_activity.scss
yufengwng, line 200: Saw a comment about using double-colon here. I didn't know about `::after`, but just following whatever convention the other studies are using.

charlottedye: I can't remember exactly what it's called (something like pseudo-element, I think) but this is just CSS's way of letting you sneak something in after an element.

ddp-workspace/projects/ddp-rgp/src/styles/_activity.scss
```diff
@@ -197,6 +197,10 @@
...
+    content: ' *';
+}
+
 .ddp-helper {
   display: none;
```
charlottedye: I think this must be an existing issue, but this gets overwritten in a few places.  If you wouldn't mind changing `display: none;` to `display none !important;`, that would be helpful.

charlottedye (dismissed):
This looks good, although I added a comment about an earlier bug that I missed if you think this is a good time to fix that.

ddp-workspace/projects/ddp-rgp/src/styles/_activity.scss
yufengwng: Done! (1b014bfd) I don't have context on where this CSS class comes into play, so I didn't test it.

charlottedye (approved):

PR #490, opened 2021-03-31
DDP-5746: Bug fix for Family History activity

alextanlan:
Fixed a bug when a name was not displayed on a card after leaving an edit dialog.


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activityTextAnswer.component.ts
```diff
@@ -77,9 +78,18 @@ export class ActivityTextAnswer implements OnChanges, OnInit {
...
+    public textInputModelChanged$ = new Subject<string>();
 
     public ngOnInit(): void {
         this.initFilteredSuggestions();
+
+        this.textInputModelChanged$.pipe(
+            debounceTime(300),
+            distinctUntilChanged()
+            )
+            .subscribe((value: string) => {
```
MocanaAtBroad: Might have to handle unsubscribe

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activityTextAnswer.component.ts
```diff
@@ -1,65 +1,66 @@
...
+                <mat-label *ngIf="block.label" [innerHTML]="block.label"></mat-label>
+                <input matInput [(ngModel)]="block.answer"
+                       [minlength]="block.minLength"
+                       [maxlength]="block.maxLength"
+                       [pattern]="block.regexPattern"
+                       [placeholder]="placeholder || block.placeholder"
+                       [disabled]="readonly"
+                       [attr.data-ddp-test]="'answer:' + block.stableId"
+                       (input)="inputSubject.next(textInput.value)"
+                       (ngModelChange)="onChange($event)"
```
MocanaAtBroad: It occurs to me that there might be a simpler solution: looks like root cause of the problem is that we originally capturing text using the "onchange" event and this event does not trigger on initial input.
So probably simpler to just tweak existing code to do same thing it is doing now, but go from:
```typescript
(change)="onChange(textInput.value)
```
to
```typescript
(input)="onChange(textInput.value)
```
(You could also consider renaming onChange method to match the name of the event that triggers call)

And since you still want to execute an `inputSubject.next(textInput.value)` , you can just put this line in the onChange method.

Approach also has virtue of avoiding adding more instance variables (every time we add an instance variable, particularly in components, my anxiety goes up. More state to worry about).

By the way: this is a suggestion. Please evaluate and use if you think makes sense to use.

MocanaAtBroad: Also, we should consider applying fix to the other textAnswer cases included in this component (like the `isEssayInputType()` case). Not seeing it now because we are not including essay questions in dialogs, but this is a bug waiting to happen.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activityTextAnswer.component.ts
alextanlan: fixed at 9be26f0d422939a3407d3a6e149be665068dd7cc

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activityTextAnswer.component.ts
alextanlan: Agree, the first implementation was overcomplicated. Thank you for the tip! 

Refactored at 9be26f0d422939a3407d3a6e149be665068dd7cc in more elegant way:
- kept `onChange` for email component as to emit changed value immediately
- used `(input) = "inputSubject.next(textInput.value)"` for both TextInput and EssayInput  in order to fire the changed textInput value. 
  Added a debounce there to decrease number of emitted values (patch answer requests).
  The `inputSubject` is also used for getting filteredSuggestions$ at the same time with the same `input event`    
  (because `Rxjs Subject` allows to share a single execution with multiple observers).

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activityTextAnswer.component.ts
```diff
@@ -146,4 +156,18 @@ export class ActivityTextAnswer implements OnChanges, OnInit {
...
         }
     }
+
+    ngOnDestroy(): void {
+        this.subscription.unsubscribe();
+    }
+
+    private initTextInputChangesEmitter(): void {
+        this.subscription = this.inputSubject.pipe(
+            debounceTime(300),
```
MocanaAtBroad, line 166: debounce help us here?

alextanlan: Yes, it helps to decrease number of patch requests. Because (input) event fires for every typed symbol.

withDebounce

withoutDebounce

MocanaAtBroad (approved):

PR #491, opened 2021-03-31
DDP-5747: Close button duplication

EdPokhvalin:
In fact, it isn't a duplication, it's a closing button from the composite answer. The problem was in the application-level styles, so I overrided them in the component level.

---
This change is [Reviewable](https://reviewable.io/reviews/broadinstitute/ddp-angular/491)


alextanlan (approved):
Looks good!

PR #493, opened 2021-04-01
DDP-5744: Fix saving child instances after wizard navigation

alextanlan:
Refactored that any changes of child instances lead to the whole (Family History) activity mutation.
It helped to fix the bug with losing changes (adding/removing) during the wizard navigation.

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityBlock/activityBlock.component.html
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityBlock/activityBlock.component.html
@@ -9,8 +9,7 @@
                                           [validationRequested]="validationRequested"
                                           [studyGuid]="studyGuid"
                                           [parentActivityInstanceGuid]="parentActivityInstanceGuid"
-                                          (deletedActivity)="onDeleteChildInstance($event)"
-                                          (editedActivity)="onEditChildInstance($event)">
+                                          (deletedActivity)="onDeleteChildInstance($event)">
                 </ddp-modal-activity-block>
             </ng-container>
             <ng-template #embedded>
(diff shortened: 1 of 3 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityBlock/activityBlock.component.html
```diff
@@ -9,7 +9,8 @@
                                           [validationRequested]="validationRequested"
                                           [studyGuid]="studyGuid"
                                           [parentActivityInstanceGuid]="parentActivityInstanceGuid"
-                                          (deleteActivity)="onDeleteChildInstance($event)">
+                                          (deletedActivity)="onDeleteChildInstance($event)"
+                                          (editedActivity)="onEditChildInstance($event)">
```
MocanaAtBroad: I wonder if we could eliminate a lot of the code added to handle "editedActivity" by instead of getting the new activityinstance from server and then pass it back via @Ouput emitters, we just modify the existing activity.
In other words:
1. we close dialog
2.we GET the activity
3. instead of emit(activityFromServer) we do something like:

```typescript
this.instance.activityName = activityFromServer.activityName;
this.instance.activityTitle = activityFromServer.activityTitle;
```
etc.

or better, use a function that copies all the properties from `activityFromServer` to `this.instance` .

Then (I hope) there would be no need of all the emitting and handling of events.

This make sense?

MocanaAtBroad: In other words, instead of trying to replace the the activityInstance with a new one from the server, we modify/mutate the one we have already.

alextanlan: Got your point - mutation on the most bottom level. 
Fixed at a7f67f78ae06ece7f69ba27a1729bc4bf77ddacb

MocanaAtBroad (approved):

PR #495, opened 2021-04-02
Update styling of expand contract button in Introduction section of Family History

alextanlan:


ddp-workspace/projects/ddp-osteo/src/styles/family-history.scss
```diff
@@ -39,21 +39,39 @@
...
+            display: none;
+        }
+
+        > summary::before {
+            content: '+';
+            font-size: 30px;
+            position: relative;
+            top: 4px;
+            left: -10px;
+        }
```
MocanaAtBroad, line 69: I really need to brush up on my CSS!

MocanaAtBroad (dismissed):
That was quick! Looks good.
I am approving, but wonder if it would be possible to introduce a little bit of animation so opening is not so sudden?
I think we can reuse this in other stuff moving forward, so this not just about this little piece.

ddp-workspace/projects/ddp-osteo/src/styles/family-history.scss
alextanlan: Added a bit of animation there at f341478afe184ea858b17eb10be0a2e893e91d55

MocanaAtBroad (approved):

PR #497, opened 2021-04-07
DDP-5797 Open an edit dialog for a new created childInstance immediately 

alextanlan:


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityBlock/activityBlock.component.html
```diff
@@ -9,6 +9,7 @@
                                           [validationRequested]="validationRequested"
                                           [studyGuid]="studyGuid"
                                           [parentActivityInstanceGuid]="parentActivityInstanceGuid"
```
MocanaAtBroad: Not one of your changes in this PR, but do we need parentActivityInstanceGuid Input param? I expect it would be part `instance`.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityBlock/activityBlock.component.html
```diff
@@ -9,6 +9,7 @@
...
+                                          [needsOpeningEditDialog]="createdInstanceGuidToOpen === instance.instanceGuid"
```
MocanaAtBroad: Consider simplifying/shortening name. Maybe `openDialog`?

MocanaAtBroad (commented):
Works well! A couple of small comments.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityBlock/activityBlock.component.html
alextanlan: I would agree. 
However there is already an existing method with the name `openDialog` (and `openEditDialog` as well) in `ModalActivityBlockComponent`.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityBlock/activityBlock.component.html
alextanlan: You are right. It is not even used there.
Removed at be6072935285637198cec61156d4196b7de1ea66.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityBlock/activityBlock.component.html
MocanaAtBroad: Indeed we do have these other method names!
IMO, I think the component @Input param name is more important than the names of those other methods.
Technically, all are "public", but only consumer is the component itself.
The input param is truly public/part of the component API , so I think more important to have nice /obvious names in the @Input (and @Output) params then in methods just used by same component.
So my advice: pick the nicest/cleanest/most concise name you can think of for @Input param and rename whatever else needs to be renamed  so you can use it.

aglushko-broad: Consider also directly  calling the openEditDialog method of ModalActivityBlockComponent using ViewChild to get the component instance in the activityBlock component class

alextanlan: @MocanaAtBroad, I have renamed **@Input** param at 7eac35c4c9ef792b0cc74580da75fb27d74e0982.

I am also wondering what you feel about @aglushko-broad suggested approach (see above ^) ?

MocanaAtBroad: I think that @aglushko-broad suggestion is a good idea!  The createdInstanceGuidToOpen is really just being used to pass an open dialog command. Getting a reference to the component and telling what to do is more direct.

alextanlan: Deal. Fixed at 5c2c87a6684eed7102fc5beba0fda241fe83760c

MocanaAtBroad (commented):
Alright!

MocanaAtBroad (approved):

PR #503, opened 2021-04-14
DDP-5815: fix fields alignment, font-size and label

aglushko-broad:
before:
image

after:
image

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/dropdownActivityPicklistQuestion.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/dropdownActivityPicklistQuestion.component.ts
@@ -9,11 +9,7 @@ import { NGXTranslateService } from '../../../services/internationalization/ngxT
 @Component({
     selector: 'ddp-activity-dropdown-picklist-question',
     template: `
-    <mat-form-field [floatLabel]="block.picklistLabel && block.selectMode === SELECT_MODE.SINGLE ? 'auto' : null">
-      <mat-label *ngIf="block.picklistLabel && block.selectMode === SELECT_MODE.SINGLE">
-        {{block.picklistLabel}}
-      </mat-label>
-
+    <mat-form-field>
       <ng-container *ngIf="block.selectMode === SELECT_MODE.MULTIPLE; then matSelect else nativeSelect">
       </ng-container>
 
@@ -39,6 +35,7 @@ import { NGXTranslateService } from '../../../services/internationalization/ngxT
                   [disabled]="readonly"
                   (change)="handleNativeSelect($event.target.value); details.show ? updateCharactersLeftIndicator(details.stableId) : null"
                   class="width">
+              <option value="">{{block.picklistLabel}}</option>
               <option *ngFor="let option of block.picklistOptions"
                       [value]="option.stableId">
                   {{option.optionLabel}}
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/dropdownActivityPicklistQuestion.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/dropdownActivityPicklistQuestion.component.ts
@@ -9,7 +9,7 @@ import { NGXTranslateService } from '../../../services/internationalization/ngxT
 @Component({
     selector: 'ddp-activity-dropdown-picklist-question',
     template: `
-    <mat-form-field [floatLabel]="block.picklistLabel && block.selectMode === SELECT_MODE.SINGLE ? 'always' : null">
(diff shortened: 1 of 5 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/dropdownActivityPicklistQuestion.component.ts
```diff
@@ -9,54 +9,60 @@ import { NGXTranslateService } from '../../../services/internationalization/ngxT
...
+                     [placeholder]="details.placeholder"
+                     [value]="details.text">
+          </mat-form-field>
+          <p *ngIf="!readonly" class="ddp-helper">
+              <span class="ddp-counter-color">
+                  {{ questionIdToCharactersLeft[details.stableId] }}
+              </span>{{ questionIdToCharacterLeftMsg[details.stableId] }}
+          </p>
+      </ng-container>
+    </mat-form-field>`,
```
MocanaAtBroad, line 62: @aglushko-broad : unfortunately git diff is not really helping to narrow down the change you made here. Could you briefly highlight/explain?

alextanlan:
You improved the alignment.
However, `ddp-activity-answer`-s are still not on the same level.

For testing it  - I moved left block to the right as to overlap them (by adding `position: relative; left: 50px`)

- Before changes (on dev env):

before

- After changes (run your branch locally):

after

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/dropdownActivityPicklistQuestion.component.ts
aglushko-broad: sure! the changes I made:
- wrap with mat-form-field and assign matNativeControl to native select in order to apply material styles for alignment
- remove the first option from native select which was equal to the label and add mat-label (I had to add it out of native select ng-template in order to let internal mechanism which uses ng-content to handle it)

aglushko-broad:
> You improved the alignment.
> However, `ddp-activity-answer`-s are still not on the same level.
> 
> For testing it - I moved left block to the right as to overlap them (by adding `position: relative; left: 50px`)
> 
> * Before changes (on dev env):
> 
> before
> 
> * After changes (run your branch locally):
> 
> after

nice catch! didn't see it on the small screen. Fixed

alextanlan:
Perfect!

alextanlan (dismissed):

MocanaAtBroad:
OK. One super minor thing: the bottom of the "g" in the words "age" and "diagnosis" is getting cut off (it shows up in your screen capture too):
Screen Shot 2021-04-15 at 12 42 57 PM

MocanaAtBroad:
Another thing I am noticing is that after introducing the Material field label, this particular question looks unbalanced (Age of diagnosis has Material label, but Cancer Type just placeholder). Looking if this something we can tweak from server configuration....

MocanaAtBroad:
Let's back off/undo the change for Material label for the pull-down.

I think it might have wider repercussions/used in another apps and each of them would have to be verified that it looks good. Even in this case, just looking at code and configuration it is not totally clear what is the right thing to do now with the cancer type field while at the same time ensuring that it continuous to be right in all our apps.

Let's just make sure the alignment/original bug reported is right. We need that. At some point we can come back and revisit the Material label problem.

aglushko-broad:
> Let's back off/undo the change for Material label for the pull-down.
> 
> I think it might have wider repercussions/used in another apps and each of them would have to be verified that it looks good. Even in this case, just looking at code and configuration it is not totally clear what is the right thing to do now with the cancer type field while at the same time ensuring that it continuous to be right in all our apps.
> 
> Let's just make sure the alignment/original bug reported is right. We need that. At some point we can come back and revisit the Material label problem.

thanks for noticing that label position for select doesn't match the position for autocomplete input! I fixed it quickly, you can take a look again. Though I see your concern that it affects all apps I'd say having the label is better than don't have a label at all, it's more noticeable in the selected state, then the user can't be sure what exactly s/he selected before.
This is how it looks now
image
But anyway it's not a problem to remove the label for now and return it how it was before. Could you please confirm one more time we want to do it, thanks!

aglushko-broad:
> Let's back off/undo the change for Material label for the pull-down.
> 
> I think it might have wider repercussions/used in another apps and each of them would have to be verified that it looks good. Even in this case, just looking at code and configuration it is not totally clear what is the right thing to do now with the cancer type field while at the same time ensuring that it continuous to be right in all our apps.
> 
> Let's just make sure the alignment/original bug reported is right. We need that. At some point we can come back and revisit the Material label problem.

@MocanaAtBroad  it's all done! I updated the screenshot in the PR description

MocanaAtBroad (commented):
Looks good! Let's merge.

alextanlan (approved):

PR #505, opened 2021-04-15
DDP-5597: Text question confirmation support

aleksey-kuznetsov:
New version for text question confirmation support which relies on new custom `ActivityMatchValidationRule`.


ddp-workspace/projects/ddp-sdk/src/lib/services/activity/activityValidatorBuilder.service.ts
```diff
@@ -62,6 +64,9 @@ export class ActivityValidatorBuilder {
         if (questionBlock.questionType === QuestionType.Date) {
             rules.push(new ActivityDateNavyValidationRule(questionBlock, this.dateService));
         }
+        if (questionBlock instanceof ActivityTextQuestionBlock && questionBlock.confirmEntry) {
```
yufengwng: Text questions with the `EMAIL` input type can also have `confirmEntry`, but it looks like `ActivityMatchValidationRule` is only meant for `TEXT` input type? I think this is causing some unwanted error messages. For example, the Brain study has a `EMAIL` question with confirmation in the Child Contact activity and I see this:

Screen Shot 2021-04-15 at 11 45 49 AM

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-text-input/activityTextInput.component.html
```diff
@@ -0,0 +1,61 @@
...
+        ></span>
+      </mat-option>
+    </mat-autocomplete>
+  </mat-form-field>
+
+  <ng-container *ngIf="block.confirmEntry">
+    <p
+      class="ddp-question-prompt ddp-question-prompt-confirmation"
+      [class.ddp-required-question-prompt]="this.block.isRequired"
+      [innerHTML]="block.confirmPrompt"
```
yufengwng, line 49: Is there a way to make this be the "label" instead of a paragraph? It works fine for a study like Brain (see my other screenshot), but for RGP, the UI is a bit different. Do we need some other flag to tell frontend to make this a label? What do you think, @charlottedye?

Screen Shot 2021-04-15 at 11 49 36 AM

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/activityValidatorBuilder.service.ts
aleksey-kuznetsov: Thanks for that! I'll only apply this validation rule only for `Text` input type.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-text-input/activityTextInput.component.html
aleksey-kuznetsov: I was just following how other questions are set up, as for RGP we could just tweak the configuration for this question (i.e. drop `confirmPrompt` and only have `prompt` for main question).

aleksey-kuznetsov:
@charlottedye please take a look :)

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/activityValidatorBuilder.service.ts
aleksey-kuznetsov: Done. It shouldn't be a problem for questions with `Email` input type from now on.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-text-input/activityTextInput.component.html
charlottedye: I think ideally the error message related to the first phone field should display under that field.

yufengwng: @charlottedye The main issue I want to highlight here is not the error messages but the prompt/label. Here's what I'm seeing for the backend configuration for this question (pseudo-code):
```
  filler_phone: {
    placeholderTemplate: "your telephone number * ",
    confirmEntry: true,
    confirmPromptTemplate: "confirm your telephone number",
    mismatchMessageTemplate: "please check your phone",
    validations: [
      {
        rule: REGEX,
        hintTemplate: "please enter valid phone"
      {
        rule: REQUIRED,
        hintTemplate: "please enter valid phone"
      }
    ]
  },
```
So to start with, we don't have a `prompt` for this question. We have a `placeholder` so that we get the UI experience of having the placeholder transition to a label. So @aleksey-kuznetsov , if we also drop `confirmPrompt`, there's no `prompt` to use. So my concern is: how do we get that second text input to use  `confirm your telephone number` as a placeholder AND also have that transition to a label when they start typing.

yufengwng: As a reminder, here's what we have in Gen2:

Screen Shot 2021-04-16 at 12 35 40 PM

charlottedye:
That's what you see before you click the next button, which is when the
error messages appear in Gen2.  After that, you should see an error message
under "Confirm Your Telephone Number", I believe, but not 100% sure

yufengwng:
> That's what you see before you click the next button, which is when the
error messages appear in Gen2.  

Yup. However, again, I'm not talking about the error messages. I am focusing on the placement of the question prompt/placeholder/label text.

charlottedye:
> > That's what you see before you click the next button, which is when the
> > error messages appear in Gen2.
> 
> Yup. However, again, I'm not talking about the error messages. I am focusing on the placement of the question prompt/placeholder/label text.

Got it!

charlottedye:
I ran the code and I thought I did everything correctly but now I'm not seeing a confirm field at all for some reason?

yufengwng:
@charlottedye Did you enable the `confirmEntry` configuration in the backend configuration files (see `enrollment-questions.conf`)? For branches, you can use `develop` branch for the backend, and this PR's branch for frontend.

charlottedye:
Apparently I forgot to pull in the latest develop in the backend! Although I still don't see confirmEntry being set to true: we should probably make that change in the backend.  I'm going ahead and making the change locally and testing it that way for now, though.

charlottedye (changes requested):
Screen Shot 2021-04-16 at 1 07 39 PM


Screen Shot 2021-04-16 at 1 06 51 PM

I think these two things need to be fixed.  First, the error message for the first phone number field should be displayed under the first phone number field.  Second, the formatting here is kind of ugly: the prompt format for the phone number field and the confirm phone number field should be the same.  I think this is what @yufengwng was mentioning earlier.

yufengwng:
Thanks @charlottedye. What I'm thinking is that, the frontend doesn't really have anything to tell it to render this in a different way. How exactly does frontend know to put `confirm your telephone number` as a placeholder instead of a prompt? There's nothing in the API that informs it to do this.

I'm starting to think we might need to take the same approach as what we did for the question prompt: instead of using `confirmPrompt`, we add a `confirmPlaceholder` to the API. If `confirmPrompt` is empty, we don't need to display a prompt for the second text input, and if `confirmPlaceholder` is available, we use that for the second text input. What do you think, @charlottedye and @aleksey-kuznetsov and @MocanaAtBroad ? Is there any other way for frontend to do this?

charlottedye:
> Thanks @charlottedye. What I'm thinking is that, the frontend doesn't really have anything to tell it to render this in a different way. How exactly does frontend know to put `confirm your telephone number` as a placeholder instead of a prompt? There's nothing in the API that informs it to do this.
> 
> I'm starting to think we might need to take the same approach as what we did for the question prompt: instead of using `confirmPrompt`, we add a `confirmPlaceholder` to the API. If `confirmPrompt` is empty, we don't need to display a prompt for the second text input, and if `confirmPlaceholder` is available, we use that for the second text input. What do you think, @charlottedye and @aleksey-kuznetsov and @MocanaAtBroad ? Is there any other way for frontend to do this?

That sounds like a good idea.  The other option, depending on time constraints, would be to just make both of them prompts.  But that would still be ugly because it wouldn't match the other questions.  So I think this is probably the way to go.

aleksey-kuznetsov:
@yufengwng are you talking about `<mat-label>` thing? If so, then we should add `confirmLabel` so that this confirm question works the same way as main question.

As for the error per field, I'm not really sure if this easily achievable.

aleksey-kuznetsov:
> Screen Shot 2021-04-16 at 1 07 39 PM
> 
> Screen Shot 2021-04-16 at 1 06 51 PM
> 
> I think these two things need to be fixed. First, the error message for the first phone number field should be displayed under the first phone number field. Second, the formatting here is kind of ugly: the prompt format for the phone number field and the confirm phone number field should be the same. I think this is what @yufengwng was mentioning earlier.

I wonder why you've got 2 identical questions and why second one is pushed to the right...

yufengwng:
Alright folks, I've created a new ticket to capture the missing requirement here: https://broadinstitute.atlassian.net/browse/DDP-5973. @charlottedye @aleksey-kuznetsov can you double-check the ticket to see if it'll give us what we need?

---

> I wonder why you've got 2 identical questions and why second one is pushed to the right...

@aleksey-kuznetsov are you talking about those 2 `*` (asterisks)? The backend text for it was misconfigured and have an extra `*` in the text. It'll be taken care of in the backend. You don't have to worry about it in this PR.

aleksey-kuznetsov:
@yufengwng sorry, It was a bit misleading for me, I didn't notice that Charlotte shared 2 screenshots instead of 1 😅

Anyway, a placeholder field is responsible for just a regular placeholder. If we need functionality around label (typical Material input labels which look like a placeholder but transition to label state when user focuses an input) then we need to also introduce `confirmLabel` just like for regular question.

yufengwng:
@aleksey-kuznetsov My understanding is that currently in Angular SDK, if the text question `prompt` is empty and `placeholderText` is provided, then Angular SDK will instead use the placeholder as the "label" (e.g. what you mentioned about Material input label being both a placeholder and a label). In fact, that is how we achieved the first text input in the screenshots for `your telephone number`.

My proposal for `confirmPlaceholderText` is similar. If `confirmPrompt` is empty and `confirmPlaceholderText` is provided, then Angular SDK should do the same thing it's doing for regular question: use `confirmPlaceholderText` as Material input label/placeholder.

aleksey-kuznetsov:
You're right, I was thinking that `<mat-label>` thing is responsible for that but setting a placeholder will work just fine.

I actually wonder why do we have this `<mat-label *ngIf="block.label"></mat-label>` chunk of code although I couldn't find any references where we would set `block.label = ...` except for Composite questions.

yufengwng:
@aleksey-kuznetsov We have implemented the change on the backend, and I have redeployed new study configuration to `dev` environment. We now have a `confirmPlaceholderText`. This is what you'll see from the API in `dev`:

Screen Shot 2021-04-27 at 2 40 16 PM

aleksey-kuznetsov:
@yufengwng thank you so much! I'll make necessary changes today.

aleksey-kuznetsov:
@yufengwng @charlottedye I've added support for `confirmPlaceholderText` field so now we should only have placeholders in RGP.

As for error messages under each field. I'm not really sure how we can achieve this, any thoughts?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-text-input/activityTextInput.component.html
```diff
@@ -0,0 +1,62 @@
...
+    ></p>
+
+    <mat-form-field class="example-input-field-confirmation">
+      <input
+        matInput
+        type="text"
+        [formControl]="formGroup.get(confirmationControlName)"
+        [placeholder]="block.confirmPlaceholder"
```
yufengwng: If `confirmPlaceholder` is not provided or empty, should we fallback and use the other `placeholder` (if available)?

yufengwng (dismissed):
Re: error message placement. I feel that since currently we're displaying the fields top-to-bottom instead of side-by-side, having the error message only show under the confirm input seems fine to me. But I'm not sure if error message placement is a hard requirement. I think Charlotte or Jeff will have a better take on this. Otherwise, this looks good to me.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-text-input/activityTextInput.component.html
aleksey-kuznetsov: This might be an option. I'll add that.

aleksey-kuznetsov:
I've also added two classes to wrapping `form` element: one which is static and other one which depends on question `stableId`. This way in RGP we can align those two fields in one line and have error message to appear underneath both of them.

yufengwng (approved):
Looks good as far as I can tell!

charlottedye (approved):

PR #528, opened 2021-04-27
DDP-5932:  Fix error with getting default/temp address

alextanlan:
- Fixed errors connected with getting either a default or a temp address for Medical release activity
- Changed console.error hadling (and reporting to Stackdriver) in order to have more meaningful messages about the errors

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
@@ -268,7 +268,7 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
             tap(() => this.stateUpdates$.next({isTemporarilyDisabled: true})),
             mergeMap((state) => this.addressService.findDefaultAddress().pipe(
                 catchError(error => {
-                    this.logger.logWarning(this.LOG_SOURCE, `An error occurred during findDefaultAddress: ${error.code}, ${error.message}`);
+                    this.logger.logDebug(this.LOG_SOURCE, `An error occurred during findDefaultAddress: ${error.code}, ${error.message}`);
                     return of(null);
                 }),
                 map(defaultAddress => [state, defaultAddress]),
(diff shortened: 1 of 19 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -220,442 +223,447 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
...
+            distinctUntilChanged()
+        );
+
+        const initializeStateAction$ = this.state$.pipe(
+            take(1),
+            tap(() => busyCounter$.next(1)),
+            // let's disable form while we try to load initial address
+            tap(() => this.stateUpdates$.next({isTemporarilyDisabled: true})),
+            mergeMap((state) => this.addressService.findDefaultAddress().pipe(
+                catchError(error => {
```
alextanlan, line 270: added catchError.

Other changes in the component are only for fixing the indent.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -220,442 +223,447 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
...
+                map(defaultAddress => [state, defaultAddress]),
+            )),
+            tap(([_, defaultAddress]: [ComponentState, Address | null]) =>
+                defaultAddress && this.inputAddress$.next(defaultAddress)),
+            take(1),
+            // filter for case where we need to go on to look for a temp address?
+            filter(([state, defaultAddress]) => !defaultAddress && !!(state as ComponentState).activityInstanceGuid),
+            // map(([state, _]) => state as ComponentState),
+            mergeMap(([state, _]) => this.addressService.getTempAddress(state.activityInstanceGuid).pipe(
+                catchError(error => {
```
alextanlan, line 283: added catchError.

Other changes in the component are only for fixing the indent.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -220,442 +223,447 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
...
+                    this.logger.logWarning(this.LOG_SOURCE, `An error occurred during getTempAddress: ${error.code}, ${error.message}`);
```
alextanlan: A question:
should we report about these errors (no default/temp address) to GCP  
or it is just enough to write a warning to console ?

aglushko-broad (dismissed):

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
MocanaAtBroad: In these 2 cases here, I think the right think to do is to capture (like you are doing), and then log as as debug (logDebug , I believe)?
I think our base code that does the http calls might be a little too creative regarding capturing certain errors, and a bunch of other stuff, but regardless the two "not founds" are not really exceptions; client just checks if an address exists and the answer, sometimes is, it does not (404).

aglushko-broad (approved):

PR #530, opened 2021-04-29
DDP-5770: Bug fix

alextanlan:
Use correct field (`activityName`) for the nested embedded activity label as it implemented for modal activity block as well


yufengwng (approved):
Thanks for the fix!

PR #533, opened 2021-04-30
DDP-5925 cmi: bootstrap angio prism login

yufengwng:
This is the minimum to setup the prism routes for Angio and get admin logins working. See the necessary Auth0 changes here: https://github.com/broadinstitute/ddp-study-server/pull/735.

### DIFF
--- a/ddp-workspace/projects/toolkit/src/lib/components/session-expired/session-expired-redesigned.component.ts
+++ b/ddp-workspace/projects/toolkit/src/lib/components/session-expired/session-expired-redesigned.component.ts
@@ -24,25 +24,15 @@ import { ActivatedRoute } from '@angular/router';
         </main>`
 })
 export class SessionExpiredRedesignedComponent extends SessionExpiredComponent implements OnInit {
-    private isAdminSessionExpired: boolean;
 
     constructor(
-        private activatedRoute: ActivatedRoute,
+        private _activatedRoute: ActivatedRoute,
         private headerConfig: HeaderConfigurationService,
         private _auth0: Auth0AdapterService) {
-        super(_auth0);
+        super(_activatedRoute, _auth0);
     }
 
     public ngOnInit(): void {
         this.headerConfig.setupDefaultHeader();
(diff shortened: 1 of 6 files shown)


ddp-workspace/projects/toolkit/src/lib/components/session-expired/session-expired-redesigned.component.ts
```diff
@@ -24,25 +24,15 @@ import { ActivatedRoute } from '@angular/router';
         </main>`
 })
 export class SessionExpiredRedesignedComponent extends SessionExpiredComponent implements OnInit {
```
yufengwng, line 26: `SessionExpiredRedesignedComponent` is used by TestBoston study, but Angio uses the parent `SessionExpiredComponent`, so I moved things into there.

MocanaAtBroad:
Changes look reasonable, but will await @aglushko-broad. I believe she has been doing some work on some of the same code, so she in better place to evaluate impact.

ddp-workspace/projects/toolkit/src/lib/components/session-expired/session-expired-redesigned.component.ts
aglushko-broad: wondering why angio can't use SessionExpiredRedesignedComponent as well. I can see that not only testboston, but atcp, brain and osteo use it as well. What's the diff between those 4 projects and angio?

yufengwng: I'm not sure either! (I haven't followed development on frontend too closely.) I think it is because CMI studies (angio, brain, etc) uses different visuals / styles, while studies like TestBoston wants to display things differently (hence the component name "redesigned"). I think it's best to keep whatever design Angio is currently using.

yufengwng:
@aglushko-broad @MocanaAtBroad Any further comments?

aglushko-broad (approved):

aglushko-broad:
LGTM. Thanks for the help!

PR #538, opened 2021-05-04
Update super constructor call in Prion footer

MocanaAtBroad:
I had updated Footer component in toolkit and ended up breaking Prion footer that subclasses.
This is the fix.

### DIFF
--- a/ddp-workspace/projects/toolkit-prion/src/lib/components/footer/prionFooter.component.ts
+++ b/ddp-workspace/projects/toolkit-prion/src/lib/components/footer/prionFooter.component.ts
@@ -2,7 +2,7 @@ import { Component, Inject } from '@angular/core';
 import { MatDialog } from '@angular/material/dialog';
 import {
   AnalyticsEventsService,
-  Auth0AdapterService,
+  Auth0AdapterService, NGXTranslateService,
   SessionMementoService,
   WindowRef
 } from 'ddp-sdk';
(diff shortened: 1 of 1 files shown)


ddp-workspace/projects/toolkit-prion/src/lib/components/footer/prionFooter.component.ts
```diff
@@ -44,8 +44,9 @@ export class PrionFooterComponent extends FooterComponent {
     private _windowRef: WindowRef,
     @Inject('toolkit.toolkitConfig') private _toolkitConfiguration: ToolkitConfigurationService,
     private _auth0Adapter: Auth0AdapterService,
-    private _session: SessionMementoService) {
-    super(_dialog, _analytics, _windowRef, _toolkitConfiguration);
+    private _session: SessionMementoService,
+    private translate: NGXTranslateService) {
```
yufengwng, line 48: Is prefixing with an underscore the convention for things that's not used in current class but need to be injected to be passed to parent class?

yufengwng (approved):

ddp-workspace/projects/toolkit-prion/src/lib/components/footer/prionFooter.component.ts
MocanaAtBroad: For what I have seen, Charlotte has adopted that convention for private instance variables.

PR #550, opened 2021-05-12
Ddp 5923 prism search

aglushko-broad:
image
image
image
image
image

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/prism/prism.component.spec.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/prism/prism.component.spec.ts
@@ -1,0 +1,379 @@
+import { Component, DebugElement } from '@angular/core';
+import { ComponentFixture, fakeAsync, TestBed, tick } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+import { NoopAnimationsModule } from '@angular/platform-browser/animations';
+import { MatStepperModule } from '@angular/material/stepper';
+import { TranslateLoader, TranslateModule, TranslateService } from '@ngx-translate/core';
+import { NEVER, Observable, of } from 'rxjs';
+import { MatProgressSpinnerModule } from '@angular/material/progress-spinner';
+import { MatInputModule } from '@angular/material/input';
+import { ReactiveFormsModule } from '@angular/forms';
+import { MatTableModule } from '@angular/material/table';
+import { PrismComponent } from './prism.component';
+import { SessionMementoService } from 'ddp-sdk';
+import { ParticipantsSearchServiceAgent } from '../../services/serviceAgents/participantsSearchServiceAgent.service';
+import { RouterTestingModule } from '@angular/router/testing';
+import { MatPaginatorModule } from '@angular/material/paginator';
+import { EnrollmentStatusType } from '../../models/enrollmentStatusType';
+import { Router } from '@angular/router';
+
+class TranslateLoaderMock implements TranslateLoader {
+  getTranslation(code: string = ''): Observable<object> {
+    const TRANSLATIONS = {
+      en: {
+        SDK: {
+          Prism: {
+            NoData: 'NoData test',
+            NoUsers: 'NoUsers test',
+            WarningSize: 'WarningSize {{count}} from {{totalCount}} test',
+            MinQueryLengthHint: 'MinQueryLengthHint {{length}} test',
+            ClearSearch: 'ClearSearch test',
+            DashboardButtonLabel: 'DashboardButtonLabel {{user}} test'
+          }
+        }
+      }
+    };
+    return of(TRANSLATIONS[code]);
+  }
+}
+// tslint:disable-next-line:max-classes-per-file
+@Component({
+  template: `
+        <ddp-prism [dashboardRoute]="dashboardRoute"
+                   [withProxy]="withProxy"
+                   [withInvitationId]="withInvitationId"
+                   [withLegacyId]="withLegacyId"></ddp-prism>
+    `})
+class WrapperComponent {
+  dashboardRoute: string;
+  withProxy: boolean;
+  withInvitationId: boolean;
+  withLegacyId: boolean;
+}
+
+describe('PrismComponent', () => {
+  let fixture: ComponentFixture<WrapperComponent>;
+  let wrapperComponent: WrapperComponent;
+  let component: PrismComponent;
+  let debugElement: DebugElement;
+  let sessionServiceSpy: jasmine.SpyObj<SessionMementoService>;
+  let participantsSearchSpy: jasmine.SpyObj<ParticipantsSearchServiceAgent>;
+  let router: Router;
+
+  const dashboardRoute = 'test-dashboard-route';
+  const data = [{
+      guid: '1234',
+      hruid: '5678',
+      firstName: 'Bob',
+      lastName: 'Adams',
+      status: EnrollmentStatusType.COMPLETED,
+      invitationId: 'TBM398WQ8P6Z',
+      legacyShortId: '12',
+      proxy: {
+        guid: '1235',
+        hruid: '5679',
+        firstName: 'Tom',
+        lastName: 'Adams',
+        email: 'proxytest@test.com',
+      }
+    },
+      {
+        guid: '1234',
+        hruid: '5678',
+        status: EnrollmentStatusType.REGISTERED,
+      },
+      {
+        guid: '1234',
+        hruid: '5678',
+        firstName: 'Bob',
+        lastName: 'Adams',
+        status: EnrollmentStatusType.ENROLLED,
+        invitationId: 'TBM398WQ8P6Z',
+        email: 'test@test.com',
+      }
+    ];
+
+  beforeEach(async() => {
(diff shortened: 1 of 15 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/prism/prism.component.ts
```diff
@@ -0,0 +1,129 @@
...
+    };
+  }
+
+  public ngOnDestroy(): void {
+    this.ngUnsubscribe.next();
+    this.ngUnsubscribe.complete();
+  }
+
+  public clearSearch(): void {
+    this.searchField.reset('', {emitEvent: true});
```
alextanlan: #minor 
not necessary options` {emitEvent: true}`.
Events emit by default there.

ddp-workspace/projects/ddp-sdk/src/lib/components/prism/prism.component.ts
```diff
@@ -0,0 +1,129 @@
...
+      distinctUntilChanged(),
+      tap(() => { this.isSearchDebounce = true; }),
+      debounceTime(300),
+      tap(() => {
+        this.isSearchLoading = true;
+        this.isSearchDebounce = false;
+      }),
+      switchMap((query) => {
+        if (query.length >= this.MIN_SEARCH_LENGTH) {
+          return this.participantsSearch.search(query);
```
alextanlan, line 91: What if the search returns an error ? 
I suggest (at least) adding logDebug info for errors.

aglushko-broad: the error is handled in serviceAgent service from which ParticipantsSearchServiceAgent inherits. logError is already there

ddp-workspace/projects/ddp-sdk/src/lib/components/prism/prism.component.ts
aglushko-broad: thanks! done!

alextanlan (dismissed):

ddp-workspace/projects/ddp-sdk/src/lib/models/searchParticipant.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export interface SearchParticipant extends SearchParticipantBase {
+    status: EnrollmentStatusType;
+    invitationId?: string;
+    proxy?: SearchParticipantBase;
+}
+
+export interface SearchParticipantResponse {
+    totalCount: number;
+    results: SearchParticipant[];
```
MocanaAtBroad: Let's move it out to own file. I think having the other 2 is fine as one is (sort of) private, but Reponse is a whole other thing (we have tried to standardize on one file per type definition, making exceptions for local/private)

ddp-workspace/projects/ddp-sdk/src/lib/models/enrollmentStatusType.ts
```diff
@@ -0,0 +1,16 @@
...
+    COMPLETED = 'COMPLETED',
+}
+export const enrollmentStatusTypeToLabel = new Map<EnrollmentStatusType, string>([
+    [EnrollmentStatusType.REGISTERED, 'Registered'],
+    [EnrollmentStatusType.ENROLLED, 'Enrolled'],
+    [EnrollmentStatusType.EXITED_BEFORE_ENROLLMENT, 'Exited before enrollment'],
+    [EnrollmentStatusType.EXITED_AFTER_ENROLLMENT, 'Exited after enrollment'],
+    [EnrollmentStatusType.CONSENT_SUSPENDED, 'Consent suspended'],
+    [EnrollmentStatusType.COMPLETED, 'Completed'],
+]);
```
MocanaAtBroad, line 16: Maybe we internationalize? (Was not a requirement, but you are most of the way there!)

ddp-workspace/projects/ddp-sdk/src/lib/components/prism/prism.component.ts
```diff
@@ -0,0 +1,129 @@
...
+@Component({
+  selector: 'ddp-prism',
+  templateUrl: './prism.component.html',
+  styleUrls: ['./prism.component.scss'],
+})
+export class PrismComponent implements OnInit, OnDestroy, AfterViewInit, OnChanges {
+  @Input() dashboardRoute: string;
+  @Input() withProxy = false;
+  @Input() withInvitationId = false;
+  @Input() withLegacyId = false;
```
MocanaAtBroad: Instead of @Input() params, let's try to drive the columns that are hidden/shown from sdk configuration service. Idea is that we can just set some configuration parameters and not have to write additional code to pick columns.

ddp-workspace/projects/ddp-sdk/src/lib/components/prism/prism.component.spec.ts
```diff
@@ -0,0 +1,379 @@
...
+    participantsSearchSpy.search.and
+        .returnValue(of({results: [{...data[0], firstName: null, lastName: null, guid}], totalCount: 1}));
+    component.searchField.patchValue('test');
+    tick(300);
+    fixture.detectChanges();
+
+    const firstUserDashboardButton = fixture.debugElement.query(By.css('.dashboard-column button')).nativeElement;
+    expect(firstUserDashboardButton.getAttribute('aria-label')).toBe(`DashboardButtonLabel ${guid} test`);
+  }));
+});
```
MocanaAtBroad, line 336: This is a very very nice test!

MocanaAtBroad (commented):
This all looks and works very nicely! A few small comments, one change request, and a question.

MocanaAtBroad:
A question, not a feature request, so don't jump to implementation. How hard to add horizontal arrow to indicate that there are more columns out of view? (We do have that horizonal scrollbar so this is a bad example, as sometimes the scrollbar is not visible, e.g., middle of a long list of results).
prismWithArrows copy

aglushko-broad:
> A question, not a feature request, so don't jump to implementation. How hard to add horizontal arrow to indicate that there are more columns out of view? (We do have that horizonal scrollbar so this is a bad example, as sometimes the scrollbar is not visible, e.g., middle of a long list of results).
> prismWithArrows copy

what do you think about a sticky horizontal scrollbar? it's visible all the time when user can see the table on the screen. I implemented such a table directive on the previous project and it shouldn't take a long time to do the same here. If you have concerns about that we can return to the arrows :)

MocanaAtBroad:
> > A question, not a feature request, so don't jump to implementation. How hard to add horizontal arrow to indicate that there are more columns out of view? (We do have that horizonal scrollbar so this is a bad example, as sometimes the scrollbar is not visible, e.g., middle of a long list of results).
That sounds interesting! What would it look like?

aglushko-broad:
> > > A question, not a feature request, so don't jump to implementation. How hard to add horizontal arrow to indicate that there are more columns out of view? (We do have that horizonal scrollbar so this is a bad example, as sometimes the scrollbar is not visible, e.g., middle of a long list of results).
> > > That sounds interesting! What would it look like?

you can find an example here http://jsfiddle.net/cowboy/45rEs/show/

ddp-workspace/projects/ddp-sdk/src/lib/components/prism/prism.component.ts
aglushko-broad: done!

ddp-workspace/projects/ddp-sdk/src/lib/models/searchParticipant.ts
aglushko-broad: done

ddp-workspace/projects/ddp-sdk/src/lib/models/enrollmentStatusType.ts
aglushko-broad: done

MocanaAtBroad:
Looks and works really well!
I do have one final request: Can we duplicate the count and current page indicator at the top of table, i.e., this 1-25 of 58 indicator, and place it above the table of results.
Screen Shot 2021-05-15 at 10 46 09 AM

(The only essential elements to duplicate are the count and and range of displayed values, but if easy enough to add the "Next/Previous" page control and the Items per page, don't see any harm in duplicating those as well.

aglushko-broad:
> Looks and works really well!
> I do have one final request: Can we duplicate the count and current page indicator at the top of table, i.e., this 1-25 of 58 indicator, and place it above the table of results.
> Screen Shot 2021-05-15 at 10 46 09 AM
> 
> (The only essential elements to duplicate are the count and and range of displayed values, but if easy enough to add the "Next/Previous" page control and the Items per page, don't see any harm in duplicating those as well.

having multiple paginators for one table might be slightly tricky (but doable)... what do you think if we just move the paginator to the top of the table? Or do you think allowing the user to switch pages at the bottom and at the top is essential? In case of just moving - I can do it in this PR, but if we want duplication I'd prefer to do it in separate PR since it requires some effort

MocanaAtBroad:
On the position of the paginator (you mean the arrows to go to next page right?), if I had a choice, I think at the bottom where you have it is the best spot (don't want users to have to scroll to top to go on to next page). But, obviously, nicer if can be done from top or bottom.

MocanaAtBroad (approved):

PR #555, opened 2021-05-13
DDP-6115 Bug fix in address component

alextanlan:
Hide `State` field in address form for countries which have `stateCode !=null`
i.g. Puerto Rico

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressInput.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressInput.component.ts
@@ -79,7 +79,11 @@ import { LoggingService } from '../../services/logging.service';
         </mat-form-field>
 
         <ng-container *ngIf="ais.countryInfo$ | async as info; else defaultStateField">
-          <ng-container *ngIf="hasStatesList(info); then showStateDropdown else showStateTextField">
+          <ng-container *ngIf="!info.stateCode">
+            <ng-container *ngIf="info.subnationalDivisionTypeName as stateLabel; else defaultStateField">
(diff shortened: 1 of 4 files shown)


MocanaAtBroad:
Alex: your change fixed the main problem with the form and the fields.

I pushed a commit that fixes the other part of the problem. 
Without this change, the form was displaying correctly, but we were still sending to server addresses that had "PR" for country **AND** state code, so the address verify/suggest would still insist there was something wrong with the address.

Try it out, and if it looks good, let's merge it.

alextanlan:
Looks good.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressInput.component.ts
```diff
@@ -79,7 +79,11 @@ import { LoggingService } from '../../services/logging.service';
         </mat-form-field>
 
         <ng-container *ngIf="ais.countryInfo$ | async as info; else defaultStateField">
-          <ng-container *ngIf="hasStatesList(info); then showStateDropdown else showStateTextField">
+          <ng-container *ngIf="!info.stateCode">
+            <ng-container *ngIf="info.subnationalDivisionTypeName as stateLabel; else defaultStateField">
```
aglushko-broad, line 83: #minor we probably don't need `as stateLabel`

aglushko-broad (approved):

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressInput.component.ts
alextanlan: I would keep it as is.
Because it helps us understand that we use `subnationalDivisionTypeName` field from data model as to display `stateLabel` in the form.

PR #558, opened 2021-05-17
DDP-6103- Fix excess page blinks when moving between Pepper activities

alextanlan:
The fixes suggest displaying `<ddp-activity>` component when we get `activityInstanceGuid` only
   as to avoid reload/partially load of data for the component.

Before fixes we had partially loaded current/previous activity that led to page blick (see screenshots below):
- after `count-me in` registration
issue1

- moving from "**About-you**" to "**Consent**"
issue2

### DIFF
--- a/ddp-workspace/projects/toolkit/src/lib/components/activity/activity.component.ts
+++ b/ddp-workspace/projects/toolkit/src/lib/components/activity/activity.component.ts
@@ -10,7 +10,8 @@ import { ActivityResponse } from 'ddp-sdk';
         <toolkit-header [showButtons]="false"
                         [stickySubtitle]="stickySubtitle">
         </toolkit-header>
-        <ddp-activity [studyGuid]="studyGuid"
+        <ddp-activity *ngIf="id"
+                      [studyGuid]="studyGuid"
                       [activityGuid]="id"
                       (submit)="navigate($event)"
                       (stickySubtitle)="showStickySubtitle($event)">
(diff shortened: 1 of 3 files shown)


ddp-workspace/projects/toolkit/src/lib/components/activity/activity.component.ts
```diff
@@ -10,7 +10,8 @@ import { ActivityResponse } from 'ddp-sdk';
         <toolkit-header [showButtons]="false"
                         [stickySubtitle]="stickySubtitle">
         </toolkit-header>
-        <ddp-activity [studyGuid]="studyGuid"
+        <ddp-activity *ngIf="id"
+                      [studyGuid]="studyGuid"
```
MocanaAtBroad, line 14: Could we rename instance variable "id" to "instanceGuid"? (Not sure how we ended up naming the param id, but at least it is used consistently)

MocanaAtBroad (commented):
Looks like you solved the problem. Very small refactor request.

aglushko-broad (dismissed):

ddp-workspace/projects/toolkit/src/lib/components/activity/activity.component.ts
alextanlan: Sure. 
Done at 43f7e4f72aa16edc94f25ca0c02ccbecbe2cbffa

aglushko-broad (approved):

PR #569, opened 2021-05-20
DDP-5926  Open selected participant dashboard in prism

aglushko-broad:
In order to open the user dashboard in prism we decided to follow a new approach and instead of temporal saving the selected user data into the session open the URL with user guid and request all user data from the backend. 
What was done:
- extend angio routing with dashboard/:id
- extend dashboard components in order to take the selected user data/guid input
- extend services for dashboard in order to be able to make requests with guid from the session and specific one
- extend subject panel component to display more user data (for now we pass each displayed data row as a separate input to make the component compatible with testboston which can provide only invitationId instead of valid participant interface)
- save search result in sessionStorage in order to be able to return from the dashboadr to the search table and see the same list of results
- introduce sessionStorage service
- introduce mockComponent test util
- move general subject panel styles from testboston to sdk

image


ddp-workspace/projects/ddp-rgp/src/app/components/activity/activity.component.html
```diff
@@ -12,7 +12,7 @@
 
   <ng-container *ngIf="isLoaded && model">
     <section class="section">
-      <ddp-subject-panel></ddp-subject-panel>
+      <ddp-subject-panel [invitationId]="invitationId"></ddp-subject-panel>
```
aglushko-broad: @MocanaAtBroad do we need this panel here at all? I'm not sure there is an admin account for rgp. Or do we plan to add it in the future?

ddp-workspace/projects/ddp-atcp/src/app/components/activityForm/app-atcp-activity.component.ts
```diff
@@ -31,7 +32,7 @@ import * as Routes from '../../router-resources';
   <main class="main main_activity" [ngClass]="{'main_sticky': isLoaded && model && model.subtitle, 'feeding-survey': model?.activityCode === ActivityCodes.FEEDING}">
         <ng-container *ngIf="isLoaded && model">
             <section class="section">
-                <ddp-subject-panel></ddp-subject-panel>
+                <ddp-subject-panel [invitationId]="invitationId"></ddp-subject-panel>
```
aglushko-broad: @MocanaAtBroad do we need this panel here? I'm not sure there is an admin account for atcp. If we plan to add it in the future I'd keep it otherwise I'd prefer to remove it to simplify the logic a little bit.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-redesigned.component.html
```diff
@@ -7,7 +7,7 @@
 >
     <ng-container *ngIf="isLoaded && model">
         <section class="section">
-            <ddp-subject-panel></ddp-subject-panel>
+            <ddp-subject-panel [invitationId]="invitationId"></ddp-subject-panel>
```
MocanaAtBroad: I am seeing that we are passing invitationId as an input param value in multiple places that we use ddp-subject-panel.
I wonder if this is necessary. Appears that invitation id is found is SessionMementoService and subject panel has its own reference to the service, to maybe we can just drop having invitationId as as an input param?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -210,11 +213,13 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
         private renderer: Renderer2,
         private submitService: SubmitAnnouncementService,
         private analytics: AnalyticsEventsService,
+        private session: SessionMementoService,
         @Inject(DOCUMENT) private document: any,
         // using Injector here as we get error using constructor injection
         // in both child and parent classes
         injector: Injector) {
         super(injector);
+        this.invitationId = this.session.session.invitationId;
```
MocanaAtBroad: What I was referring to in comment above: subject-panel has its own reference to SessionMementoService, so don't think finding invitationId in this component helps us much.

ddp-workspace/projects/ddp-testboston/src/assets/i18n/zh.json
```diff
@@ -619,7 +619,10 @@
     "TooltipAlt": "信息",
     "SubjectPanel": {
       "Link": "返回受试者查找",
-      "Text": "受试者邀请代码："
+      "InvitationCode": "受试者邀请代码：",
+      "Email": "[ZH] Email:",
+      "Name": "[ZH] Name:",
+      "ShortId": "[ZH] Subject Short ID/HRUID:"
```
MocanaAtBroad, line 625: Don't think this is hurting anything, but the subject panel will only be viewable by study administrators and while we do support a ton of different languages for testboston participants, it is OK (at this point) to show subject panel labels always in English.
At any rate, this is fine, though.

MocanaAtBroad (commented):
Looks good overall. Have question about use input params in subject panel component.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-redesigned.component.html
aglushko-broad: Theoretically, we can make invitationId as optional input and in case it wasn't passed try to find it in the session, we can do it, but still, I'd prefer to have it as an input for the selected user via a new prism in angio since we don't put invitationId into session in angio, only in testboston, wdyt?

ddp-workspace/projects/ddp-testboston/src/assets/i18n/zh.json
aglushko-broad: yes, I realized it later but decided keeping this since the work was already done :)

MocanaAtBroad (dismissed):

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-redesigned.component.html
aglushko-broad: I updated subjectPanel component, now it takes just one argument - subject instead of lists of inputs. This helps to not modify other studies

MocanaAtBroad (dismissed):
Excellent!

ddp-workspace/projects/ddp-sdk/src/lib/components/prism/prism.component.spec.ts
```diff
@@ -223,6 +216,14 @@ describe('PrismComponent', () => {
     expect(noUsersElement.textContent.trim()).toBe('NoUsers test');
   }));
 
+  it('sets initial data correctly', fakeAsync(() => {
```
alextanlan: I believe you don't need `fakeAsync` here (and at line 403 as  well)

ddp-workspace/projects/ddp-sdk/src/lib/components/subjectPanel/subjectPanel.component.ts
```diff
@@ -0,0 +1,62 @@
+import { Component, Inject, Input, OnInit } from '@angular/core';
+import { SessionMementoService } from '../../services/sessionMemento.service';
+import { ConfigurationService } from '../../services/configuration.service';
+import { InvitationPipe } from '../../pipes/invitationFormatter.pipe';
+import { SearchParticipant } from '../../models/searchParticipant';
+
+interface SubjectField {
```
alextanlan, line 7: #optional
suggest moving the interface to models folder

ddp-workspace/projects/ddp-sdk/src/lib/components/subjectPanel/subjectPanel.component.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    constructor(
+        private session: SessionMementoService,
+        private invitationPipe: InvitationPipe,
+        @Inject('ddp.config') private config: ConfigurationService) { }
+
+    ngOnInit(): void {
+        if (this.subject) {
+            this.subjectFields = [
+                ...(this.subject.firstName || this.subject.lastName
```
alextanlan, line 30: interesting idea with `[...[]]`

ddp-workspace/projects/ddp-sdk/src/lib/components/subjectPanel/subjectPanel.component.ts
aglushko-broad: I'd prefer to keep it here since it's not a global entity interface but rather a helper interface for component view

ddp-workspace/projects/ddp-sdk/src/lib/services/sessionStorage.service.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    let sessionStorageMock: Storage;
+
+    const sessionItem = 'sessionResult';
+    beforeEach(() => {
+        sessionStorageMock = {
+            getItem: jasmine.createSpy('getItem').and.returnValue(sessionItem),
+            setItem: jasmine.createSpy('setItem'),
+            removeItem: jasmine.createSpy('removeItem'),
+            clear: jasmine.createSpy('clear'),
+        } as unknown as Storage;
```
alextanlan, line 15: ` as unknown as Storage`
what can it mean?

aglushko-broad: when described object differs from the interface simple assertion doesn't help, "as unknown as" is a known trick in unit tests to provide some mock data with only necessary properties for complicated interfaces; it's also might be used in the code when there are difficulties with types compatibility from 3d party libraries.

ddp-workspace/projects/ddp-sdk/src/lib/components/prism/prism.component.spec.ts
aglushko-broad: fixed!

ddp-workspace/projects/ddp-sdk/src/lib/services/sessionStorage.service.spec.ts
alextanlan: Cool!
I did not know that.

alextanlan (approved):

PR #572, opened 2021-05-21
DDP-5926 Turn on unit tests run for toolkit and each study

aglushko-broad:


ddp-workspace/projects/toolkit-prion/src/test.ts
```diff
@@ -1,6 +1,6 @@
 // This file is required by karma.conf.js and loads recursively all the .spec and framework files
 
-import 'core-js/es7/reflect';
+// import 'core-js/es7/reflect';
```
alextanlan, line 3: Just wondering why you comment this ?

aglushko-broad: it throws an error. I wanted to remove it first, but then found out it's commented in other studies in polyfills.ts files. Toolkit doesn't have polyfills file. So I decided to follow the same pattern and commented it.

alextanlan (approved):

MocanaAtBroad (approved):
Looks good! Want to keep going and fix the linter (good: call linter on study project, sdk and toolkit. Better: call it on study project and sdk and/or toolkit if they have been modified). Either is better than having linter take longer to run than all the tests!.

PR #581, opened 2021-05-25
DDP-5806 File Upload extended functionality

alextanlan:
**_Update for 2 Jun, 2021:_**

image


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.html
```diff
@@ -1,27 +1,51 @@
+<!--  TODO: Add i18n  -->
 <ddp-question-prompt [block]="block"></ddp-question-prompt>
-<div class="file-upload-content">
+
+<div class="file-requirements" *ngIf="fileMaxSize || fileMimeTypes.length">
+    <div class="file-max-size" *ngIf="fileMaxSize">Max size: {{fileMaxSize}}</div>
+    <div class="file-types" *ngIf="fileMimeTypes.length">Allowed mime-types: {{fileMimeTypes.join(', ')}}</div>
+</div>
+
+<div class="file-upload-content" [ngClass]="{'readonly': readonly}">
```
aglushko-broad: #minor there is a recomendation from Angular team 
https://angular.io/guide/built-in-directives#adding-and-removing-classes-with-ngclass
"To add or remove a single class, use class binding rather than NgClass."
means `[class.readonly]="readonly"`

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.scss
```diff
@@ -9,22 +30,23 @@
...
+    background-color: $drop-block-color;
     display: flex;
     flex-direction: column;
     justify-content: center;
-}
 
-.drop-block.file-over {
-    background-color: lightyellow;
-}
+    &.file-over {
```
aglushko-broad, line 42: wondering if we need to change the styles for that for readonly state

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.html
```diff
@@ -1,27 +1,51 @@
...
+
+        <button *ngIf="!readonly" #uploadBtn
+                mat-raised-button color="primary"
+                class="submit-btn"
+                [disabled]="!isFileSelected"
+                (click)="submitFileUpload()">{{uploadedFile?.isUploaded? 'Reupload' : 'Submit'}}
+        </button>
     </div>
-    <div class="file-info">
-        <span class="file-name">{{file?.name || 'No file uploaded'}}</span>
```
aglushko-broad, line 18: don't we need "No file uploaded" message anymore?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.html
```diff
@@ -1,27 +1,51 @@
...
+
+    <div class="uploaded-file">
+        <div class="title">Uploaded:</div>
+
+        <mat-chip-list>
```
aglushko-broad, line 41: are we going to support multiple files uploading in the future? otherwise we probably can omit mat-chip-list

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.spec.ts
```diff
@@ -0,0 +1,203 @@
...
+                uploadUrl: ',',
+                name: '1.png',
+                size: 1000,
+                isUploaded: true
+            });
+        });
+
+        it('should display an uploaded file data', () => {
+            const uploadedFile = fixture.debugElement.query(By.css('.uploaded-file-chip')).nativeElement;
+            expect(uploadedFile.textContent.trim()).toBe('1.png (size: 1000) cancel');
```
aglushko-broad: #minor we can skip icon content using toContain instead toBe method

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.spec.ts
```diff
@@ -0,0 +1,203 @@
...
+        });
+
+        it('should call undoUploadedFile by click on remove uploaded file button', () => {
+            spyOn(component, 'undoUploadedFile');
+            const removeUploadedFileBtn = fixture.debugElement.query(By.css('.uploaded-file mat-icon')).nativeElement;
+            expect(removeUploadedFileBtn).toBeTruthy();
```
aglushko-broad: #minor we can remove this check since if it's not truthy the test will throw error and fail anyway

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.spec.ts
```diff
@@ -0,0 +1,203 @@
...
+            const localValidator = new ActivityFileValidationRule({} as any);
+            localValidator.result = errorMessage;
+            localValidator.recalculate = () => false;
+            component.block.validators = [localValidator];
+
+            component.submitFileUpload();
+            expect(component.errorMessage).toBe(errorMessage);
+        });
+    });
+});
```
aglushko-broad, line 194: Cool!

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
```diff
@@ -58,31 +81,130 @@ export class ActivityFileAnswer implements OnDestroy {
...
+    private initFileRequirements(): void {
+        this.fileMaxSize = this.block.maxFileSize;
+        this.fileMimeTypes = this.block.mimeTypes || [];
+    }
+
+    private initUploadedFile(): void {
+        if (this.block?.answer) {
+            this.setUploadedFile({
+                uploadGuid: '',
+                uploadUrl: ',',
```
aglushko-broad: should it be just an empty string?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
```diff
@@ -58,31 +81,130 @@ export class ActivityFileAnswer implements OnDestroy {
...
+        }
+
+        if (this.uploadedFile) {
+            this.openConfirmDialog();
+        } else {
+            this.uploadFile();
+        }
+    }
+
+    public openConfirmDialog(): void {
```
aglushko-broad: should it be private?

aglushko-broad: can we rename to 'openReuploadConfirmDualog' to better specify the goal?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
```diff
@@ -58,31 +81,130 @@ export class ActivityFileAnswer implements OnDestroy {
...
+        const dialogRef = this.dialog.open(ConfirmDialogComponent, config);
+        dialogRef.afterClosed().subscribe((confirmUpload: boolean) => {
+            if (confirmUpload) {
+                this.uploadFile();
+            }
+        });
+    }
+
+    undoUploadedFile(): void {
+        const panelClass = 'file-upload-confirm-dialog';
```
aglushko-broad: if this panel class is common for all dialogs of this component can we move it to private readonly class property?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
```diff
@@ -58,31 +81,130 @@ export class ActivityFileAnswer implements OnDestroy {
...
+                });
+                this.resetSelectedFile();
+                this.errorMessage = '';
             });
     }
 
-    ngOnDestroy(): void {
-        this.ngUnsubscribe.next();
-        this.ngUnsubscribe.complete();
+    private setUploadedFile(file: UploadFile | null): void {
```
aglushko-broad: #minor I'm not sure we need this set method... are we going to extend it with more functionality? if not probably it's fine to use the explicit set in the code

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
```diff
@@ -58,31 +81,130 @@ export class ActivityFileAnswer implements OnDestroy {
...
+        this.uploadedFile = file;
+    }
+
+    private patchAnswer(file: UploadFile | null): void {
+        this.block.answer = file ? { fileName: file.name, fileSize: file.size } as ActivityFileAnswerDto : null;
```
aglushko-broad: why do we need type casting here?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
```diff
@@ -58,31 +81,130 @@ export class ActivityFileAnswer implements OnDestroy {
...
+        this.isFileSelected = false;
+    }
+
+    private getFailedLocalValidator(): ActivityFileValidationRule {
+        return this.block.validators
+            .filter(validator => validator instanceof ActivityFileValidationRule)
+            .find(validator => {
+                const preUploadFileAnswer: ActivityFileAnswerDto = {
+                    fileName: this.fileToUpload && this.fileToUpload.name,
+                    fileSize: this.fileToUpload && this.fileToUpload.size
```
aglushko-broad: #minor 

fileName: this.fileToUpload?.name,                    
fileSize: this.fileToUpload?.size

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
```diff
@@ -1,51 +1,74 @@
...
 
-    file: File;
-    fileUploadData: FileUploadResponse;
+    selectedFile: File;
     isFileSelected: boolean;
+    uploadedFile: UploadFile | null;
+    fileMaxSize: number;
+    fileMimeTypes: string[];
+    errorMessage: string;
+    private fileToUpload: UploadFile | null;
```
aglushko-broad: it seems like it was better before with fileUploadData property. In fact, fileToUpload property is a combination of fileUploadData property and selectedFile property which means we duplicate information in the class instance state. I'd get rid of selectedFile then or get back to fileUploadData and just create the fileToUpload object locally in methods where it's needed.

ddp-workspace/projects/ddp-sdk/src/lib/models/uploadFile.ts
```diff
@@ -0,0 +1,7 @@
+import { FileUploadResponse } from './fileUploadResponse';
+
+export interface UploadFile extends FileUploadResponse {
```
aglushko-broad: I'm not sure if there is a benefit of putting the local component interfaces into the model folder. This model does not participate in contracts between backend and frontend and is not reused between different parts of the app. My suggetion is keeping such interfaces in the component which uses it. To keep the scope small enough.

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityFileValidationRule.ts
```diff
@@ -0,0 +1,36 @@
...
+            this.result = 'The file mime-type is not allowed';
+            return false;
+        }
+
+        this.result = null;
+        return true;
+    }
+
+    private isMaxFileSizeExceeded(answer: ActivityFileAnswerDto): boolean {
+        return this.block.maxFileSize && answer && answer.fileSize > this.block.maxFileSize;
```
aglushko-broad: #minor answer && answer.fileSize => answer?.fileSize
same for the line 33

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.scss
alextanlan: It is overridden on line 13:

```
.file-upload-content.readonly {
    opacity: 0.7;

    .drop-block {
        background-color: $drop-block-color;
        cursor: not-allowed;
    }
```

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.html
alextanlan: There is no design mockup for that.
I leave an empty space if no uploaded file, for now.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.html
alextanlan: There is no any design mockup for 'File Upload' functionality.
Not sure if we are going to support multiple files uploading.
I have chosen `mat-chip` because it is already an existing material component with included remove functionality.

aglushko-broad: I see, optionally you can consider just remove icon-button next to the file name

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.html
alextanlan: fixed at 4e8150675e0c6ee9724cd52ec02d8f73120cc52a

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.spec.ts
alextanlan: Thanks, fixed at 4e8150675e0c6ee9724cd52ec02d8f73120cc52a

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
alextanlan: Just for more comprehension)
Removed at 4e8150675e0c6ee9724cd52ec02d8f73120cc52a

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
alextanlan: Fixed at 4e8150675e0c6ee9724cd52ec02d8f73120cc52a, thanks.

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityFileValidationRule.ts
alextanlan: Ok, fixed at 4e8150675e0c6ee9724cd52ec02d8f73120cc52a

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
alextanlan: Technically yes, it could be.
However it is used in several test cases.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
alextanlan: let's keep it as is

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.spec.ts
alextanlan: Fixed at 0f692defcc2f194893faa6d53ca7b38a2ac473fc

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
alextanlan: Right, it is a typo, thanks.
fixed at 0f692defcc2f194893faa6d53ca7b38a2ac473fc

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
alextanlan: Deal, fixed at 0f692defcc2f194893faa6d53ca7b38a2ac473fc

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
alextanlan: Fixed at 0f692defcc2f194893faa6d53ca7b38a2ac473fc

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.html
```diff
@@ -1,27 +1,51 @@
+<!--  TODO: Add i18n  -->
 <ddp-question-prompt [block]="block"></ddp-question-prompt>
-<div class="file-upload-content">
+
+<div class="file-requirements" *ngIf="fileMaxSize || fileMimeTypes.length">
+    <div class="file-max-size" *ngIf="fileMaxSize">Max size: {{fileMaxSize}}</div>
+    <div class="file-types" *ngIf="fileMimeTypes.length">Allowed mime-types: {{fileMimeTypes.join(', ')}}</div>
```
MocanaAtBroad: Strings need to be internationalized.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.html
```diff
@@ -1,27 +1,51 @@
...
             <mat-icon>attach_file</mat-icon>
         </button>
 
+        <div class="selected-file">{{selectedFile?.name}}</div>
+
+        <button *ngIf="!readonly" #uploadBtn
+                mat-raised-button color="primary"
+                class="submit-btn"
+                [disabled]="!isFileSelected"
+                (click)="submitFileUpload()">{{uploadedFile?.isUploaded? 'Reupload' : 'Submit'}}
```
MocanaAtBroad: Internationalization.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.html
```diff
@@ -1,27 +1,51 @@
...
+        </button>
     </div>
-    <div class="file-info">
-        <span class="file-name">{{file?.name || 'No file uploaded'}}</span>
+
+    <div class="uploaded-file">
+        <div class="title">Uploaded:</div>
```
MocanaAtBroad: Same. Please check all code for strings that need to be i18n'ed.

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityFileValidationRule.ts
```diff
@@ -0,0 +1,36 @@
...
+export class ActivityFileValidationRule extends ActivityAbstractValidationRule {
+    constructor(private block: ActivityFileQuestionBlock) {
+        super(block);
+    }
+
+    public recalculate(answerBeforeUpload?: ActivityFileAnswerDto): boolean {
+        const answer = answerBeforeUpload || this.block.answer;
+
+        if (this.isMaxFileSizeExceeded(answer)) {
+            this.result = 'File size exceeded maximum of 100 bytes';
```
MocanaAtBroad: Internationalization and hard-coded value.

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityFileValidationRule.ts
```diff
@@ -0,0 +1,36 @@
+import { ActivityAbstractValidationRule } from './activityAbstractValidationRule';
+import { ActivityFileQuestionBlock } from '../../../models/activity/activityFileQuestionBlock';
+import { ActivityFileAnswerDto } from '../../../models/activity/activityFileAnswerDto';
+
+export class ActivityFileValidationRule extends ActivityAbstractValidationRule {
```
MocanaAtBroad, line 10: Actually, does this rule ever execute?
When trying add a large file, the displayed error appears to come from server.

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityFileValidationRule.ts
```diff
@@ -0,0 +1,36 @@
...
+    constructor(private block: ActivityFileQuestionBlock) {
+        super(block);
+    }
+
+    public recalculate(answerBeforeUpload?: ActivityFileAnswerDto): boolean {
```
MocanaAtBroad, line 15: If we are going to execute this rule or show the max size for file, let's display the file size in either KBs or MBs instead of just bytes. In Mac and Windows files in bytes are rarely seen, and we people should not have to do calculation.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.html
MocanaAtBroad: Let's convert max size to Kilobytes or Megabytes using appropriate abbreviation.

MocanaAtBroad: We probably need to refer to file types or file extensions. People don't like mimes.

ddp-workspace/projects/ddp-sdk/src/lib/services/modal.service.ts
```diff
@@ -14,26 +14,26 @@ const DIALOG_HEIGHT = 160;
 const VERTICAL_GAP = 15; // margin between dialog and delete button
 
 @Injectable()
-export class ActivityBlockModalService {
+export class ModalService {
```
MocanaAtBroad: I think the refactor is a good idea.
Can we rename to ModalDialogService?

ddp-workspace/projects/ddp-angio/src/assets/i18n/en.json
```diff
@@ -699,6 +699,13 @@
             "Email": "Email:",
             "Name": "Name:",
             "ShortId": "Subject Short ID/HRUID:"
+        },
+        "FileUpload": {
+            "ConfirmReuploadTitle": "Are you sure you want \nto reupload the file?",
+            "ConfirmUndoUploadTitle": "Are you sure you want \nto undo the uploaded file?",
```
MocanaAtBroad: Consider rewording to something like: "Are you sure you want to remove uploaded file from server?"

ddp-workspace/projects/ddp-angio/src/assets/i18n/en.json
```diff
@@ -699,6 +699,13 @@
...
+            "ReuploadBtnText": "Reupload",
+            "UndoBtnText": "Undo",
```
MocanaAtBroad: I think "OK" would be more familiar than "Undo" to users.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.html
MocanaAtBroad: We should give the button a label like "Upload" rather than Submit. We have been using "Submit" at the completion of forms.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.html
```diff
@@ -1,27 +1,51 @@
...
+
+        <mat-chip-list>
+            <mat-chip *ngIf="uploadedFile?.isUploaded"
+                      class="uploaded-file-chip"
+                      color="primary" selected
+                      (removed)="undoUploadedFile()">{{uploadedFile.name}} (size: {{uploadedFile.size}})
+                <mat-icon matChipRemove *ngIf="!readonly" #undoUploadBtn>cancel</mat-icon>
```
MocanaAtBroad: We can actually tell the server to delete the upload if they press the cancel button.
The way to do this is to PATCH an answer with a null value.
So right now, after the client uploads the contents of the file, your code correctly issues a PATCH that looks like:
```json
{"answers":[{"stableId":"FILE_UPLOAD_1","answerGuid":"SM27JVPJDB","value":"2kn3N8lkJXzmh12sFEMj"}]}
```
To delete the upload (i.e., when the matChipRemove is clicked) we would want to PATCH:
```json
{"answers":[{"stableId":"FILE_UPLOAD_1","answerGuid":"SM27JVPJDB","value":null}]}

File: ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts [hunk 27]
```diff
@@ -1,88 +1,212 @@
...
-        formData.append('file', this.file);
+    submitFileUpload(): void {
+        const failedLocalValidator = this.getFailedLocalValidator();
+        if (failedLocalValidator) {
+            this.errorMessage = failedLocalValidator.result;
+            return;
+        }
+
+        if (this.uploadedFile) {
+            this.openReuploadConfirmDualog();
```
MocanaAtBroad: Typo.

ddp-workspace/projects/ddp-angio/src/assets/i18n/en.json
alextanlan: Fixed at 9a8302427db3a7037cfa1c1715618308c8ff8e51

ddp-workspace/projects/ddp-angio/src/assets/i18n/en.json
alextanlan: Fixed at 9a8302427db3a7037cfa1c1715618308c8ff8e51

ddp-workspace/projects/ddp-sdk/src/lib/services/modal.service.ts
alextanlan: Renamed at fdaba73aac86dc5241544432083394fae34022a1

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityFileValidationRule.ts
alextanlan: Extracted hard-coded values at 9e99630b172eb4b407212bdcde6dc0fc3d1b35c7

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.html
alextanlan: Renamed the label at 1e5f66ed5e3c0ab09f821ff006e5324212f4d37e

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
alextanlan: Fixed at 4ccdfdfaeaaf697e67cfe838da8109dc0fda5462.
Thanks!

ddp-workspace/projects/ddp-sdk/src/lib/models/uploadFile.ts
alextanlan: I don't think it is an issue. 
@MocanaAtBroad, what is your opinion ?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.html
alextanlan: **_As for i18n:_**
I have added i18n for ddp-angio only (in test purpose) at 96f90878ac4a661c1aa6a55c2216e04e08aabeb3

we don't know for now where the File Upload functionality will be applied (in which activity).
It does not make sense to apply FileUpload translations for all existing studies right now, in my opinion. 

What do you think about it?

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityFileValidationRule.ts
alextanlan: It's my pain)
Before the implementation I have been told that there is no any server pre-validation. 
Server only sends fields `maxFileSize`/ `allowed mime-types` as no notify client about the rules.
I created a local client validation using these fields via `ActivityFileValidationRule` and it executed.
Then (suddenly) Eugenii implemented server-side validation on getUploadUrl step.
And for now the client-side validation is kind of `backup validation`.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.html
alextanlan: Right. It should work exactly like you wrote.
For example:
1.Upload a file
upload1

2. Deleting the already uploaded file
upload2

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.html
alextanlan: Added format:
- file size to Mb/kB
- mime-type to file extension
f49dcc17a39e5a7d0292664acf067ce18b74151a

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityFileValidationRule.ts
alextanlan: Added file size format pipe as to transform to Mb/kb and use in in template.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
alextanlan: Refactored at 497c9ebd1345abb708c9fd0851e40d38fe7405b5

ddp-workspace/projects/ddp-sdk/src/lib/components/confirmDialog/confirmDialog.component.html
```diff
@@ -0,0 +1,16 @@
+<div class="confirm-dialog">
+    <span class="confirm-dialog-title"
+          [innerHTML]="title | translate">
+    </span>
+    <div class="confirm-dialog-buttons">
+        <button mat-flat-button class="ctrl-btn confirm-btn"
+                color="warn"
```
aglushko-broad: are we sure that confirm button should be in red color?

aglushko-broad:
I'm wondering if we can upload a file automatically after the user selected the file? Why do we need to additionally ask her/him to click the upload button?

ddp-workspace/projects/ddp-sdk/src/lib/components/confirmDialog/confirmDialog.component.html
alextanlan: I am not. 
However, the dialogs for now notify about **_Remove_**/**_Re-upload_** a file. 
They are needed to highlight that the operation can not be undo. 

image

image

alextanlan: Do you have some suggestions about it ?

aglushko-broad: well, the color might be configurable I guess, but it might be extended later if we feel that the red color is the right one here

alextanlan:
> I'm wondering if we can upload a file automatically after the user selected the file? Why do we need to additionally ask her/him 
> to click the upload button?

1. There are no requirements/design mockups for it. Perhaps, we will need to change it when it is used in a real study. 
2. In my opinion, having this interim step a user can have a chance to
       - change their mind if a wrong file is selected by mistake
       - see the file name when a validation error occurs as to easily analyze the error

ddp-workspace/projects/ddp-sdk/src/lib/components/confirmDialog/confirmDialog.component.html
alextanlan: @aglushko-broad, which color are you suggesting to have there by default ?

alextanlan: Another thing is that the confirmation dialog is used in other places (i.g. modalActivityBlock component)

aglushko-broad: I don't have any suggestion, my comment initially was rather about "should we just copy it from original dialog or maybe even discuss with the designer what's better here" but as long as I'm not a designer I can't say which color might be better here. As I already wrote in the previous comment if red sounds right for you feel free to do nothing and ignore the comment. It wouldn't be a problem to add config functionality in the future if we need it.

alextanlan: Ok, then I propose leaving it as is for now because we do not have any comprehension how it will be used later.

ddp-workspace/projects/ddp-sdk/src/lib/models/uploadFile.ts
MocanaAtBroad: I think it is OK to for it to be in the `models` directory as interface gets used from multiple places (component and service) we don't have a better home for it.
But @aglushko-broad observation about a distinction between classes and interfaces that are part of the client/server API vs those that are just for internal consumption is a good one and we should consider doing a little bit of refactoring here. (I did a quick count. We have 63 files directly in model directory, so probably time to reorganize the closet).

ddp-workspace/projects/ddp-angio/src/assets/i18n/en.json
```diff
@@ -699,6 +699,22 @@
...
+            "UploadBtnText": "Upload",
+            "UndoBtnText": "OK",
+            "CancelBtnText": "Cancel",
+            "MaxSize": "Max size: {{maxSize}}",
+            "AllowedFiles": "Allowed files: {{allowedFileTypes}}",
+            "DragFiles": "Drag files to upload",
+            "ChooseFile": "Choose file",
+            "Uploaded": "Uploaded:",
+            "Size": "(size: {{size}})",
+            "MaxSizeErrorMessage": "File size exceeds allowed size",
```
MocanaAtBroad: Please reword to: "Maximum file size exceeded" (avoids using the word "size" twice).

ddp-workspace/projects/ddp-angio/src/assets/i18n/en.json
```diff
@@ -699,6 +699,22 @@
...
+            "InvalidTypeErrorMessage": "File type not belongs to allowed list"
```
MocanaAtBroad: "Type of file selected is not valid"

MocanaAtBroad (commented):
Still looking...

ddp-workspace/projects/ddp-angio/src/assets/i18n/en.json
```diff
@@ -699,6 +699,22 @@
...
+        },
+        "FileUpload": {
+            "ConfirmReuploadTitle": "Are you sure you want \nto reupload the file?",
+            "ConfirmUndoUploadTitle": "Are you sure you want \nto remove uploaded file from server?",
+            "ReuploadBtnText": "Reupload",
+            "UploadBtnText": "Upload",
+            "UndoBtnText": "OK",
+            "CancelBtnText": "Cancel",
+            "MaxSize": "Max size: {{maxSize}}",
+            "AllowedFiles": "Allowed files: {{allowedFileTypes}}",
```
MocanaAtBroad: `AllowedFileTypes` probably more descriptive.

ddp-workspace/projects/ddp-sdk/src/lib/models/fileSizeMeasure.ts
```diff
@@ -0,0 +1,4 @@
+export enum FILE_SIZE_MEASURE {
+    kB = 1024,
```
MocanaAtBroad: In my Mac, I see file size listed using "KB" and "MB" so better if we go with these abbreviations.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
```diff
@@ -1,45 +1,72 @@
...
-
-    file: File;
-    fileUploadData: FileUploadResponse;
-    isFileSelected: boolean;
+    @ViewChild('uploadBtn', {read: ElementRef}) private uploadButtonRef: ElementRef;
+    @ViewChild('undoUploadBtn', {read: ElementRef}) private undoUploadButtonRef: ElementRef;
+    fileToUpload: UploadFile | null;
+    uploadedFile: ActivityFileAnswerDto | null;
+    fileMaxSize: number;
+    allowedFileTypes: string[];
```
MocanaAtBroad: Consider getting rid of instance variables. fileMaxSize and allowedFileTypes come directly from `block`; just copies.

ddp-workspace/projects/ddp-sdk/src/lib/models/uploadFile.ts
```diff
@@ -0,0 +1,6 @@
...
+    file: File;
+    isReadyToUpload: boolean;
```
MocanaAtBroad: Not finding how property `isReadyToUpload` is being used; only see places where the value is set to `true` or `false` but don't see anywhere it is read. Could you double-check `isReadytoUpload` is needed?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
```diff
@@ -52,37 +79,127 @@ export class ActivityFileAnswer implements OnDestroy {
                 .pipe(
```
MocanaAtBroad, line 65: Consider refactoring the method signature for `getUploadUrl` . This component should not care what the request body for REST API looks like.  Might be better keeping this structural knowledge out of component and include fileName, fileSize, etc as arguments in `getUploadUrl` rather than `FileUploadBody`

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
```diff
@@ -52,37 +79,127 @@ export class ActivityFileAnswer implements OnDestroy {
...
+                fileSize: this.block.answer.fileSize
+            });
+        }
+    }
+
+    private uploadFile(): void {
+        this.fileUploadService.uploadFile(this.fileToUpload.uploadUrl, this.fileToUpload.file)
             .pipe(
                 catchError(err => {
                     this.logger.logDebug('ActivityFileAnswer uploadFile error:', err);
```
MocanaAtBroad, line 175: This line gives me Typescript compiler error. Looks to me like we need to set the method signature in the LoggingService correctly (currently expects 0 arguments which is obviously incorrect).

ddp-workspace/projects/ddp-sdk/src/lib/pipes/fileSizeFormatter.pipe.ts
```diff
@@ -0,0 +1,14 @@
...
+    name: 'fileSize'
+})
+export class FileSizeFormatterPipe implements PipeTransform {
+    transform(value: number): string {
+        const isFileSizeLessThenOneMb = value < FILE_SIZE_MEASURE.Mb;
+        const fractionDigits = isFileSizeLessThenOneMb ? 0 : 2;
+        const measure = isFileSizeLessThenOneMb ? FILE_SIZE_MEASURE.kB : FILE_SIZE_MEASURE.Mb;
+        return (value / measure).toFixed(fractionDigits) + FILE_SIZE_MEASURE[measure];
+    }
+}
```
MocanaAtBroad, line 43: The pipe is a good idea!

MocanaAtBroad: We should deal with files that are less than 1KB in size (in our test configuration with have a file with max size of 100 bytes).
Screen Shot 2021-06-01 at 1 10 44 PM

MocanaAtBroad: I think there are a lot of threshold and corner cases here in when to use bytes, MBs decimals,etc.
I am seeing that this is problem that other people have tried to solve. Maybe we should just use the this library or copy the function?
https://www.npmjs.com/package/filesize

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
```diff
@@ -52,37 +79,127 @@ export class ActivityFileAnswer implements OnDestroy {
                 .pipe(
                     catchError(err => {
                         this.logger.logDebug('ActivityFileAnswer getUploadUrl error:', err);
+                        this.errorMessage = err.message;
```
MocanaAtBroad, line 68: We should clear `errorMessage` whenever a new file is selected
1. We select a file that is too big
2. UI shows error "File is too big" (or something like that)
3. We select a file that is of allowed size.
4. UI keeps showing "File is too big" error.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.html
```diff
@@ -1,27 +1,58 @@
...
         <button mat-raised-button color="primary" class="upload-btn"
-                (click)="fileUpload.click()">Choose file
+                [disabled]="readonly"
+                (click)="fileUpload.click()">{{'SDK.FileUpload.ChooseFile' | translate}}
             <mat-icon>attach_file</mat-icon>
         </button>
 
+        <div class="selected-file">{{fileToUpload?.file.name}}</div>
+
+        <button *ngIf="!readonly" #uploadBtn
```
MocanaAtBroad: We should disable or hide "upload" button if there is no file selected.

MocanaAtBroad:
>> alextanlan 5 days ago • 
 Author Member
As for i18n:
I have added i18n for ddp-angio only (in test purpose) at 96f9087

>> we don't know for now where the File Upload functionality will be applied (in which activity).
It does not make sense to apply FileUpload translations for all existing studies right now, in my opinion.

>> What do you think about it?

The way we currently do translations, we copy the translations entries from SDK to the project that is using SDK.
In this case, because feature is not being used at this time in any project, don't see reason to copy new entries. For new projects (even if they don't have file upload), I think OK for stuff to be copied.

But bottom line: no need to copy to translations to any existing study at this time.

MocanaAtBroad:
>> aglushko-broad 6 hours ago Member
I don't have any suggestion, my comment initially was rather about "should we just copy it from original dialog or maybe even discuss with the designer what's better here" but as long as I'm not a designer I can't say which color might be better here. As I already wrote in the previous comment if red sounds right for you feel free to do nothing and ignore the comment. It wouldn't be a problem to add config functionality in the future if we need it.

>> @alextanlan alextanlan 6 hours ago • 
 Author Member
Ok, then I propose leaving it as is for now because we do not have any comprehension how it will be used later.

I would not choose red for the button and use whatever we are using for confirmation elsewhere. Gray? Green?

But would add that at this time we don't really have a web design we are working out of and might have to revisit a few things, like the overall layout, fonts, etc. 
Would be nice to combine the button that opens the file selection dialog with the file drag and drop area; have one thing that you can click or drag files into instead of separate widgets.
But I think this is OK for now. Looking for the basic functionality to be in place and we can come back and revisit when we have a clean design.

MocanaAtBroad (commented):
> I'm wondering if we can upload a file automatically after the user selected the file? Why do we need to additionally ask her/him to click the upload button?

Would be good to get rid of additional step/need for button (I have MANY times personally thought I had uploaded attachments to later find out that I forgot to click the upload button)

ddp-workspace/projects/ddp-angio/src/assets/i18n/en.json
alextanlan: Fixed at fadf33db0ccdfa4570a290bc9f84d4fd41f8ddac

ddp-workspace/projects/ddp-angio/src/assets/i18n/en.json
alextanlan: Fixed at fadf33db0ccdfa4570a290bc9f84d4fd41f8ddac

ddp-workspace/projects/ddp-angio/src/assets/i18n/en.json
alextanlan: Fixed at fadf33db0ccdfa4570a290bc9f84d4fd41f8ddac

ddp-workspace/projects/ddp-sdk/src/lib/models/fileSizeMeasure.ts
alextanlan: Fixed at cc816fd0f9f2c1028aa89e71a22ca29403ea42db

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.html
alextanlan: Right. I lost it by mistake during a previous refactoring.
Returned back at b836160f17b9a3addca82f9f11b9faa17a2408b7.

ddp-workspace/projects/ddp-sdk/src/lib/models/uploadFile.ts
alextanlan: Thanks for noticing.
Please see https://github.com/broadinstitute/ddp-angular/pull/581#discussion_r643471373
Fixed at b836160f17b9a3addca82f9f11b9faa17a2408b7

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
alextanlan: Ok. Fixed at 1bfa63d0862c3e41cd28bbde460c29108f6d47de

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
alextanlan: It is interesting taking into account that `logger.logDebug` is called similarly in 26 places (except that one). 
Can you please provide a screenshot with Typescript compiler error ? 
Perhaps we just need to add `tslint:disable` there ?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
alextanlan: My bad. 
Fixed at 8d677f21d8cc286f7d0db97f4f7ce19197a03dcf

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
MocanaAtBroad: Not something you introduced in your PR anyways.
Fix  in this PR https://github.com/broadinstitute/ddp-angular/pull/595/files

alextanlan: Yes, I saw your fix. 
It's odd that we didn't see the error earlier. 
Another point is that `console.debug` should have a list of arguments "out-of-the-box" 
(https://console.spec.whatwg.org/#debug)

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
alextanlan: Fixed at c5dca4a0a0de154ddb2a0f45182c503a4294e7b0

ddp-workspace/projects/ddp-sdk/src/lib/components/confirmDialog/confirmDialog.component.html
alextanlan: Made the button color configurable and set the default 'primary' theme color palette at ec6bcb0eb8955cf5f4c58a4b5f1c314c9b9c1448 
image

Eventually, we can override the color in specific study css if needed.

alextanlan:
> Would be nice to combine the button that opens the file selection dialog with the file drag and drop area; have one thing that you can click or drag files into instead of separate widgets.

done at 56662b7502b58c3e4a7e0038d73d34c11fdeb282

alextanlan:
>Would be good to get rid of additional step/need for button (I have MANY times personally thought I had uploaded attachments to later find out that I forgot to click the upload button)

Removed Upload/Reupload button (as an additional step during the file uploading) at fd88a6eb17817ca6762f1b7201ebb5466eb0f148

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.html
alextanlan: Eventually removed (see https://github.com/broadinstitute/ddp-angular/pull/581#issuecomment-852943813)

ddp-workspace/projects/ddp-sdk/src/lib/models/uploadFile.ts
alextanlan: Eventually removed (see https://github.com/broadinstitute/ddp-angular/pull/581#issuecomment-852943813)

ddp-workspace/projects/ddp-sdk/src/lib/pipes/fileSizeFormatter.pipe.ts
alextanlan: Improved at b05a866d623b7e42d7aee4ea49f566984c9507e6

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
MocanaAtBroad: Just copied the console.debug method signature (I think the error has been there all along but it works like a "warning") the correct Javascript is generated anyways.

MocanaAtBroad:
We need to disable the submit button while we get the URL to do the authorization and while there is an upload in progress. We might be able to use the component busy mechanism we use from mailing address? Take a look.
(You probably know this already, but the trick I use to simulate a slow connection/longer time for upload is to turn on throttling in the Chrome Dev tools. Easier to see the submit button enabling/disabling)
Screen Shot 2021-06-03 at 9 04 56 AM

MocanaAtBroad:
I think we also need some kind of indicator that there is an upload in progress. Perhaps just a simple GIF? Perhaps some text saying something like "Uploading MyFile.txt..."
We don't have nice design to work from, but hopefully at some point we will and we can just swap out the images, tweak layouts, etc., but the basic elements and functionality will be there.

alextanlan:
> Was this implementation copied more or less from the filesize library I linked to?

Yes, it was. The thing is that the library has 'BSD-3-Clause' License. I was not sure if we can use it.
Therefore I extracted/copied the main idea of the transform method.

alextanlan:
> We need to disable the submit button while we get the URL to do the authorization and while there is an upload in progress. We might be able to use the component busy mechanism we use from mailing address? Take a look.

Implemented at 44fcefbe583704af9e3fd2f28cdd3063a3f731bc

alextanlan:
> I think we also need some kind of indicator that there is an upload in progress. Perhaps just a simple GIF? Perhaps some text saying something like "Uploading MyFile.txt..."


Added a progres-bar at 3f38c8b403dae066497a6459995bc2975340a0e3

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
```diff
@@ -1,88 +1,203 @@
...
+                private logger: LoggingService,
+                private dialog: MatDialog,
+                private modalDialogService: ModalDialogService) {
+    }
+
+    ngOnInit(): void {
+        this.initUploadedFile();
     }
 
     onFilesSelected(files: File[]): void {
```
MocanaAtBroad: I am getting a type warning in template. Change below gets rid of it.
```suggestion
    onFilesSelected(files: File[] | FileList): void {
```

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.html
```diff
@@ -1,27 +1,46 @@
...
-                (click)="fileUpload.click()">Choose file
-            <mat-icon>attach_file</mat-icon>
-        </button>
-
     </div>
-    <div class="file-info">
-        <span class="file-name">{{file?.name || 'No file uploaded'}}</span>
+
+    <div class="uploaded-file">
+        <div class="title" translate>SDK.FileUpload.Uploaded</div>
```
MocanaAtBroad: I don't think this label makes sense anymore. Even when it is true that it is "**Uploaded**", it is already clear from the mat-chip"
Screen Shot 2021-06-06 at 4 39 28 PM

Consider changing it to appear only when upload in progress and with the name of the file.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activitySection.component.ts
```diff
@@ -58,7 +58,8 @@ import { SubmissionManager } from '../../services/serviceAgents/submissionManage
                                            [readonly]="readonly"
                                            [validationRequested]="validationRequested"
                                            [studyGuid]="studyGuid"
-                                           [activityGuid]="activityGuid">
+                                           [activityGuid]="activityGuid"
+                                           (componentBusy)="embeddedComponentBusy.emit($event)">
```
MocanaAtBroad: I think this is the correct thing to do (emit true or false to indicate whether busy or not), but now the @Output() variable name is incorrect. We should probably rename now just `componentBusy` as it is just not embedded components that we are worried about.

ddp-workspace/projects/ddp-sdk/src/lib/models/uploadFile.ts
MocanaAtBroad: After last refactor and reviewing latest changes, I am under impression that this interface can just go away. Only thing in it that matters is the File (one attribute) and the upload url is the other argument to actually do the upload and appears it would be easier to just pass it as a function argument. **_MIGHT_** be able to get rid of fileToUpload instance variable altogether, but not 100% sure, but please do take a look. The less instance variables we have, the less state to track and less chance of subtle bugs coming into play.

MocanaAtBroad (commented):
Now that you have added the "Upload in progress" indicator, seems to me that the last piece of functionality missing here is to be able to cancel the upload. A "cancel" button with an "Are you sure you want to cancel upload?" confirmation dialog. Hard to add?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.ts
alextanlan: Fixed at f929b71f78f178edac82e717e16d6cc391afd774

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activitySection.component.ts
alextanlan: Renamed at e76e653f7aac823ccc154ca39530a64606c8a2cd

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.html
alextanlan: Changed at 8f3c18efd56ba1cbea0060bae1c156be4ffb7d7f

alextanlan:
> After last refactor and reviewing latest changes, I am under impression that this interface can just go away. Only thing in it that matters is the File (one attribute) and the upload url is the other argument to actually do the upload and appears it would be easier to just pass it as a function argument. MIGHT be able to get rid of fileToUpload instance variable altogether, but not 100% sure, but please do take a look. The less instance variables we have, the less state to track and less chance of subtle bugs coming into play.

Changed at 8f3c18efd56ba1cbea0060bae1c156be4ffb7d7f

alextanlan:
> Now that you have added the "Upload in progress" indicator, seems to me that the last piece of functionality missing here is to be able to cancel the upload. A "cancel" button with an "Are you sure you want to cancel upload?" confirmation dialog. Hard to add?

It seems to be tricky for me. The thing is that we patch answers "in background" and can not control from a component if the answer has been already patched.

MocanaAtBroad:
> > Now that you have added the "Upload in progress" indicator, seems to me that the last piece of functionality missing here is to be able to cancel the upload. A "cancel" button with an "Are you sure you want to cancel upload?" confirmation dialog. Hard to add?
> 
> It seems to be tricky for me. The thing is that we patch answers "in background" and can not control from a component if the answer has been already patched.

Well, I think the Cancel button would only be visible during the "Uploading" stage and disappear as soon as upload is completed/before the PATCH is queued on the client. Once upload is completed , it should be no more than a few seconds before they are able to click to remove the upload.

The case I am worried about is when we have really large files (e.g.., 100MB on a bad connection) and realize they uploaded wrong file. Seems like punishment to make them wait until upload is completed before they can do the real one.

alextanlan:
> I think the Cancel button would only be visible during the "Uploading" stage and disappear as soon as upload is completed/before the PATCH is queued on the client.

Thank you for the idea ! Applied at d11f84322337471b79b209ef1547cc905999a58c

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/embeddedActivityBlock/embeddedActivityBlock.component.ts
```diff
@@ -60,9 +60,16 @@ export class EmbeddedActivityBlockComponent implements OnInit {
...
+        config.data = {
+            title: 'SDK.ConfirmDeletion',
+            confirmBtnText: 'SDK.DeleteButton',
+            cancelBtnText: 'SDK.CancelBtn',
+            confirmBtnColor: 'warn'
+        };
 
-        const dialogRef = this.dialog.open(ActivityDeleteDialogComponent, config);
+        const dialogRef = this.dialog.open(ConfirmDialogComponent, config);
         dialogRef.afterClosed().subscribe((confirmDelete: boolean) => {
```
MocanaAtBroad, line 73: Do we have unsubscribe from dialog events? (Highlighting this instance but there are a couple other ones in file)

MocanaAtBroad (commented):
The cancellation looks and works nice!
Just had one question observation.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/embeddedActivityBlock/embeddedActivityBlock.component.ts
aglushko-broad: we don't need to unsubscribe from afterClosed result since the observable itself completes

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer/activityFileAnswer.component.html
MocanaAtBroad: Not needed in any of our current studies. But coming soon!

MocanaAtBroad (approved):

PR #594, opened 2021-06-01
DDP-6183 Fix a bug when embedded nested activities should prevent parent activity submission

alextanlan:
Fixed that embedded nested activities validate before parent activity submission.

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activitySection.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activitySection.component.ts
@@ -1,4 +1,15 @@
-import { ChangeDetectorRef, Component, EventEmitter, Inject, Input, OnDestroy, OnInit, Output } from '@angular/core';
+import {
+    ChangeDetectorRef,
+    Component,
+    EventEmitter,
+    Inject,
+    Input,
+    OnDestroy,
+    OnInit,
+    Output
+} from '@angular/core';
+import { Subscription } from 'rxjs';
+
 import { ActivitySection } from '../../models/activity/activitySection';
 import { ActivityBlock } from '../../models/activity/activityBlock';
 import { BlockType } from '../../models/activity/blockType';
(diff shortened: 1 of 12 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activitySection.component.ts
```diff
@@ -103,9 +114,10 @@ export class ActivitySectionComponent implements OnInit, OnDestroy {
     @Output() embeddedComponentsValidationStatus: EventEmitter<boolean> = new EventEmitter();
     @Output() embeddedComponentBusy: EventEmitter<boolean> = new EventEmitter(true);
     private subscription: Subscription;
-    private embeddedValidationStatus: boolean[] = new Array(2).fill(true);
+    private embeddedValidationStatus: boolean[] = new Array(3).fill(true);
```
aglushko-broad, line 117: It feels like this solution with static numbers may work not correctly in general (it's also not flexible but it's the second problem)
I can see we have a loop for section.blocks. So we can have multiple ddp-institutions-form, ddp-address-embedded and ddp-activity-block elements in the current component. In this case values in embeddedValidationStatus will be overwritten. Can we try to make the validation result value a part of the block interface?
In order to make the solution more flexible, we can use Map and string id keys instead of an array of some fixed length

aglushko-broad (approved):

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activitySection.component.ts
aglushko-broad: but it's not in the scope of the current task, so might be fixed later and separately

alextanlan: Good point. 
Created [DDP-6286](https://broadinstitute.atlassian.net/browse/DDP-6286) ticket for that.

PR #599, opened 2021-06-02
Ddp 6247 pancan bootstrap

aglushko-broad:


ddp-workspace/projects/ddp-pancan/.editorconfig
```diff
@@ -0,0 +1,13 @@
+# Editor configuration, see https://editorconfig.org
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 2
```
alextanlan, line 7: I'd  suggest 4 like in most of our studies. 
Also it is recommended there [DDP Angular Code style](https://broadinstitute.atlassian.net/wiki/spaces/DDP/pages/273874947/2.2.4+Code+style)

aglushko-broad: I see that we have a point "Use 4 spaces per indentation." in code style but absolutely all studies .editorconfig files contain indent_size = 2
@MocanaAtBroad could you please clarify what should we use here, thanks!

aglushko-broad:
@MocanaAtBroad currenlty pepperConfig can't be generated properly, the error message is shown "no secret exists at secret/pepper/dev/v1/pancan/conf"

aglushko-broad:
> @MocanaAtBroad currenlty pepperConfig can't be generated properly, the error message is shown "no secret exists at secret/pepper/dev/v1/pancan/conf"

not an issue anymore

ddp-workspace/projects/ddp-pancan/.editorconfig
MocanaAtBroad: I didn't even know this file existed! Looks like the autogenerated says 2 characters but we have been using 4. (I am assuming that 4 must be the IDE setting for Typescript?) At any rate, should be 4 but I don't think it really matters.

By the way: the link that Alex included for code styles was something that Mikhail with some input from me put together some years ago. And at that time the coding styles we were looking at were those of the typescript compiler. I would very much welcome revisiting. A lot has changed in Typescript and Angular since then.

MocanaAtBroad (dismissed):
I think this looks good. Looks like you have to resolve some merge issues so you will need another approval, but let me know when you think is ready. The most important thing with this PR is to get the study setup soon.

alextanlan (approved):

PR #608, opened 2021-06-04
DDP-6253 Add translations for Participation page (PanCan study)

alextanlan:


ddp-workspace/projects/ddp-pancan/src/assets/i18n/en.json
```diff
@@ -49,6 +49,42 @@
         "Text": "With your help, patients can be the driving force behind transformative cancer research."
       }
     },
+    "ParticipationPage": {
+      "MainTitle": "PARTICIPATION PAGE",
```
aglushko-broad: probably we need to replace it with "Here’s how to participate"? Plus I guess it might be just a Title instead of MainTitle

alextanlan: My bad, thanks for noticing.
Fixed at 5808da7943495b62d120cafd3ee958a74cadf165

aglushko-broad (approved):

PR #610, opened 2021-06-04
DDP-6296 fix subject panel for prism testboston

aglushko-broad:


ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard-redesigned.component.ts
```diff
@@ -11,8 +11,10 @@ import { map, take, filter } from 'rxjs/operators';
     template: `
         <main class="main">
             <section class="section">
-                <ddp-subject-panel *ngIf="!selectedUserGuid || (selectedUser$ | async) as selectedUser" [subject]="selectedUser">
-                </ddp-subject-panel>
+                <ng-container *ngIf="{selectedUser: selectedUser$ | async} as data">
```
MocanaAtBroad, line 14: I am approving but do have a question about the approach here (for my own education). What is the advantage of creating a `data` object in the ngIf instead of just taking whatever object `selectedUser$ | async` returns?

MocanaAtBroad (approved):

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard-redesigned.component.ts
aglushko-broad: the difference is that in the object approach we save the selectedUser$ stream result even if it's nullable and can use it futher. NgIf gets truthy value and can render the html under ng-container
If I'd just use <ng-container *ngIf="selectedUser$ | async"> and selectedUser is null then ngIf gets falsy value and we don't render html under container. but here we use ngIf not for condition but for generating variable for async stream (currently Angular doesn't have other mechanics for that)

aglushko-broad: the bug was related to that *ngIf="!selectedUserGuid || (selectedUser$ | async) as selectedUser" expression saved "true" to selectedUser variable when selectedUserGuid was false and that's not expected behavior, it should be null, so the mistake was made in the syntax

PR #613, opened 2021-06-07
DDP-6303 update header and add i18n for header and footer for PanCan

aglushko-broad:
- replaced toolkit header with custom header since we can't use it here in PanCan
- copied based styles for layout and hamburger menu header from Osteo
- used redesigned components and update i18n for them
- add i18n for header and footer

### DIFF
--- a/ddp-workspace/projects/ddp-pancan/src/variables.scss
+++ b/ddp-workspace/projects/ddp-pancan/src/variables.scss
@@ -1,0 +1,2 @@
+$header-height: 6rem;
+$footer-height: 14.94rem;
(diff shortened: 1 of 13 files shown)


ddp-workspace/projects/ddp-pancan/src/variables.scss
```diff
@@ -0,0 +1,2 @@
+$header-height: 6rem;
+$footer-height: 14.94rem;
```
alextanlan, line 2: #minor
Do we need such an accuracy there ?

alextanlan (approved):

ddp-workspace/projects/ddp-pancan/src/variables.scss
aglushko-broad: why not? any issues about that?

alextanlan: Not an issue, just a strange value, as for me.

aglushko-broad: I calculated it based on mockup sizes and keep 2 numbers after the dot. I hope it's more clear now.

PR #615, opened 2021-06-07
DDP-6294 rgp: no validate when next section in read-only mode

yufengwng:
There is an edge-case in RGP, where an older migrated user (perhaps from 2017) will not be able to navigate to next sections when survey is in read-only mode. Given how RGP is setup, there's a bit of mismatch of survey questions between Pepper and Gen2, e.g. there are required questions that older migration users might not have answered.

Currently, we run validations when user clicks "next" button, even in read-only mode. Therefore, we get into situation where an older migration user might not be able to view other sections in their survey since they might be missing answers to required questions.

This PR overrides the behavior so that RGP allows navigating without running validations in read-only mode. CC-ing Charlotte and Aleksey since they had worked on RGP frontend.

See DDP-6294 for more details.


MocanaAtBroad (approved):
Code looks good to me. If you are sure this is what you want to do, then 👍

charlottedye (approved):

PR #620, opened 2021-06-08
DDP-5580: Remove COVID banner

aleksey-kuznetsov:


charlottedye (approved):
Looks good to me, although at some point someone will need to do the same thing for the Auth0 pages

PR #625, opened 2021-06-11
A-T: Fix constructor arguments for DI

aleksey-kuznetsov:
For some reason a dependency injection error appeared, that's fixed now.


aleksey-kuznetsov:
And just for convenience I've added missing translations for login button on session expired page.

yufengwng (approved):

PR #630, opened 2021-06-14
DDP-6362 rgp: fix sorting and date handling of study messages

yufengwng:
Not 100% sure about this, but I think this will fix the sorting issue as well as the date off-by-one issue.

* For sorting, we should sort based on the original timestamp of the workflow status.
* For the date display, using the `new Date()` constructor is not good, so we do a bit more processing.


aleksey-kuznetsov (approved):
That should do the trick. Thank you!

PR #635, opened 2021-06-15
Minor updates to reflect capped enrollment and new team members

zyme:
Requests originally from B12-56 and TB12-81, but also via additional requests from tbos team in various meetings.

Screen Shot 2021-06-15 at 1 40 31 PM
Screen Shot 2021-06-15 at 1 40 25 PM
Screen Shot 2021-06-15 at 1 40 10 PM

### DIFF
--- a/ddp-workspace/projects/ddp-testboston/src/assets/i18n/en.json
+++ b/ddp-workspace/projects/ddp-testboston/src/assets/i18n/en.json
@@ -67,7 +67,7 @@
       "Title": "We want to help our communities and learn more about COVID-19 infection patterns in the Boston area.",
       "Text": "TestBoston is a COVID-19 study supported by Brigham and Women's Hospital and the Broad Institute of MIT and Harvard.",
       "Code": "I have an invitation code",
-      "NoCode": "We are currently fully enrolled.  Please check back in for future updates!"
+      "Full": "We are currently fully enrolled.  Please check back in for future updates!"
     },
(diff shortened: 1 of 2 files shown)


ddp-workspace/projects/ddp-testboston/src/assets/i18n/en.json
```diff
@@ -67,7 +67,7 @@
       "Title": "We want to help our communities and learn more about COVID-19 infection patterns in the Boston area.",
       "Text": "TestBoston is a COVID-19 study supported by Brigham and Women's Hospital and the Broad Institute of MIT and Harvard.",
       "Code": "I have an invitation code",
-      "NoCode": "I don’t have an invitation code and want to join the study"
+      "NoCode": "We are currently fully enrolled.  Please check back in for future updates!"
```
yufengwng: Might be better to introduce a new translation variable? What about other languages?

zyme: Ah, I see your point--if we leave `NoCode` as is, the translations will be **wrong** (as in the meaning is totally different).  I've updated this so it uses a new variable, so that at least other languages will see the untranslated value instead of the text linking them to MGB.

MocanaAtBroad (approved):
It does what you want it to do, but consider making the the "fully enrolled" message more salient. Maybe inside a yellow box? Not very noticeable on first view.

PR #638, opened 2021-06-16
DDP-6349: Add new members to about us page

aleksey-kuznetsov:


charlottedye (changes requested):
It looks like the names should be in alphabetical order (both from looking at the spreadsheet and from looking at how the names are currently arranged).

aleksey-kuznetsov:
> It looks like the names should be in alphabetical order (both from looking at the spreadsheet and from looking at how the names are currently arranged).

Correct! Thanks for that.

yufengwng:
What's the status of this?

yufengwng (approved):
Looks good on my end. Let's merge.

PR #640, opened 2021-06-16
rgp: improve study messages handling

yufengwng:
For DDP-6362:
* Added conditions to implement the `Don't show when` column from spreadsheet.
* Tweaked conditions so we hide messages that depend on dates and the date is missing.

For DDP-6400:
* Added some CSS to space out the "show more" container.

### DIFF
--- a/ddp-workspace/projects/ddp-rgp/src/app/components/study-messages/study-messages.component.scss
+++ b/ddp-workspace/projects/ddp-rgp/src/app/components/study-messages/study-messages.component.scss
@@ -23,6 +23,10 @@
   }
 }
 
+.show-more-container {
+  margin-top: 10px;
+}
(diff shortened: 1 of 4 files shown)


ddp-workspace/projects/ddp-rgp/src/app/components/study-messages/study-messages.component.scss
```diff
@@ -23,6 +23,10 @@
   }
 }
 
+.show-more-container {
+  margin-top: 10px;
```
yufengwng, line 27: This is based on padding for `p` tags, which in RGP looks like it uses hardcoded pixels.

charlottedye (approved):
Looks good!

PR #646, opened 2021-06-17
DDP-6419 fix handling of invalid numeric answer

yufengwng:
See DDP-6419 for a bit more details. The old code emits a `null` when the input is invalid (out-of-range). This causes a PATCH API request with `null` as the answer value, and causes the `block.answer` to be `null` which means the validation rule doesn't have anything to validate against.

You can reproduce the bug in `basil`. Configs for `basil` are here, along with the RareX config changes that surfaced the bug: https://github.com/broadinstitute/ddp-study-server/pull/818.

Screen Shot 2021-06-17 at 11 50 45 AM

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activityNumericAnswer.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activityNumericAnswer.component.ts
@@ -53,7 +53,11 @@ export class ActivityNumericAnswer implements OnInit, OnChanges {
             this.numericField.setValue('');
             this.emitAnswer(null);
         } else if (this.numericField.invalid) {
-            this.emitAnswer(null);
+            // If it's invalid, it most likely mean it's out-of-range, which
+            // also implies we have an INT_RANGE validation rule. We should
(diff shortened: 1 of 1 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activityNumericAnswer.component.ts
```diff
@@ -53,7 +53,11 @@ export class ActivityNumericAnswer implements OnInit, OnChanges {
             this.numericField.setValue('');
             this.emitAnswer(null);
         } else if (this.numericField.invalid) {
-            this.emitAnswer(null);
```
yufengwng, line 56: I've tracked this line down to [here](https://github.com/broadinstitute/ddp/pull/1103#issuecomment-545430633) and [here](https://github.com/broadinstitute/ddp/pull/1103/commits/6ab7f8f5c37931410316cace0259c1b764fc71dc). It appears that there might have been a reason why we're emitting `null` here, but I'm not too sure if it still applies anymore.

MocanaAtBroad (approved):

PR #661, opened 2021-06-22
DDP-6260 Create Participation page for PANCAN

alextanlan:
Created Participation page for Pancan

### DIFF
--- a/ddp-workspace/projects/ddp-pancan/src/app/components/participation/participation.component.html
+++ b/ddp-workspace/projects/ddp-pancan/src/app/components/participation/participation.component.html
@@ -1,27 +1,23 @@
 <main class="participation">
     <h2 class="title" translate>App.ParticipationPage.Title</h2>
-    <section *ngFor="let stepNumber of '12345'.split(''); let last=last;" class="step">
+    <section *ngFor="let step of 'App.ParticipationPage.Steps' | translate; let i = index;" class="step">
         <div class="step__data">
-            <span class="step__label"
-                  [innerText]="translationStepData(stepNumber, '.Step') | translate">
-            </span>
-            <span class="step__time" *ngIf="!last"
-                  [innerText]="translationStepData(stepNumber, '.Time') | translate">
-            </span>
-            <h2 class="step__title" [innerText]="translationStepData(stepNumber, '.Title') | translate"></h2>
-            <p class="step__text" [innerText]="translationStepData(stepNumber, '.Text') | translate"></p>
+            <span class="step__label">{{step.Step}}</span>
+            <span class="step__time" *ngIf="step.Time">{{step.Time}}</span>
+            <h2 class="step__title" >{{step.Title}}</h2>
+            <p class="step__text">{{step.Text}}</p>
 
-            <a mat-stroked-button *ngIf="!last"
+            <a mat-stroked-button *ngIf="step.Button"
                target="_blank"
-               [href]="stepsData[stepNumber - 1].link"
+               [href]="stepsData[i].link"
                class="step__btn">
-                {{translationStepData(stepNumber, '.Button') | translate}}
+                {{step.Button}}
             </a>
         </div>
         <div class="step__image">
             <img class="image"
-                 [src]="'../../../assets/images/step' + stepNumber + '_big.png'"
-                 [alt]="translationStepData(stepNumber, '.ImageAlt') | translate">
+                 [src]="'../../../assets/images/step' + (i+1) + '_big.png'"
+                 [alt]="step.ImageAlt">
         </div>
     </section>
 </main>
--- a/ddp-workspace/projects/ddp-pancan/src/app/components/participation/participation.component.html
+++ b/ddp-workspace/projects/ddp-pancan/src/app/components/participation/participation.component.html
@@ -1,0 +1,28 @@
+<main class="participation">
+    <h2 class="title" translate>App.ParticipationPage.Title</h2>
+    <section *ngFor="let stepNumber of '12345'.split(''); let last=last;" class="step">
+        <div class="step__data">
+            <span class="step__label"
+                  [innerText]="translationStepData(stepNumber, '.Step') | translate">
+            </span>
+            <span class="step__time" *ngIf="!last"
+                  [innerText]="translationStepData(stepNumber, '.Time') | translate">
+            </span>
+            <h2 class="step__title" [innerText]="translationStepData(stepNumber, '.Title') | translate"></h2>
+            <p class="step__text" [innerText]="translationStepData(stepNumber, '.Text') | translate"></p>
+
+            <a mat-stroked-button *ngIf="!last"
+               target="_blank"
+               [href]="stepsData[stepNumber - 1].link"
+               class="step__btn">
+                {{translationStepData(stepNumber, '.Button') | translate}}
+            </a>
+        </div>
+        <div class="step__image">
+            <img class="image"
+                 [src]="'../../../assets/images/step' + stepNumber + '_big.png'"
+                 width="340" height="300"
+                 [alt]="translationStepData(stepNumber, '.ImageAlt') | translate">
+        </div>
+    </section>
+</main>
--- a/ddp-workspace/projects/ddp-pancan/src/app/components/participation/participation.component.html
+++ b/ddp-workspace/projects/ddp-pancan/src/app/components/participation/participation.component.html
@@ -7,9 +7,9 @@
             <h2 class="step__title" >{{step.Title}}</h2>
             <p class="step__text">{{step.Text}}</p>
 
-            <a mat-stroked-button *ngIf="step.Button"
+            <a mat-stroked-button *ngIf="step.Button && stepsHref[i]"
                target="_blank"
-               [href]="stepsData[i].link"
+               [href]="stepsHref[i]"
                class="step__btn">
                 {{step.Button}}
             </a>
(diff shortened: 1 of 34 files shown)


ddp-workspace/projects/ddp-pancan/src/app/components/participation/participation.component.html
```diff
@@ -0,0 +1,27 @@
+<main class="participation">
+    <h2 class="title" translate>App.ParticipationPage.Title</h2>
+    <section *ngFor="let stepNumber of '12345'.split(''); let last=last;" class="step">
+        <div class="step__data">
+            <span class="step__label"
+                  [innerText]="translationStepData(stepNumber, '.Step') | translate">
```
aglushko-broad: what do you think if we change the structure for ParticipationPage in json file? instead of the object, it might be an array. So we can replace the construction '12345'.split('') with 'App.ParticipationPage.Steps' | translate. And in general, follow the approach we chose in the faq-section component.

ddp-workspace/projects/ddp-pancan/src/app/components/participation/participation.component.scss
```diff
@@ -0,0 +1,83 @@
...
+            padding-bottom: 3rem;
+            flex-direction: row;
+        }
+
+        &__data {
+            max-width: 700px;
+        }
+
+        &__label {
+            @extend .participate-step-label;
```
aglushko-broad: #minor what do you think about attaching the class to the html element instead of using extend? after conversion scss to css we get the next code
.participate-step-label,
.participation step__label {
 ... // styles for .participate-step-label
}
 which a little bit harder to debug when we want to disable some styles only for one element. On the other hand, assigning the simple class to the html element allows us to enable/disable this class for the certain element in the devtools for example.

aglushko-broad: I'd recommend using extend in case when just assigning the class doesn't work because of css specificity

ddp-workspace/projects/ddp-pancan/src/app/components/participation/participation.component.html
```diff
@@ -0,0 +1,27 @@
...
+            </span>
+            <span class="step__time" *ngIf="!last"
```
aglushko-broad: I'm not sure it's the best way to identify if we should display time or not. It's better to have a data structure array and check if the item contains the time. In case if in the future something will be changed and time will be removed from other items it will be harder to fix

aglushko-broad: same for button

ddp-workspace/projects/ddp-pancan/src/app/components/participation/participation.component.scss
alextanlan: I used `@extend` as to re-use common styles for different elements there.
I believe the best choice would be using `scss mixin` instead of `extend`
See c5fad5eb38992a6589fbfe841ece62c13fb0c1fd

ddp-workspace/projects/ddp-pancan/src/styles/common.scss
```diff
@@ -1,40 +1,57 @@
...
 
-.static-page-paragraph-text {
+@mixin static-page-paragraph-text {
     opacity: 0.8;
     font-size: 1.5rem;
     letter-spacing: -0.1px;
     line-height: 34px;
 }
 
-.btn-common {
```
aglushko-broad, line 18: I used btn-common, btn-outline and btn-header classes in html. I suppose it's not working now.
mixin is not always the best choice. Mixins are very good if styles are configurable based on the passed value. But for regular cases, simple classes are a better choice. It's easier to search and debug and use rght in html w/o creating additional class

ddp-workspace/projects/ddp-pancan/src/app/components/participation/participation.component.html
alextanlan: Agree, it is more generic way.
Fixed at 9d6ef0c989ea9767fd5ef6d023506ef7eb753b1d

ddp-workspace/projects/ddp-pancan/src/app/components/participation/participation.component.html
alextanlan: Fixed at 9d6ef0c989ea9767fd5ef6d023506ef7eb753b1d

ddp-workspace/projects/ddp-pancan/src/app/components/participation/participation.component.html
```diff
@@ -0,0 +1,23 @@
+<main class="participation">
+    <h2 class="title" translate>App.ParticipationPage.Title</h2>
+    <section *ngFor="let step of 'App.ParticipationPage.Steps' | translate; let i = index;" class="step">
+        <div class="step__data">
+            <span class="step__label">{{step.Step}}</span>
+            <span class="step__time" *ngIf="step.Time">{{step.Time}}</span>
+            <h2 class="step__title" >{{step.Title}}</h2>
+            <p class="step__text">{{step.Text}}</p>
+
+            <a mat-stroked-button *ngIf="step.Button"
```
aglushko-broad: #minor  *ngIf="step.Button && stepsData[i].link"

ddp-workspace/projects/ddp-pancan/src/styles/common.scss
alextanlan: I was not able to find places where 
>  used btn-common, btn-outline and btn-header classes in html.

Seems like the classes are used in css only.
And they should be working currently.

Using `mixins` is just more convenient way to re-use common styles/wrap several style into one (then `@extend`), imho.
They can be used without any passed value as well. 

Anyway, would you like me to get those styles back as class selectors ?

ddp-workspace/projects/ddp-pancan/src/app/components/participation/participation.component.html
alextanlan: Fixed at eba10a42aa99709c0691ef48a2798fe93f9ba713

ddp-workspace/projects/ddp-pancan/src/styles/common.scss
aglushko-broad: oh, I added the comment but didn't open the PR where I use it
https://github.com/broadinstitute/ddp-angular/pull/662/files
here

yes please! ability to use simple style classes right in html is very important, we need to keep it.

aglushko-broad: like I said, it's easier to debug, plus you can add/remove classes conditionally in html based on some logic. We can't do it with mixins

alextanlan: Reverted at 276b15b92785892acc656f8ccae03443170efcb9

ddp-workspace/projects/ddp-pancan/src/app/components/participation/participation.component.scss
alextanlan: Finally reverted. (see https://github.com/broadinstitute/ddp-angular/pull/661#discussion_r657472528)

ddp-workspace/projects/ddp-pancan/src/app/components/welcome/welcome.component.scss
```diff
@@ -108,15 +108,11 @@ section {
...
-                letter-spacing: -0.07px;
-                line-height: 24px;
-                margin-right: 8px;
+                @extend .participate-step-label;
             }
 
             &__time {
-                letter-spacing: -0.07px;
-                line-height: 24px;
+                @extend .participate-step-time;
```
aglushko-broad, line 115: do we need this line? I see we already assigned the class in html

ddp-workspace/projects/ddp-pancan/src/app/components/participation/participation.component.scss
```diff
@@ -0,0 +1,76 @@
+@import '../../../styles/common.scss';
+
+.participation {
+    text-align: left;
+
+    .title {
+        @extend .static-page-title;
```
aglushko-broad: I can see that we use static-page-title class in html for step__title . Can we do it for this title element as well? (same for other extends)

ddp-workspace/projects/ddp-pancan/src/app/components/welcome/welcome.component.scss
```diff
@@ -108,15 +108,11 @@ section {
...
             &__title {
                 display: inline;
-                font-family: $primary-font-medium;
-                letter-spacing: -0.07px;
-                line-height: 24px;
-                margin-right: 8px;
+                @extend .participate-step-label;
             }
 
             &__time {
```
aglushko-broad: looks like we can remove class step__time and use .participate-step-time class on the element

ddp-workspace/projects/ddp-pancan/src/app/components/participation/participation.component.scss
alextanlan: Fixed at 316abf2f0e57626829a3c6149f4cf9089db50904

ddp-workspace/projects/ddp-pancan/src/app/components/welcome/welcome.component.scss
alextanlan: Fixed at 316abf2f0e57626829a3c6149f4cf9089db50904

aglushko-broad (approved):

PR #672, opened 2021-06-25
Circadia: responsive layout & minor fixes

aleksey-kuznetsov:


ddp-workspace/projects/ddp-circadia/src/app/components/activity/activity.component.ts
```diff
@@ -12,4 +14,31 @@ import {
...
+
+    this.initFixedStepper();
+  }
+
+  private initFixedStepper(): void {
+    const isLoaded$ = this.getIsLoaded$();
+
+    const sub = isLoaded$.pipe(filter(isLoaded => isLoaded)).subscribe(() => {
+      if (this.model.statusCode !== ActivityStatusCodes.COMPLETE) {
+        // @ts-ignore
```
aglushko-broad: please add a comment which explains which error is being suppressed.

aleksey-kuznetsov: @aglushko-broad cannot say for sure is that a bug or not.
[This](https://github.com/broadinstitute/ddp-angular/blob/eaae88fad3b714d54feee259074c002fd05116f6/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts#L504) exact line breaks the sectioning when you jump from one activity with sections to another with sections. Removing this line fixes this behavior so that on each new activity the stepper is initialized again.

This line was introduced under "Fix small problems" commit so I don't really know whether removing this line will break anything in other studies.

aleksey-kuznetsov: As an example, if you submit an activity that has 3 sections, `currentSectionIndex` will still store `2`. Once new activity after the submission was rendered `currentSectionIndex` will not be reset to 0.

aglushko-broad: My opinion here - it shouldn't break anything if we remove take(1), we subcribes on subject observable, doesn't look like removing take(1) will call additional server requests, also check this line
https://github.com/broadinstitute/ddp-angular/blob/develop/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-redesigned.component.ts#L46
no take(1) and should work with no issues.

aleksey-kuznetsov: @aglushko-broad alright, I'll remove that line in a separate PR.

ddp-workspace/projects/ddp-circadia/src/app/components/pages/home/home.component.ts
```diff
@@ -58,7 +58,11 @@ export class HomeComponent implements OnInit, AfterViewInit, OnDestroy {
       return;
     }
 
-    const yOffset = el.offsetTop - this.headerHeight;
+    const currentFontSize = +getComputedStyle(
+      document.documentElement,
+    ).fontSize.slice(0, 1);
```
aglushko-broad: that looks a little bit tricky. What's the minimal and maximum font-size might be?

aleksey-kuznetsov: All measurements in the application are done using `rem` unit. On desktop `1rem` is equal to `10px` (by setting `html` element `font-size` to `62.5%`). On smaller screens we reduce it down to `9px` (`font-size: 56.25%`).

Since `font-size` is dynamic, we have to get current value for it which this line is responsible for.

aleksey-kuznetsov: This way just by changing `font-size` on `html` we can affect on whole layout without changing each element's styles.

aglushko-broad: so when font-size is 10px the currentFontSize  will be 1, but when 9px the currentFontSize will be 9. And 
const yOffset = el.offsetTop - 10; vs const yOffset = el.offsetTop - 90;
is that expected behavior?

aleksey-kuznetsov: Oh, correct, that is my mistake.

Thank you! I'll fix that.

aglushko-broad:
don't forget about screenshots next time :)

aglushko-broad (approved):

PR #674, opened 2021-06-29
DDP-6487 refactor handling of visibility updates

yufengwng:
See DDP-6487. Not too sure if this is right, but here's my guess of what's causing the issue:

* When embedded activity component is visible and you click submit, it fails validation.
* `embeddedComponentsValidStatusChanged` observable resolves to `false` (means invalid).
* `isAllFormContentValid` resolves to `false` and we're prevented from submitting.
* You answer question and hide the embedded activity component via block visibility response from server.
* `BaseActivityComponent.updateVisibility()` runs and sets the embedded activity block to hidden, and causes Angular to offload the block from rendering.
* Which means `ActivitySectionComponent` has no chance of updating `embeddedComponentsValidationStatus`.
* Which means `isAllFormContentValid` will never be `true` and we can't ever submit the form...


alextanlan:
Please, do **not** merge it!  
I am trying to get to the issue root cause.

PR #682, opened 2021-07-01
Feature/ddp 5912 session expired dialog

PavelYakupov:


aleksey-kuznetsov (changes requested):
@PavelYakupov you have to also add path `Route.SessionExpired` to `ConfigurationService` under `sessionExpiredUrl` field.
Also, don't forget to add this url the to list of allowed callback urls in Auth0, this route is used as a `returnTo` path when logging out user due to expired session.

aleksey-kuznetsov (approved):

PR #690, opened 2021-07-05
FIX: Fix a console error (for all studies)

alextanlan:
Fixed a console error that appeared in many studies
`   TypeError: Cannot read property 'participantGuid' of null`

screenshot


aglushko-broad:
thanks!

aglushko-broad (approved):

PR #691, opened 2021-07-06
DDP-6487: Fix validations which take hidden questions into account

alextanlan:
The bug worklow is [here (DDP-6487)](https://broadinstitute.atlassian.net/browse/DDP-6487) 
The ticket is also related to [DDP-6286](https://broadinstitute.atlassian.net/browse/DDP-6286).

Implemented: 
- removed `updateVisibility` method from baseActivity component
- used `updateVisibility` method in activitySection component only
- fixed incorrect `embeddedValidationStatus` working (in activitySection.component)
- applied a workaround for reset embeddedValidationStatus of hidden embedded activities  (see [the comment)](https://github.com/broadinstitute/ddp-angular/pull/691#discussion_r664725537)

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityBlock/activityBlock.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityBlock/activityBlock.component.ts
@@ -33,7 +33,7 @@ export class ActivityBlockComponent implements OnInit, OnDestroy {
     @Input() validationRequested: boolean;
     @Input() studyGuid: string;
     @Input() parentActivityInstanceGuid: string;
-    @Output() validStatusChanged = new EventEmitter<boolean>();
+    @Output() validStatusChanged = new EventEmitter<{id: string; value: boolean}>();
     @Output() embeddedComponentBusy = new EventEmitter<boolean>();
     @ViewChildren(ModalActivityBlockComponent) private modalActivities: QueryList<ModalActivityBlockComponent>;
     isModal: boolean;
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityBlock/activityBlock.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityBlock/activityBlock.component.ts
@@ -85,7 +85,9 @@ export class ActivityBlockComponent implements OnInit, OnDestroy {
     }
 
     private openCreatedInstanceDialog(instanceGuid: string): void {
-        this.modalActivities.find(activity => activity.instance.instanceGuid === instanceGuid)
-            .openEditDialog();
+        const modalActivity = this.modalActivities.find(activity => activity.instance.instanceGuid === instanceGuid);
+        if (modalActivity) {
+            modalActivity.openEditDialog();
+        }
     }
 }
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/embeddedActivityBlock/embeddedActivityBlock.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/embeddedActivityBlock/embeddedActivityBlock.component.ts
@@ -35,7 +35,7 @@ export class EmbeddedActivityBlockComponent implements OnInit {
     @Input() validationRequested: boolean;
     @Input() studyGuid: string;
     @Output() componentBusy = new EventEmitter<boolean>();
-    @Output() validStatusChanged = new EventEmitter<boolean>();
+    @Output() validStatusChanged = new EventEmitter<{id: string; value: boolean}>();
     @Output() deletedActivity = new EventEmitter<string>();
     @ViewChild('delete_button', {read: ElementRef}) private deleteButtonRef: ElementRef;
 
@@ -112,7 +112,7 @@ export class EmbeddedActivityBlockComponent implements OnInit {
             tap(() => this.componentBusy.emit(true)),
             tap(() => {
                 activityValidStatus = this.activity.validate();
-                this.validStatusChanged.emit(activityValidStatus);
+                this.validStatusChanged.emit({id: this.instance.instanceGuid, value: activityValidStatus});
                 if (!activityValidStatus) {
                     this.componentBusy.emit(false);
                 }
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.html
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.html
@@ -142,10 +142,10 @@
                     </div>
                     <div class="col-lg-12 col-md-12 col-sm-12 col-xs-12">
                         <div *ngIf="displayGlobalError$ | async" class="ErrorMessage">
-                            Q11<span translate>SDK.ValidateError</span>
+                            <span translate>SDK.ValidateError</span>
                         </div>
                         <div *ngIf="communicationErrorOccurred" class="ErrorMessage">
-                            Q12<span translate>SDK.CommunicationError</span>
+                            <span translate>SDK.CommunicationError</span>
                         </div>
                     </div>
                 </div>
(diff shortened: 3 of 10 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.html
```diff
@@ -0,0 +1,155 @@
+<div class="Wrapper">
```
alextanlan, line 1: Nothing was changed in the file.
Just extracted the template into a separate file.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activitySection.component.html
```diff
@@ -0,0 +1,79 @@
...
+                                   [activityGuid]="activityGuid"
+                                   (componentBusy)="componentBusy.emit($event)">
+            </ddp-activity-question>
+        </div>
+        <div *ngIf="isInstitution(block)">
+            <ddp-institutions-form [block]="block"
+                                   [studyGuid]="studyGuid"
+                                   [readonly]="readonly"
+                                   [validationRequested]="validationRequested"
+                                   (validStatusChanged)="updateValidationStatusInSection(block.id, $event)"
```
alextanlan: Renamed `updateValidationStatus` => `updateValidationStatusInSection`
Passing `id` as a key (instead of hard-coded numbers) - because previously the validStatus were overrided many times (the validStatus map had only 3 keys (`0`, `1`, `2`). 

The same for lines 51, 75.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activitySection.component.ts
```diff
@@ -140,16 +62,22 @@ export class ActivitySectionComponent implements OnInit, OnDestroy {
                     if (block.shown !== element.shown) {
                         block.shown = element.shown;
                         blockVisibilityChanged = true;
+
+                        if (block.blockType === BlockType.Activity && !block.shown) {
+                            this.updateValidationForHiddenEmbeddedActivity(block as ActivityActivityBlock);
```
alextanlan, line 75: A workaround as to reset validation statuses for hidden embedded activities.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/embeddedActivityBlock/embeddedActivityBlock.component.ts
```diff
@@ -112,7 +112,7 @@ export class EmbeddedActivityBlockComponent implements OnInit {
             tap(() => this.componentBusy.emit(true)),
             tap(() => {
                 activityValidStatus = this.activity.validate();
-                this.validStatusChanged.emit(activityValidStatus);
+                this.validStatusChanged.emit({id: this.instance.instanceGuid, value: activityValidStatus});
```
yufengwng, line 115: Block guids and instance guids are generated separately, meaning there is a possibility that they can have the same ids. I doubt we'll ever encounter a conflict, so I'll be okay going along with this, but we might want to make a note of it.

If the ids aren't used for anything other than mapping the value, then perhaps a simple scheme here is to prepend a unique prefix? E.g. `b-` for blocks and `i-` for instances?

yufengwng (commented):
Having a size 3 array for tracking state of embedded validations did seem odd to me when I looked at the code, so thanks for refactoring that. I agree that using a map here is a better way to go. Your workaround of resetting state for embedded activities by adding a special check makes sense to me as well.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activitySection.component.html
```diff
@@ -0,0 +1,79 @@
...
+                         [activityGuid]="activityGuid">
+        </ddp-group-block>
+    </div>
+    <div *ngIf="isActivityBlock(block)">
+        <ddp-activity-block [block]="block"
+                            [readonly]="readonly"
+                            [validationRequested]="validationRequested"
+                            [studyGuid]="studyGuid"
+                            [parentActivityInstanceGuid]="activityGuid"
+                            (validStatusChanged)="updateValidationStatusInSection($event.id, $event.value)"
```
aglushko-broad: can it be block.id instead of $event.id?

alextanlan: No.
That activityBlock has a list of embedded activities (childInstances) inside of itself with different `instanceId`-s.
They can emit their own `validationStatus`.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/embeddedActivityBlock/embeddedActivityBlock.component.ts
alextanlan: Added at e51002dec7c5a6c2e7f356d065ca3a9dbf0fa034

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activitySection.component.html
aglushko-broad: but we didn't care about it before and considered only one returned result for the whole block. Was it a wrong approach?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activitySection.component.ts
aglushko-broad: probably makes sence to rename the method from updateVisibility to updateVisibilityAndValidation

alextanlan: Fixed

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activitySection.component.html
alextanlan: I feel like it was an incorrect decision. 
We missed the fact that `ActivityBlock` can contain several `embeddedActivities` (like 'Siblings'/ 'Children').
Previously we used one key for `activityBlock`. However a value (validation status) was overriden from one `embeddedActivity` validation status to another.

aglushko-broad (approved):

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/embeddedActivityBlock/embeddedActivityBlock.component.ts
yufengwng: Do we need `idPrefix.instance` here so things match up?

alextanlan: It is added at the top level in `activitySection` component template (activitySection.component.html) for consistency.

`EmbeddedActivityBlockComponent` (**id**) =>
    `ActivityBlockComponent` (**id**) =>
        `ActivitySectionComponent` (**idPrefix.instance + $event.id**) =>

yufengwng: Ah, nice! I see it now.

yufengwng (approved):
I tried this locally for the RGP study and it seems to work well, so I think it's ready to go!

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityBlock/activityBlock.component.ts
```diff
@@ -85,7 +85,9 @@ export class ActivityBlockComponent implements OnInit, OnDestroy {
     }
 
     private openCreatedInstanceDialog(instanceGuid: string): void {
-        this.modalActivities.find(activity => activity.instance.instanceGuid === instanceGuid)
-            .openEditDialog();
+        const modalActivity = this.modalActivities.find(activity => activity.instance.instanceGuid === instanceGuid);
```
MocanaAtBroad, line 88: I think you can do something like:
this.modalActivities.find(activity => activity.instance.instanceGuid === instanceGuid)?.openEditDialog();

MocanaAtBroad (commented):
Looks good! Super minor comment.

yufengwng:
Merging now so we can get this into the release.

PR #695, opened 2021-07-08
pancan: initial activity routes

yufengwng:
Add some initial routes so we can start seeing some activities.


aglushko-broad:
thanks!

aglushko-broad (approved):

PR #701, opened 2021-07-12
Rare-X / RefId 19/20b

aleksey-kuznetsov:
- Added help message on dashboard page
- Default to open participant activities row when there is only 1 participant in the list


yufengwng (dismissed):
I see we've added a quick help tip right above the participant list, and we're automatically expanding the single row whenever we load data or right after deleting participant. LGTM!

yufengwng (approved):

PR #702, opened 2021-07-12
DDP-6548 picklist autocomplete reset + display list of selections be default

aglushko-broad:
- allow the user to reset the selected value
- show suggestions lit when user first time focuses on input
- fix disabled style for submit button in basil


MocanaAtBroad (approved):
Looks good! Let put this baby to use!

PR #703, opened 2021-07-13
Rare-X / RefId-97

aleksey-kuznetsov:
- Display currently selected participant name
- Open previous participant activities upon return to dashboard

### DIFF
--- a/ddp-workspace/projects/ddp-rarex/src/app/components/top-bar/top-bar.component.ts
+++ b/ddp-workspace/projects/ddp-rarex/src/app/components/top-bar/top-bar.component.ts
@@ -1,10 +1,17 @@
-import { Component, OnInit } from '@angular/core';
-
-import { SessionMementoService, UserProfileServiceAgent } from 'ddp-sdk';
+import { Component, OnDestroy, OnInit } from '@angular/core';
+import { NavigationEnd, Router } from '@angular/router';
 import { filter, map, switchMap, take } from 'rxjs/operators';
+
+import {
+  CompositeDisposable,
+  SessionMementoService,
+  UserProfileServiceAgent,
+} from 'ddp-sdk';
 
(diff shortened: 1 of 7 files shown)


ddp-workspace/projects/ddp-rarex/src/app/services/participant.service.ts
```diff
@@ -0,0 +1,58 @@
...
+  private prevParticipantGuid$ = new BehaviorSubject<string | null>(null);
+  private currentParticipantProfile$ = new BehaviorSubject<UserProfile | null>(
+    null,
+  );
+  private sessionSub = this.sessionService.sessionObservable
+    .pipe(
+      takeUntil(this.destroyed),
+      filter(session => !!session),
+      switchMap(session =>
+        this.userProfileService.profile.pipe(
```
yufengwng, line 26: Does this mean we only fetch/store the profile when we have a session (e.g. logged in), so this is effectively static? For RareX, the name gets populated in the backend after they submit the consent. Does this handle "refreshing" the profile to fetch the new name?

ddp-workspace/projects/ddp-rarex/src/app/components/top-bar/top-bar.component.ts
```diff
@@ -65,4 +94,14 @@ export class TopBarComponent implements OnInit {
         this.profile = profile;
       });
   }
+
+  private initRouteListener(): void {
+    const routeSub = this.router.events
+      .pipe(filter(e => e instanceof NavigationEnd))
+      .subscribe((e: NavigationEnd) => {
+        this.isParticipantNameShown = e.url === `/${RoutePaths.Survey}`;
```
yufengwng, line 102: Does this mean we're only showing name on when viewing the surveys? What about the dashboard?

PR #709, opened 2021-07-14
DDP-6563 Pancan: lowercase fonts extension

aglushko-broad:
fonts don't work on dev. I'm assuming it might be related to the fonts extension. Can't be 100 sure, but change is small enough to try it out


alextanlan:
Let's try.

alextanlan (approved):

PR #711, opened 2021-07-15
DDP-5915 Added favicon

PavelYakupov:


aleksey-kuznetsov (changes requested):
Take a look at `ddp-workspace/projects/ddp-sdk/src/lib/services/renewSessionNotifier.service.ts` service and how it sets favicons dynamically. Unfortunately, the name is hardcoded so you have to change the name of your favicon files and their paths (favicons should be accessible by `<root>/favicon.ico` path).

aleksey-kuznetsov (approved):

PR #719, opened 2021-07-16
Ddp 6472 participants dashboard

aglushko-broad:
image


ddp-workspace/projects/ddp-pancan/src/assets/i18n/es.json
```diff
@@ -369,7 +369,13 @@
...
         },
         "Dashboard": {
-            "Title": "Tablero de Progreso"
+            "Title": "Tablero de Progreso",
+            "ParticipantsTitle": "[ES]Participant Dashboard",
+            "AddParticipant": "[ES]Add Participant",
+            "ShowPanel": "[ES]Show",
+            "HidePanel": "[ES]Hide",
+            "UserLabel": "Usted",
+            "ChildLabel": "[ES]Your child"
```
alextanlan: #optional
`"Su hijo(a)" ` 
(picked it up from `ddp-study-builder` email template in Spanish)

aglushko-broad: done!

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard-redesigned.component.ts
```diff
@@ -51,37 +80,118 @@ import { map, take, filter } from 'rxjs/operators';
...
+    }
+
+    private getDashboardParticipants(): Observable<DashboardParticipant[]> {
+        return combineLatest([
+            this.getOperatorActivities()
+                .pipe(
+                    tap((userActivities) => {
+                        this.operatorActivities = userActivities;
+                    }),
+                    switchMap((userActivities) => userActivities?.length ? this.userProfile.profile : of(null))),
```
alextanlan: suggest extracting lines 171-176 as a constant in order to increase the code readability.

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard-redesigned.component.ts
```diff
@@ -51,37 +80,118 @@ import { map, take, filter } from 'rxjs/operators';
...
+                label: (participant.userProfile.firstName || participant.userProfile.lastName)
+                    ? `${participant.userProfile.firstName} ${participant.userProfile.lastName}`
+                    : `${this.translate.instant('Toolkit.Dashboard.ChildLabel')}${i > 0 ? ' #' + (i + 1) : ''}`,
+            }));
+            if (operatorParticipant) {
+                result.unshift({
+                    userGuid: this.session.session.userGuid,
+                    label: (operatorParticipant.profile.firstName || operatorParticipant.profile.lastName)
+                        ? `${operatorParticipant.profile.firstName} ${operatorParticipant.profile.lastName}`
+                        : this.translate.instant('Toolkit.Dashboard.UserLabel'),
```
alextanlan, line 184: Code blocks `lines 181-183` and `lines 188-190` (labels definitions) seems pretty similar. 
Does it make sense to extract them into a separate method ?

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard.component.spec.ts
```diff
@@ -64,4 +76,41 @@ describe('DashboardComponent', () => {
...
+        expect(sessionSpy.setParticipant).not.toHaveBeenCalled();
+    });
+
+    it('should set participant in session before action navigation for participant dashboard', () => {
+        toolkitConfigMock.useParticipantDashboard = true;
+        component.ngOnInit();
+        fixture.detectChanges();
+
+        const participantGuid = 'userGuid567';
+        component.navigate('1', participantGuid);
```
alextanlan, line 125: Would be nice to have a named constant for '1' value. (same for line 112)

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard-redesigned.component.ts
aglushko-broad: that's similar, but has enough differences, I'd prefer to not combine these 2 together

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard.component.ts
```diff
@@ -59,13 +59,18 @@ export class DashboardComponent implements OnInit, OnDestroy {
     private anchor: Subscription = new Subscription();
 
     constructor(
-        private router: Router,
+        protected router: Router,
```
alextanlan, line 71: Just wondering why you need `protected` here (and in line 65) ?

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard.component.spec.ts
aglushko-broad: why? this number is totally random and might be any other string

alextanlan: It seems like a `magic` value.

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard.component.ts
aglushko-broad: because we start using router and session in the DashboardRedesignedComponent. I think it's better than service instance renaming

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard.component.spec.ts
aglushko-broad: it's a test, it's ok using random string like '123', we don't check anywhere that this number was passed correclty means can pass anything

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard-redesigned.component.ts
aglushko-broad: good idea, done!

alextanlan (dismissed):

ddp-workspace/projects/toolkit/src/lib/services/toolkitConfiguration.service.ts
```diff
@@ -82,4 +82,6 @@ export class ToolkitConfigurationService {
 
     cookies: Cookies | null;
     usePrionPrivacyPolicyTemplate: boolean;
+    useParticipantDashboard: boolean;
+    addParticipantUrl: string;
```
MocanaAtBroad, line 88: Could you add a comment describing what these are? (yes: we should do it for all)

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard-redesigned.component.ts
```diff
@@ -51,37 +80,119 @@ import { map, take, filter } from 'rxjs/operators';
...
+                                            {{'Toolkit.Dashboard.HidePanel' | translate}}
+                                            <mat-icon>expand_less</mat-icon>
+                                        </ng-container>
+                                        <ng-template #show>
+                                            {{'Toolkit.Dashboard.ShowPanel' | translate}}
+                                            <mat-icon>expand_more</mat-icon>
+                                        </ng-template>
+                                    </mat-panel-description>
+                                </mat-expansion-panel-header>
+                                <ng-template matExpansionPanelContent>
```
MocanaAtBroad, line 105: Works OK without prefix #? (Not seeing it referenced anywhere so maybe does not matter)?

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard-redesigned.component.ts
```diff
@@ -51,37 +80,119 @@ import { map, take, filter } from 'rxjs/operators';
...
+            }),
+            switchMap((userActivities) => userActivities?.length ? this.userProfile.profile : of(null)));
+
+        return combineLatest([
+            operatorParticipant$,
+            this.governedParticipantsAgent.getGovernedStudyParticipants(this.config.studyGuid)
+        ]).pipe(map(([operatorParticipant, participants]) => {
+            const result: DashboardParticipant[] = participants.map((participant, i) => ({
+                userGuid: participant.userGuid,
+                label: (participant.userProfile.firstName || participant.userProfile.lastName)
```
MocanaAtBroad, line 175: Can we always count on a userProfile being there? Would `participant.userProfile?.firstName` make sense?

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard-redesigned.component.ts
```diff
@@ -51,37 +80,119 @@ import { map, take, filter } from 'rxjs/operators';
...
+    styles: [`
+        .full-width {
+            width: 100%;
+        }
+    `]
 })
 export class DashboardRedesignedComponent extends DashboardComponent implements OnInit {
     public invitationId: string | null = null;
+    public dashboardParticipants$: Observable<DashboardParticipant[]>;
+    public panelsState = new Map<string, boolean>();
```
MocanaAtBroad: May I suggest renaming to `participantUserGuidToPanelIsOpen`?

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard.component.ts
```diff
@@ -89,11 +94,18 @@ export class DashboardComponent implements OnInit, OnDestroy {
         this.announcementMessages[index].shown = false;
     }
 
-    public navigate(id: string): void {
+    public navigate(id: string, participantGuid?: string): void {
```
MocanaAtBroad: This was here already, but can we rename `id` to `activityInstanceGuid`? I think that is what it is. (Took me me some effort to figure it out).

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard.component.ts
```diff
@@ -89,11 +94,18 @@ export class DashboardComponent implements OnInit, OnDestroy {
...
+        if (this.useParticipantDashboard) {
+            this.session.setParticipant(participantGuid);
```
MocanaAtBroad, line 110: Not getting why we set the session participant guid (we should rename here to as that is what it is).
Is it that the session maintain the current participant being operated on? I think we should add some comments to the Session class to be crystal clear what these things mean.

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard-redesigned.component.ts
```diff
@@ -51,37 +80,119 @@ import { map, take, filter } from 'rxjs/operators';
...
+        this.dashboardParticipants$ = this.getDashboardParticipants();
+    }
+
+    private getOperatorActivities(): Observable<Array<ActivityInstance> | null> {
+        return of(this.session.session.userGuid).pipe(
+            tap((userGuid) => {
+                this.session.setParticipant(userGuid);
+            }),
+            switchMap(() => this.userActivityServiceAgent.getActivities(of(this.config.studyGuid))));
+    }
```
MocanaAtBroad, line 163: Consider refactoring the code a bit.
A simple assignment instead of observable/tap can be used to `setParticipant` .
If intent is to set the participant in session so the `getActivities` refers to a specific participant, probably extending `getActivities` to include participantGuid as an argument.

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard-redesigned.component.ts
```diff
@@ -51,37 +80,119 @@ import { map, take, filter } from 'rxjs/operators';
...
+
+    private getDashboardParticipants(): Observable<DashboardParticipant[]> {
+        const operatorParticipant$ = this.getOperatorActivities().pipe(
+            tap((userActivities) => {
+                this.operatorActivities = userActivities;
+            }),
```
MocanaAtBroad: Have similar concern here with this tap. Yes, I think it is doing the right thing, but seems a bit too clever a use of side effects/doing double-duty/more than one thing that is not necessarily have the same intent objective.

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard-redesigned.component.ts
```diff
@@ -51,37 +80,119 @@ import { map, take, filter } from 'rxjs/operators';
...
+        protected session: SessionMementoService,
+        protected router: Router,
         private _announcements: AnnouncementsServiceAgent,
         private userInvitation: UserInvitationServiceAgent,
         _participantsSearch: ParticipantsSearchServiceAgent,
-        @Inject('toolkit.toolkitConfig') public config: ToolkitConfigurationService) {
-        super(_router, _announcements, _participantsSearch, config);
+        private governedParticipantsAgent: GovernedParticipantsServiceAgent,
+        private userActivityServiceAgent: UserActivityServiceAgent,
+        private userProfile: UserProfileServiceAgent,
```
MocanaAtBroad: I confused this variable for a profile rather than for a service. Please rename.

ddp-workspace/projects/ddp-sdk/src/lib/components/user/activities/userActivities.component.ts
```diff
@@ -201,10 +202,13 @@ import { take } from 'rxjs/operators';
...
 export class UserActivitiesComponent implements OnInit, OnDestroy, OnChanges, AfterContentInit {
     @Input() studyGuid: string;
+    // in case we use multiple ddp-user-activities component instances on the page we can't use session participantGuid
+    @Input() participantGuid: string;
+    @Input() activities: Array<ActivityInstance>;
     @Input() displayedColumns: Array<DashboardColumns> = ['name', 'summary', 'date', 'status', 'actions'];
     @Output() open: EventEmitter<string> = new EventEmitter();
     @Output() loadedEvent: EventEmitter<boolean> = new EventEmitter<boolean>();
-    public dataSource: UserActivitiesDataSource;
+    public dataSource: UserActivitiesDataSource | Array<ActivityInstance>;
```
MocanaAtBroad: Does it make sense in some cases getting the activities from and input parameter vs in other cases getting them from a datasource? Would it be clearer if activities to be displayed came from one clear path?
(I am surprised the table will take an array or a datasource! Had to look it up in APIs)

MocanaAtBroad (commented):
Looks good and works well! Just a few comments.

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard-redesigned.component.ts
aglushko-broad: material expancion panel component uses this prefix for identifying lazy loaded content
https://material.angular.io/components/expansion/overview#lazy-rendering
I think it's better using lazy loading here since every activity table component makes api request

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard-redesigned.component.ts
aglushko-broad: based on the Participant interface userProfile property is always defined

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard.component.ts
aglushko-broad: the current implementation of activity services allows navigating through activities of another user (participant). It's implemented by using participant guid from the session object. It was already used in atcp for example and in the prism.
We set the session participant guid before navigating in order to use already implemented mechanisms.
Not sure what we should rename, could you please be more specific? participantGuid?

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard-redesigned.component.ts
aglushko-broad: I'm not sure I get it.. what's you suggestion? where do you suggect to move this assignment?

ddp-workspace/projects/toolkit/src/lib/services/toolkitConfiguration.service.ts
aglushko-broad: done!

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard-redesigned.component.ts
aglushko-broad: sure, done!

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard.component.ts
aglushko-broad: done

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard-redesigned.component.ts
aglushko-broad: the thing is `setParticipant` should be done exactly in the data stream, at emission. Not when we create the stream. Otherwise we can set multiple participants and only after that call for activities with the last set participant. I updated the code and added a comment about that.
Also about you suggestion to move setParticipant into getActivities method. I thought it over, even implemented it, but it looked less obvious then why getActivities service can't be called before subscription happened, we have to add a comment about that, which is not making the situation better.

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard-redesigned.component.ts
aglushko-broad: nvm, I removed this piece of code

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard-redesigned.component.ts
aglushko-broad: done

ddp-workspace/projects/ddp-sdk/src/lib/components/user/activities/userActivities.component.ts
aglushko-broad: Actually it's a very good suggestion, I liked it and thought about that, but my impression was we used it in much places and it would be hard to refactor, but in fact we use it only in toolkit dashboard components. So done!

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard.component.ts
MocanaAtBroad: I would have expected session.setParticipant to take a user object or something like that. Takes a guid so should be called setParticipantGuid

MocanaAtBroad (approved):
I added a comment, but I think you did the important refactoring. We can come back at some point to the session object, getActivities etc.

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard.component.ts
aglushko-broad: I see your point here. I think it's fine that setParticipant takes id only. Probably the point of the method name setParticipant 
 is about that we update the session object as it may behaves from participant point of view, Right now we only need guid for that. It might need more data later

PR #721, opened 2021-07-19
DDP-6327: Support new `STUDY_REDIRECT` workflow

alextanlan:
Can be checked in Pancan study for now.
The workflow for testing:
```
- Click "Join Count me in"
- Choose: self (first option)
- Choose Diagnosis/type angio => choose "AngioSarcoma"
- Submit prequal 
```


ddp-workspace/projects/ddp-sdk/src/lib/components/confirmDialog/confirmDialog.component.html
```diff
@@ -2,6 +2,11 @@
     <span class="confirm-dialog-title"
           [innerHTML]="title | translate">
     </span>
+    <ng-container *ngIf="content">
```
aglushko-broad: #minor can remove ng-container and apply *ngIf directly to the div

ddp-workspace/projects/ddp-sdk/src/lib/components/confirmDialog/confirmDialog.component.ts
```diff
@@ -9,12 +9,16 @@ import { ThemePalette } from '@angular/material/core';
 })
 export class ConfirmDialogComponent {
     title: string;
+    content?: string;
+    contentSubstitutions?: object;
```
aglushko-broad: I'd recommend using not just an object but probably the object type {[key: string]: string} since values should be strings

ddp-workspace/projects/toolkit/src/lib/services/commands/studyRedirectCommand.service.ts
```diff
@@ -0,0 +1,54 @@
...
+        return new Observable(e => e);
+    }
+
+    private openDialog(): void {
+        const config = {
+            hasBackdrop: true,
+            autoFocus: false,
+            scrollStrategy: new NoopScrollStrategy(),
+            panelClass: 'study-redirect-dialog',
+            maxWidth: '500px',
```
aglushko-broad, line 28: wondering if we want to use ModalDialogService.getDialogConfig here..

ddp-workspace/projects/toolkit/src/lib/services/commands/studyRedirectCommand.service.ts
```diff
@@ -0,0 +1,54 @@
...
+                confirmBtnText: 'Toolkit.Dialogs.StudyRedirect.GoToProject',
+                cancelBtnText: 'Toolkit.Dialogs.StudyRedirect.Cancel',
+                confirmBtnColor: 'primary'
+            }
+        };
+
+        const dialogRef = this.dialog.open(ConfirmDialogComponent, config);
+        dialogRef.afterClosed().subscribe((confirmRedirect: boolean) => {
+            if (confirmRedirect) {
+                this.redirect(this.action.data?.redirectUrl);
```
aglushko-broad: if data is not defined should we through the error instead changing the href?

ddp-workspace/projects/ddp-sdk/src/lib/components/confirmDialog/confirmDialog.component.html
alextanlan: Fixed.

ddp-workspace/projects/ddp-sdk/src/lib/components/confirmDialog/confirmDialog.component.ts
alextanlan: Thanks, fixed.

ddp-workspace/projects/toolkit/src/lib/services/commands/studyRedirectCommand.service.ts
alextanlan: I feel it is not worth injecting the service, because we don't need to specify the dialog position / to link the dialog to any element.

ddp-workspace/projects/toolkit/src/lib/services/commands/studyRedirectCommand.service.ts
alextanlan: I guess the `data` must be defined in the case. 
Anyway, refactored a bit.

ddp-workspace/projects/toolkit/src/lib/services/commands/studyRedirectCommand.service.ts
aglushko-broad: got it! how about using DEFAULT_DIALOG_SETTINGS const?

aglushko-broad (approved):

ddp-workspace/projects/toolkit/src/lib/services/commands/studyRedirectCommand.service.ts
alextanlan: Will fix in the next PR

PR #734, opened 2021-07-23
DDP-6625 make groups in autocomplete picklist unselectable

aglushko-broad:


MocanaAtBroad:
Since we are matching group names as we type, we probably should underline the matching group name substring?
Screen Shot 2021-07-27 at 9 07 46 AM

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
```diff
@@ -57,8 +53,8 @@ export class AutocompleteActivityPicklistQuestion extends BaseActivityPicklistQu
     public ngOnInit(): void {
         const answer = this.block.answer && this.block.answer[0];
         if (answer) {
-            this.inputFormControl
-                .setValue(answer.detail || this.block.picklistSuggestions.find(option => option.value === answer.stableId));
+            this.inputFormControl.setValue(answer.detail || this.block.picklistGroups.flatMap(group => group.options)
```
MocanaAtBroad, line 56: Wouldn't this assignment end up setting a a value of type ActivityPicklistOption and wouldn't that be a problem?

aglushko-broad:
> Since we are matching group names as we type, we probably should underline the matching group name substring?
> Screen Shot 2021-07-27 at 9 07 46 AM

done

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
MocanaAtBroad: I see. Value can be a string or an iActivityPicklistOption.
Would it be simpler to convert values to strings and look up the option just when you have to? Just seeing lot of checks of string types and "string | ActivityPicklistOption"  type definitions?
(This a question/something to consider, not a request for a change).

ddp-workspace/projects/toolkit/tsconfig.lib.json
```diff
@@ -14,7 +14,7 @@
     "types": [],
     "lib": [
       "dom",
-      "es2018"
+      "es2019"
```
MocanaAtBroad, line 17: Presumably for flatMap? Make sure you can `ng build` the pancan project.

MocanaAtBroad (commented):
Code is clean and works nice. Just a couple of comments that might or might not require changes.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
aglushko-broad: we need to use the object for auocomplete model. Label is for displaying and stableId is a value. When user picks something from the list we send the whole object that contains the right id. It's faster than search for the id based on the label, so I'd prefer to not change it.
I think it's fine having unit types when we work with such controls like autocmplete with custom value

ddp-workspace/projects/toolkit/tsconfig.lib.json
aglushko-broad: Correct! Yes, it works w/o errors :)

alextanlan (approved):

PR #740, opened 2021-07-27
DDP-6606 Added possibility to hide block title, fixed prequal title

PavelYakupov:


aleksey-kuznetsov:
@yufengwng @MocanaAtBroad is it necessary to add these SDK changes? We may just configure the activity's title to be blank and not show it.

yufengwng:
I agree with @aleksey-kuznetsov. Title for content blocks are optional, you don't have to configure a "template" in the backend config file for the block. So in the JSON response client will get a `null` instead. Or you can configure it as blank.

PR #747, opened 2021-07-29
DDP-6563 fix activity steps styles

aglushko-broad:
mockups https://projects.invisionapp.com/d/main#/console/21368369/455066475/preview
before
image

after
image


alextanlan:
in Osteo and Brain studies the `Consent` activity steps looks differently.
I guess Pancan should look similar if we don't have special requirements, do we? 
May be it should be discussed with Erin / Marco ?

screen1

screen2

ddp-workspace/projects/ddp-pancan/src/styles/activities.scss
```diff
@@ -82,34 +82,70 @@
...
+        padding-left: 8px;
+    }
 
     &.completed {
         cursor: pointer;
-        color: $activity-step-completed;
 
+        .activity-step {
+            &__number {
+                background-color: #5a42d0;
```
alextanlan: `$activity-step-inactive` can be used here.

aglushko-broad:
> in Osteo and Brain studies the `Consent` activity steps looks differently.
> I guess Pancan should look similar if we don't have special requirements, do we?
> May be it should be discussed with Erin / Marco ?
> 
> screen1
> 
> screen2

Mocups look like a special requirement to me https://projects.invisionapp.com/d/main#/console/21368369/455066475/preview. So we have requirements. It was discussed and confirmed with Marco

ddp-workspace/projects/ddp-pancan/src/styles/activities.scss
aglushko-broad: good catch! done

alextanlan (approved):

PR #752, opened 2021-07-30
DDP-6563 fix password and error pages

aglushko-broad:
before
image
after
image
before
image
after
image


alextanlan:
What about applying the same for small screens (at least tablets) ? 
image

aglushko-broad:
> What about applying the same for small screens (at least tablets) ?
> image

Please specify what exactly should be fixed/changed for tables&mobiles and why. Thanks

alextanlan:
Nevermind, decline my previous comment.
I briefly checked osteo/brain. They have same styling.

alextanlan (approved):

PR #755, opened 2021-07-31
Mail address component fixes for DDP-6537

MocanaAtBroad:
Added a bunch of comments. I think comments tell story if you read them from top to bottom on the "files changed" view
**Update from 8/2/2021**
Initial PR broke some functionality, particularly the section validation and the scrolling to the first error in page. Tried to restore that functionality and still enable proper integration of mail address component.
Tried to be clear with my thinking on changes with comments in PR and code.

(I think I mentioned it before, but this convinces me that the activity component is due for some refactoring. Too much to take on now, but this is getting out of hand!)


ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -558,21 +558,27 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
...
+        const activityRequirementsMet$: Observable<boolean> = currentAddress$.pipe(
             map(currentAddress => this.meetsActivityRequirements(currentAddress))
         );
 
         const emitValidStatusAction$ = combineLatest([
             this.formValidStatusChanged$,
             this.formErrorMessages$,
             this.verifyFieldErrors$,
-            activitityRequirementsMet$
+            activityRequirementsMet$
```
MocanaAtBroad, line 581: typo

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -190,28 +190,48 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
...
+                                this.visitedSectionIndexes[nextIndex] = true;
+                                this.saveLastVisitedSectionIndex(nextIndex);
+                                this.currentSectionIndex = nextIndex;
+                            }
+                        })
+                    )
+                ),
+                take(1)
+            ).subscribe();
         }
```
MocanaAtBroad, line 223: From activity component this was the big change: what we do when we hit the Next button:
First, we trigger this.submitAnnouncementService.announceSubmit().

The timer(100) line there is to allow the mailing address to (perhaps, becoming "busy" after getting submit announced).

Mailing address component listens for this signal to trigger a "real" saving of the mailing address.

Then we listen for the page not being busy before proceeding to do validations increment steps, etc. Most of that stuff that happens once not busy is pretty much what was there before.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/baseActivity.component.ts
```diff
@@ -238,7 +238,6 @@ export abstract class BaseActivityComponent implements OnChanges, OnDestroy {
 
     protected resetValidationState(): void {
         this.setFlags(false);
-        this.submitAttempted.next(false);
```
MocanaAtBroad, line 242: This was triggering the submit announce, so the announcement was happening twice, including when going backwards. Just not needed and totally unrelated to "resetting validation state" anyways.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -190,28 +190,48 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
...
+                delay(1),
+                filter(pageIsBusy => !pageIsBusy),
+                debounceTime(250),
+                tap(() => {
+                    // if we move forwards or backwards, let's reset our validation display
+                    this.resetValidationState();
+                    this.currentSectionIndex = previousIndex;
+                }),
+                take(1)
+            ).subscribe();
```
MocanaAtBroad, line 241: Here to, we hold on with decrementing steps while the page is busy.
Note that we DO NOT do a submit announce! Seems to me that if going backwards we should not save address like going forwards.
Note that the mailing address, on its own, does save copy of the fields automatically to temp address IF there is no default address in existence. This happens anyways, so if some fields are entered for first time and hit the Previous button, when you come back the address fields entered should still be there.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -448,8 +449,8 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
         let processSubmitAnnouncement$;
         if (this.submitService) {
             processSubmitAnnouncement$ = this.submitService.submitAnnounced$.pipe(
-                tap(() => this.saveTrigger$.next()
-                ));
+                tap(() => this.saveTrigger$.next())
+            );
```
MocanaAtBroad, line 453: Whitespace. Ignore.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -509,34 +510,33 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
...
-        // "Real" as opposed to "Temp"
+        // "Real" as opposed to "Temp". Important we return the saved address as properties are added on server
         const saveRealAddressAction$ = this.saveTrigger$.pipe(
             withLatestFrom(this.formErrorMessages$, currentAddress$),
             filter(([_, formErrors, addressToSave]) => {
                 return !formErrors?.length && canSaveRealAddress(addressToSave);
             }),
             tap(() => busyCounter$.next(1)),
             concatMap(([_, formErrors, addressToSave]) => this.addressService.saveAddress(addressToSave, false)),
-            removeTempAddressOperator(),
```
MocanaAtBroad: These "operator" here was the cause of the blanking of fields on submit.
We are feeding back to the form the address sent back from server after save (the server sometimes add fields values that we need to send back).
But problem here was that, at the end of the definition of the the saveRealAddressAction$ observable we were returning the result of the deletion (an HTTP response object). So no name, no street1 field etc, so corresponding form field got blanked out.
Taking the deletion out of here corrected this problem.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -509,34 +510,33 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
...
+            filter(([_, state]) => !!state.activityInstanceGuid),
+            tap(() => busyCounter$.next(1)),
+            concatMap(([_, state]) => this.addressService.deleteTempAddress(state.activityInstanceGuid)),
+            catchError(() => {
+                this.logger.logDebug(this.LOG_SOURCE, 'Temp delete failed. This is OK.');
+                return of(null);
+            }),
+            tap(() => busyCounter$.next(-1))
+        );
+
```
MocanaAtBroad, line 539: We did want to delete the temp address if the real mailing address is saved (this is just for cleanup purpose), so deletion factored into its own action. Triggered by saveRealAddressAction$ observable emitting after completion of its action.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -558,21 +558,27 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
             tap(address => this.inputAddress$.next(address))
         );
 
-        const emitComponentBusyAction$ = combineLatest([this.isInputComponentBusy$, isThisComponentBusy$]).pipe(
+        const addressComponentBusy$ = combineLatest([this.isInputComponentBusy$, isThisComponentBusy$]).pipe(
             map(busyFlags => busyFlags.some(val => val)),
             distinctUntilChanged(),
+            share());
+
```
MocanaAtBroad, line 565: Refactoring component busy so we can reuse the observable...

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -558,21 +558,27 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
...
+        const emitComponentBusyAction$ = addressComponentBusy$.pipe(
+            // let's smooth out our busyness signals
+            debounce(busy => interval(busy ? 0 : 250)),
+            distinctUntilChanged(),
             tap(isBusy => this.componentBusy.emit(isBusy))
         );
```
MocanaAtBroad, line 571: If we are busy, we want to report that right away. If we are not busy, we want to wait just a little bit before reporting just in case we become busy again.
distinctUntilChanged just keeps the number of emissions to only changes between busy and not busy.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -660,16 +677,6 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
...
-        this.componentBusy.pipe(
-            filter(busy => !busy),
-            tap(() => {
-                this.ngUnsubscribe.next();
-                this.ngUnsubscribe.complete();
-                this.logger.logDebug(this.LOG_SOURCE, 'Unsubscribe completed.');
-            }),
-            take(1)
-        ).subscribe();
+        this.destroyComponentSignal$.next();
```
MocanaAtBroad, line 688: This was another cause of some of the problem: when we move from one section to the next in the activity, ngOnDestroy gets called on the no longer visible component.
But there was a bug here and ngOnDestroy sometimes was not triggering the unsubscribe.
And this lead to unsubscribed code receiving the submitannouncement from activity component and trying to save even though the component was not visible and not fully loaded.
So refactored just to signal that ngOnDestroy has been called. Real code is above.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -581,15 +587,26 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
...
 
+        // We need to trigger an unsubscribe, but don't want to do it prematurely. Wait til we are not busy!
+        combineLatest([this.destroyComponentSignal$, addressComponentBusy$]).pipe(
+            filter(([_, busy]) => !busy),
+            take(1))
+            .subscribe(() => {
+                this.ngUnsubscribe.next();
+                this.ngUnsubscribe.complete();
+            });
+
```
MocanaAtBroad, line 605: This is the unsubscribe that most of the code in this component depends on.
(we were having a problem sometime with either the unsubscribe not occurring at all and sometimes happening prematurely, while saving or cleaning up was still going on). More on this below.

aglushko-broad:
Marco, could you please sync with develop branch and resolve the conflicts?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -66,7 +66,7 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
         private logger: LoggingService,
         private windowRef: WindowRef,
         private renderer: Renderer2,
-        private submitService: SubmitAnnouncementService,
+        private submitAnnouncementService: SubmitAnnouncementService,
```
MocanaAtBroad, line 69: Rename to make clearer what this is.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -91,9 +91,7 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
     public ngOnInit(): void {
         this.getActivity();
         this.initStepperState();
-        const submitSub = this.submitAttempted.subscribe((isSubmit: boolean) => {
-            this.submitService.announceSubmit(isSubmit);
-        });
+        const submitSub = this.submitAttempted.subscribe(() => this.submitAnnouncementService.announceSubmit(null));
```
aglushko-broad: are we sure we want to ignore the subject value and always pass null?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -189,52 +187,51 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
...
-                    this.isLoaded$.next(true);
-                    this.currentSectionIndex = nextIndex;
-                    this.visitedSectionIndexes[nextIndex] = true;
-                    this.saveLastVisitedSectionIndex(nextIndex);
-                });
-            }
+            this.submitAnnouncementService.announceSubmit(null);
+            // The announcement could make listener components busy, but not instantly
+            // introduce a wait before we check whether we busy or not
+            timer(100).pipe(
```
aglushko-broad, line 193: why it's 100? should we use this.timeToDebounce?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -189,52 +187,51 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
...
             }
-
-            this.isPageBusy.pipe(
-                tap(_ => this.isLoaded$.next(false)),
-                debounceTime(this.timeToDebounce),
-                filter(busy => !busy),
+            this.isPageBusy.pipe(startWith(true)).pipe(
+                delay(1),
+                filter(pageIsBusy => !pageIsBusy),
+                debounceTime(250),
```
aglushko-broad: let's use this.timeToDebounce

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -518,48 +510,41 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
...
-                if (!canSave) {
-                    this.block.hasValidAddress = false;
-                }
-                return !formErrors?.length && canSave;
+                return !formErrors?.length && canSaveRealAddress(addressToSave);
             }),
             tap(() => busyCounter$.next(1)),
             concatMap(([_, formErrors, addressToSave]) => this.addressService.saveAddress(addressToSave, false)),
-            tap(() => busyCounter$.next(-1)),
             removeTempAddressOperator(),
```
alextanlan: Seems like it should be removed as well (the operator is removed at line 521)

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -273,12 +274,8 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
                 }),
                 map(defaultAddress => [state, defaultAddress]),
             )),
-            tap(([_, defaultAddress]: [ComponentState, Address | null]) => {
-                if (defaultAddress) {
-                    this.inputAddress$.next(defaultAddress);
-                    this.block.hasValidAddress = true;
```
aglushko-broad, line 279: if we don't use hasValidAddress anymore should we refactor MailAddressBlock model?

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -518,48 +510,41 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
...
          */
         const touchFormOnSubmitWithBadAddressAction$ = this.validationRequested$.pipe(
             withLatestFrom(currentAddress$),
-            filter(([_, addressToSave]) => {
-                return !canSaveRealAddress(addressToSave)
-                    || (this.block.requireVerified && !this.block.hasValidAddress);
-            }),
-            tap(() => this.addressInputComponent.markAddressTouched())
+            filter(([_, addressToSave]) => !canSaveRealAddress(addressToSave)),
+            tap(() => this.addressInputComponent.touchAllControls())
```
alextanlan: There is no method `touchAllControls`, it is changed/renamed in 11e4d8af187c509bb7a5863f54663906552a0fd4

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
alextanlan: Good catch!

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
MocanaAtBroad: b80a3d821e7a653381ed4603d6b37d767d247c33

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
MocanaAtBroad: We should probably convert the type to void. It is just a signal that submit is occurring. And sometime we don't have the body anyways.

MocanaAtBroad: c41ed31263e59cf692d6c16b0cb2b283ba41555a

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
MocanaAtBroad: Was trying to keep time to smallest possible value. The value of 250 would work too, I suppose, though this is not really debouncing either.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
MocanaAtBroad: f49d5e7f83d0349d6db5750a2f7642c2f11bf8b5

MocanaAtBroad:
Just converted to draft for now as I realize I don't think it fully deals with error validation/might have merged this functionality out by mistake. Will come back to this later today...

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
MocanaAtBroad: Additional changes since last time: had to add additional things to incrementStep to get back the "scroll to error" feature. This was happening before when triggering a form submit, but we are doing less than that. The comments in code describe what is going on.

MocanaAtBroad: Also added back incrementStep and decrementStep the "isLoaded$" observable updates. Appears that stepper depends on it. Hopefully I did this right?

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
MocanaAtBroad: I added it back in a different place. (Though right now this is a redundant validation, but it is convenient).

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -574,44 +558,63 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
...
-                this.validStatusChanged.emit(status);
-            })
+            tap(status => this.validStatusChanged.emit(status))
         );
 
+        // At least for now, we'll report the status of mailing address through block too
+        // (needed to limit status checks to blocks that are visible within a section
+        const updateBlockValidStatusAction$ = this.validStatusChanged.pipe(
+            tap(validStatus => this.block.hasValidAddress  = validStatus)
+        );
```
MocanaAtBroad, line 594: As the comment indicates, we are setting the same value for validAddress in the block as we are sending via the validStatusChanged.
The code in develop branch was setting the hasValidAddress in different location of code with probably the same value. I am making it explicit here.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -189,52 +187,59 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
...
+                        filter(pageIsBusy => !pageIsBusy),
+                        tap(() => {
+                            // run validations and compute flags for blocks to enable scrolling to errors
+                            this.currentSection.validate();
+                            this.model && this.model.shouldScrollToFirstInvalidQuestion();
+                            // triggers scrolling
+                            this.validationRequested = true;
+                        }),
+                        // delay needed for validationRequested to be processed
+                        delay(10),
```
aglushko-broad: I see too many new magic numbers. I'm not that familiar with this piece of functionality as you guys, but I feel like we have problems with chosen approach and can't extend/fix something w/o small delay/timeout hacks with random timers here and where which is not very good... Does any alternatives exist? Or do you know if some refactoring can fix it?

MocanaAtBroad: In this case, the delay is more of a "yield". In other words, need to allow for asynchronous operations triggered by the assignment:
```typescript
this.validationRequested = true;
```
Otherwise, code in this method would continue to execute and the code that is listening for changes to the validationRequested variable would not execute.

I have updated the delay to be 0. It works just as well:
a9c206014

I also removed another delay that is not needed in `decrementStep`

Perhaps using a rxjs schedulers would achieve same thing? I have not explored it and don't have a lot of experience with them.

Reality is that my my main concern here is to make sure I can  integrate the mail address component without breaking  existing functionality.

As I have mentioned before, I think this component is overdue for some refactoring. Different people have continued to tack on functionality over years, added their own flags, observables, ngOnChange handlers etcetera to grow it. And as @alextanlan is well aware, we also have to worry about integrating activities inside of activities.

This change to activity.component.ts is not going to be beautiful. Hopefully just enough to work and not introduce bugs.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
MocanaAtBroad: @alextanlan Could you please double-check my updates to isLoaded$ subject? I think in earlier version I had broken displaying next and previous section by setting by my use of `isLoaded$.next(false)` and don't think I fully understand how it fits in the bigger picture.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -189,52 +187,63 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
...
+                        tap(() => {
+                            this.sendSectionAnalytics();
+                            tap(_ => this.isLoaded$.next(true)),
```
alextanlan: Incorrectly placed `tap` (into another `tap`).
Same for line 242.

MocanaAtBroad: Yes. This was not good!
a9c206014

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
alextanlan: I am in progress..

alextanlan:
Adding `take(1)` in  `initStepperState` method pipe (activity.component) would solve [brocken stepper logic bug](https://broadinstitute.atlassian.net/browse/DDP-6673), I believe.

MocanaAtBroad:
> Adding `take(1)` in `initStepperState` method pipe (activity.component) would solve [brocken stepper logic bug](https://broadinstitute.atlassian.net/browse/DDP-6673), I believe.

7c0d5a80f02c0351e953ad5c99ec711612fe0ced

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -189,52 +187,59 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
...
-            this.isPageBusy.pipe(
-                tap(_ => this.isLoaded$.next(false)),
+            this.isPageBusy.pipe(startWith(true)).pipe(
+                filter(pageIsBusy => !pageIsBusy),
                 debounceTime(this.timeToDebounce),
-                filter(busy => !busy),
+                tap(() => {
+                    // if we move forwards or backwards, let's reset our validation display
+                    this.resetValidationState();
+                    this.isLoaded$.next(true);
```
alextanlan: Seems like it should be eliminated as well ( as you removed all `isLoaded$` in `incrementStep`/`decrementStep`)

MocanaAtBroad: Yup. Looks like it is not needed. Removed: e96b37efb

alextanlan:
LGTM

alextanlan (approved):

MocanaAtBroad:
Thanks for you help on this PR! FYI: I have created a ticket a tech debt ticket to refactor activity component.
https://broadinstitute.atlassian.net/browse/DDP-6699
When we are done with PanCan we can figure out what we want longer term

PR #786, opened 2021-08-09
Add sorting policy to picklist autocomplete

MocanaAtBroad:
Try to deal with the sorting issue in a somewhat declarative manner.
Users want options sorted alphabetically in the specific language. At least for now, I think it is easiest to do on client side, so here it is.

**August 10 Updates:**
Had to make a couple of changes:
1. The stableId for option that sorting will push to last in list is UNSURE (was 'UNKNOWN'). This is set in Pancan app.module.ts
2. Added explicit creation of instance of sorting policy in ddp.module.ts . Found that Angular injection does not like to have arguments in constructors that are not other injected service or parameters.


ddp-workspace/projects/ddp-pancan/src/app/app.module.ts
```diff
@@ -155,6 +147,11 @@ export function translateFactory(translate: TranslateService,
             ],
             multi: true
         },
+        // Ensure that sorting of autocomplete picklist options is as specified
+        {
+            provide: PicklistSortingPolicy,
+            useValue: new PicklistSortingPolicy(SortOrder.ALPHABETICAL, 'UNKNOWN')
+        }
```
MocanaAtBroad, line 165: The default policy leave things exactly as they are. This makes things alphabetical and pushes the UNKNOWN option to last.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
```diff
@@ -73,10 +74,15 @@ export class AutocompleteActivityPicklistQuestion extends BaseActivityPicklistQu
...
+
+
         userQueryStream.pipe(startWith('')).subscribe((value: string | ActivityPicklistOption) => {
             const query = typeof value === 'string' ? value : value.optionLabel;
-            this.filteredGroups = this.filterOutEmptyGroups(query ? this.filterGroups(query) : this.block.picklistGroups.slice());
-            this.filteredOptions = query ? this.filterOptions(query, this.block.picklistOptions) : this.block.picklistOptions.slice();
+            this.filteredGroups = this.filterOutEmptyGroups(query ? this.filterGroups(query, sortedPicklistGroups)
+                : sortedPicklistGroups.slice());
+            this.filteredOptions = query ? this.filterOptions(query, sortedPicklistOptions) : sortedPicklistOptions.slice();
         });
```
MocanaAtBroad, line 87: Just applying the sorting of incoming options when component is setup.

ddp-workspace/projects/ddp-sdk/src/lib/services/picklistSortingPolicy.service.ts
```diff
@@ -0,0 +1,49 @@
...
+export class PicklistSortingPolicy {
+    constructor(readonly mainSortOrder: SortOrder = SortOrder.NONE, readonly lastStableId?: string) {}
+
+    public sortPicklistGroups(groups: ActivityPicklistNormalizedGroup[]): ActivityPicklistNormalizedGroup[] {
+        const groupsCopy = groups.slice();
+        if (this.mainSortOrder === SortOrder.ALPHABETICAL) {
+            const sortedPicklistGroups = groupsCopy.sort(this.buildComparator('name'));
+            sortedPicklistGroups.forEach(group => group.options = this.sortPicklistOptions(group.options));
+        }
+        return groupsCopy;
```
MocanaAtBroad, line 17: SortOrder.NONE should just result in copy of options in exactly same order. Otherwise we sort stuff.

ddp-workspace/projects/ddp-sdk/src/lib/services/picklistSortingPolicy.service.ts
```diff
@@ -0,0 +1,49 @@
...
+        const optionsCopy = options.slice();
+        if (this.lastStableId || this.mainSortOrder === SortOrder.ALPHABETICAL) {
+            const comparePicklistOptions = (a: ActivityPicklistOption, b: ActivityPicklistOption, lastOptionStableId?: string) => {
+                if (lastOptionStableId) {
+                    if (a.stableId === lastOptionStableId) {
+                        return 1;
+                    } else if (b.stableId === lastOptionStableId) {
+                        return -1;
+                    }
+                }
```
MocanaAtBroad, line 30: Push down the list the lastOption if there is one.

aglushko-broad (dismissed):

ddp-workspace/projects/ddp-sdk/src/lib/services/picklistSortingPolicy.service.ts
```diff
@@ -0,0 +1,49 @@
+import { ActivityPicklistNormalizedGroup, ActivityPicklistOption } from 'ddp-sdk';
+import { SortOrder } from './sortOrder';
+import { Injectable } from '@angular/core';
+
+@Injectable({
+    providedIn: 'root',
```
alextanlan: I believe `{provideIn: 'root'}` is not needed because the service is already provided in ddp.module

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
```diff
@@ -73,10 +74,15 @@ export class AutocompleteActivityPicklistQuestion extends BaseActivityPicklistQu
...
+        const sortedPicklistGroups = this.sortPolicy.sortPicklistGroups(this.block.picklistGroups);
+        const sortedPicklistOptions = this.sortPolicy.sortPicklistOptions(this.block.picklistOptions);
+
+
         userQueryStream.pipe(startWith('')).subscribe((value: string | ActivityPicklistOption) => {
             const query = typeof value === 'string' ? value : value.optionLabel;
-            this.filteredGroups = this.filterOutEmptyGroups(query ? this.filterGroups(query) : this.block.picklistGroups.slice());
-            this.filteredOptions = query ? this.filterOptions(query, this.block.picklistOptions) : this.block.picklistOptions.slice();
+            this.filteredGroups = this.filterOutEmptyGroups(query ? this.filterGroups(query, sortedPicklistGroups)
+                : sortedPicklistGroups.slice());
```
alextanlan, line 84: #minor#optional

Seems like `.slice()` isn't necessary. (line 85 as well)
`sortPolicy` methods return a copy of groups/options and they never mutate in the component.

ddp-workspace/projects/ddp-sdk/src/lib/services/picklistSortingPolicy.service.ts
aglushko-broad: I'd recommend to not remove provideIn, it's recommended way to inject services in applications
https://angular.io/guide/providers
for example it improves tree-shaking
it's better to not provide it in sdk module in this case

alextanlan: On the one hand it makes sense,
on the other hand all services in `ddp-sdk` are provided in the module only.

alextanlan (approved):

PR #787, opened 2021-08-10
DDP-6667: LMS page redesign

alextanlan:
_**Update mid of 10 Aug, 2021**_:
`LMS page WILL NOT BE TRANSLATED/will only be in English.`

Created some fixes 5827bfabb64499a9fcb36db49cd4247414247ad2, 4c963bfb49c259982ffa212e2963f554cc4ae74a, c0aa188dd26375208fc8eb9579c10f53f21f982b

================

a1
a2
a3
a4
a5


MocanaAtBroad (dismissed):
Looks good!

ddp-workspace/projects/ddp-pancan/src/app/components/splash-pages/lms-page/lms-page.component.ts
```diff
@@ -11,19 +17,41 @@ import { FormControl } from '@angular/forms';
...
+        private dialog: MatDialog,
+        @Inject('toolkit.toolkitConfig') private config: ToolkitConfigurationService,
+        private translateService: TranslateService,
+        private log: LoggingService
+    ) {
         this.phone = config.lmsPagePhone;
         this.email = config.lmsPageEmail;
     }
 
+    notify(): void {
```
aglushko-broad: do we want to rename to openJoinMailingList?

ddp-workspace/projects/ddp-pancan/src/app/components/splash-pages/lms-page/lms-page.component.ts
```diff
@@ -11,19 +17,41 @@ import { FormControl } from '@angular/forms';
...
+        this.dialog.open(JoinMailingListComponent, {
+            ...JOIN_MAILING_LIST_DIALOG_SETTINGS,
+            data: {
+                studyGuid: this.config.lmsStudyGuid,
+                useLanguage: 'en'
```
aglushko-broad, line 37: does it mean that we should always open JoinMailingList dialog in english only for all languages? if not, could you please explain this part a liitle

ddp-workspace/projects/ddp-pancan/src/app/components/splash-pages/lms-page/lms-page.component.ts
alextanlan: Renamed

ddp-workspace/projects/ddp-pancan/src/app/components/splash-pages/lms-page/lms-page.component.ts
alextanlan: We got a message from CMI that LMs page will not be translated. 
It means we should keep the page (and related stuff like joinMailingList) in English even though a user can switch to another language (Spanish) for whole study.
Other pages and related joinMailingList dialog should follow the chosen language.

aglushko-broad: 2 questions then: should we switch back to the selected language when we navigate from lms page and isn't enough to switch englush translation only in lms parent component? why do we need to to it one more time in dialog component? the service is singletone and should be global for the whole app

MocanaAtBroad (dismissed):

ddp-workspace/projects/ddp-pancan/src/app/components/splash-pages/lms-page/lms-page.component.ts
alextanlan: > should we switch back to the selected language when we navigate from lms page

We should not. We do not change the current language in translate service. We just get needed translations and apply it locally.
q1

alextanlan: >isn't enough to switch englush translation only in lms parent component? why do we need to to it one more time in dialog component?

It is not enough. I believe it is because we use `JoinMailingList.component` not directly on LMS page , but as an entry component.

MocanaAtBroad (approved):

PR #808, opened 2021-08-13
PANCAN: FIX: Fix a bug with incorrect Next/Submit button

alextanlan:
**_Flow to reproduce :_** (https://broadinstitute.slack.com/archives/C0231BHPM7F/p1628791888059700)
In pancan dev:
Choose : child option
Age 10 or 11 so that we get Assent:
Choose diagnosis: colon cancer
Once you complete consent-assent; you will see this new: Consent Form Addendum: Stool Collection activity.
You will see the **Submit** button (should be **Next**)

-- 

**_Thoughts:_**
the root of the issue is same like for [PR#780](https://github.com/broadinstitute/ddp-angular/pull/780) - using properties from a previous activity.

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
@@ -22,7 +22,7 @@ import { PatchAnswerResponse } from '../../models/activity/patchAnswerResponse';
 import { ActivitySection } from '../../models/activity/activitySection';
 import { AnalyticsEventCategories } from '../../models/analyticsEventCategories';
 import { CompositeDisposable } from '../../compositeDisposable';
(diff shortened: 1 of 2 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -382,8 +382,7 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
                         this.visitedSectionIndexes = this.visitedSectionIndexes
                             .map((value, index) => index <= this.currentSectionIndex);
                     }
-                }),
-                take(1)
```
alextanlan, line 386: If we take it only once it does not update  `this.currentSectionIndex` (line 381) and uses a previous activity value.

MocanaAtBroad (approved):

PR #813, opened 2021-08-16
DDP-6783: Add exception for sorting autocomplete lists [PANCAN]

alextanlan:
PanCan: all the pickilist autocomplete lists should sort options alphabetically (and then leaves the UNSURE option as last).
EXCEPT for cancer diagnostic in prequal.

Implemented a temporary solution on client-side. 
I believe, generally (in future) we should have several sorting options on client-side and on back-end set special flag/field for `ActivityPicklistQuestion`, like  `sort: NONE / ALPHABET / ANOTHER SORTING`.

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-composite-answer/activityCompositeAnswer.component.html
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-composite-answer/activityCompositeAnswer.component.html
@@ -14,6 +14,7 @@
                 [block]="childBlock"
                 [readonly]="readonly"
                 [validationRequested]="validationRequested"
+                [studyGuid]="config.studyGuid"
                 (valueChanged)="updateValue(row, column, $event)"
                 (componentBusy)="componentBusy.next($event)">
             </ddp-activity-answer>
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-composite-answer/activityCompositeAnswer.component.html
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-composite-answer/activityCompositeAnswer.component.html
@@ -14,7 +14,6 @@
                 [block]="childBlock"
                 [readonly]="readonly"
                 [validationRequested]="validationRequested"
-                [studyGuid]="config.studyGuid"
                 (valueChanged)="updateValue(row, column, $event)"
                 (componentBusy)="componentBusy.next($event)">
             </ddp-activity-answer>
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
@@ -1,4 +1,4 @@
-import { Component, Input, OnChanges, OnDestroy, OnInit, SimpleChanges } from '@angular/core';
+import { Component, OnChanges, OnDestroy, OnInit, SimpleChanges } from '@angular/core';
 import { BaseActivityPicklistQuestion } from './baseActivityPicklistQuestion.component';
 import { NGXTranslateService } from '../../../services/internationalization/ngxTranslate.service';
 import { Subject } from 'rxjs';
@@ -47,9 +47,7 @@ import { PicklistSortingPolicy } from '../../../services/picklistSortingPolicy.s
 })
 export class AutocompleteActivityPicklistQuestion extends BaseActivityPicklistQuestion implements OnInit, OnDestroy, OnChanges {
     private readonly ngUnsubscribe = new Subject();
-    private readonly DO_NOT_SORT_LIST_STABLE_IDS = ['PRIMARY_CANCER_SELF', 'PRIMARY_CANCER_CHILD'];
 
-    @Input() studyGuid: string;
     filteredGroups: ActivityPicklistNormalizedGroup[] = [];
     // options w/o a group
     filteredOptions: ActivityPicklistOption[] = [];
@@ -76,15 +74,9 @@ export class AutocompleteActivityPicklistQuestion extends BaseActivityPicklistQu
             takeUntil(this.ngUnsubscribe)
         );
 
-        let sortedPicklistGroups;
-        let sortedPicklistOptions;
-        if (this.shouldBeSorted) {
-            sortedPicklistGroups = this.sortPolicy.sortPicklistGroups(this.block.picklistGroups);
-            sortedPicklistOptions = this.sortPolicy.sortPicklistOptions(this.block.picklistOptions);
-        } else {
-            sortedPicklistGroups = this.block.picklistGroups;
-            sortedPicklistOptions = this.block.picklistOptions;
-        }
+        const sortedPicklistGroups = this.sortPolicy.sortPicklistGroups(this.block.picklistGroups);
+        const sortedPicklistOptions = this.sortPolicy.sortPicklistOptions(this.block.picklistOptions);
+
 
         userQueryStream.pipe(startWith('')).subscribe((value: string | ActivityPicklistOption) => {
             const query = typeof value === 'string' ? value : value.optionLabel;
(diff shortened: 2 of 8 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
```diff
@@ -150,6 +158,11 @@ export class AutocompleteActivityPicklistQuestion extends BaseActivityPicklistQu
         return options.filter(option => option.optionLabel.toLowerCase().includes(filterValue));
     }
 
+    private get shouldBeSorted(): boolean {
+        const doNotSort = this.studyGuid === 'cmi-pancan' && this.DO_NOT_SORT_LIST_STABLE_IDS.includes(this.block.stableId);
```
aglushko-broad: what do you think if instead of specifying exception study id and the list of exception picklists ids we can do one this things
1. Add new model property to question json, for example noSorting: boolean and specify on backend this field as true for prequlifier cancers list in config. Then check this field on frontend
OR if changes on backend would be hard to implement
2. Add new property in ConfigurationService (for example picklistsWithNoSorting: string[]) for list of ids for every study and then check on this sdk component if study has such configuration and do not sort picklists from that list

alextanlan: As I wrote in [the PR description](https://github.com/broadinstitute/ddp-angular/pull/813#issue-713306386):
  > Implemented a temporary solution on client-side. (because we're running out of time)

For now we are using Autocomplete component in Pancan study only. Thus having list of `picklistsWithNoSorting` ids  either in config or in the component itself seem pretty the same, being a temporary decision.

More optimal way is your suggestion#1: get sorting flag/value for `Autocomplete picklist` from back-end as a separate field. It's also written in [description](https://github.com/broadinstitute/ddp-angular/pull/813#issue-713306386). Moreover it can be not just boolean, but an enum/value from a list of sorting order, if we have more than one way of sorting (e.g. 'none', 'alphabet A-Z',  'alphabet Z-A', 'randomly mixed', ...). I guess, we will have chosen that option in the foreseeable future.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
```diff
@@ -47,7 +47,9 @@ import { PicklistSortingPolicy } from '../../../services/picklistSortingPolicy.s
 })
 export class AutocompleteActivityPicklistQuestion extends BaseActivityPicklistQuestion implements OnInit, OnDestroy, OnChanges {
     private readonly ngUnsubscribe = new Subject();
+    private readonly DO_NOT_SORT_LIST_STABLE_IDS = ['PRIMARY_CANCER_SELF', 'PRIMARY_CANCER_CHILD'];
 
+    @Input() studyGuid: string;
```
aglushko-broad: we can access config from any sdk component and take studyGuid from it, no need to pass it via input

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
aglushko-broad: you're right that it's much better having the backend configuration for that, but as far as I understand there is no task in backlog for moving the sorting config on a backend, which means it would be on FE for uncertain time. 
I'd recommend to keep study specific configs in ConfigurationService (app.module.ts) even for temporal solutions inctead of keeping it right in sdk component. It will help to keep sdk component clean and simple. Moreover each sdk component has access to ConfigurationService inctance and can easily extract any study specific configuration

MocanaAtBroad: RIght. Was hoping for solution that kept the SDK code without dependencies to Pan Can.
The piece that eventually would make sense to make part of the platform and include in our API is that our study configuration can specify, for each picklist question, a sort policy. We don't have time to do that, so what we need to do in client is to hard-code the knowledge of which question has picklist options that need to be sorted and which questions do not. 
I think we can do better than hard-code the stable ids into the component.
As you recall, to add the SortPolicy in the first place I went to great lengths to make sure it was something that could be specified outside of the component source code. I hope that we can do something similar for this ticket.
It ought to be possible using a combination configuration service and (maybe, but hopefully not) Angular service injection.
IMPORTANT! If this looks too hard to finish by today, this is OK. Will just go with the hard-coded solution and fix after launch. We have ALMOST run out of time here.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-composite-answer/activityCompositeAnswer.component.html
```diff
@@ -14,6 +14,7 @@
                 [block]="childBlock"
                 [readonly]="readonly"
                 [validationRequested]="validationRequested"
+                [studyGuid]="config.studyGuid"
```
MocanaAtBroad: The config.studyGuid is, for all practical purposes, a read-only global. If you need it, just get by injecting the config service were needed.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
alextanlan: Fixed.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
alextanlan: I see.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-composite-answer/activityCompositeAnswer.component.html
alextanlan: ok

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
```diff
@@ -150,6 +162,10 @@ export class AutocompleteActivityPicklistQuestion extends BaseActivityPicklistQu
         return options.filter(option => option.optionLabel.toLowerCase().includes(filterValue));
     }
 
+    private get shouldBeSorted(): boolean {
+        return !this.config.picklistsWithNoSorting.includes(this.block.stableId);
```
aglushko-broad, line 166: we need to check if picklistsWithNoSorting is defined (will be not defined for most of studies)

aglushko-broad (approved):

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
```diff
@@ -74,9 +80,15 @@ export class AutocompleteActivityPicklistQuestion extends BaseActivityPicklistQu
...
-
+        let sortedPicklistGroups;
+        let sortedPicklistOptions;
+        if (this.shouldBeSorted) {
+            sortedPicklistGroups = this.sortPolicy.sortPicklistGroups(this.block.picklistGroups);
+            sortedPicklistOptions = this.sortPolicy.sortPicklistOptions(this.block.picklistOptions);
+        } else {
+            sortedPicklistGroups = this.block.picklistGroups;
+            sortedPicklistOptions = this.block.picklistOptions;
+        }
```
MocanaAtBroad, line 91: Decision to sort or not to sort is encoded in SortPolicy. You just have to configure it depending on what you want for the question.
https://github.com/broadinstitute/ddp-angular/blob/develop/ddp-workspace/projects/ddp-sdk/src/lib/services/sortOrder.ts
The default is NONE
https://github.com/broadinstitute/ddp-angular/blob/69f7d745da2e89fec6a6b7e6a996f5c662cf6838/ddp-workspace/projects/ddp-sdk/src/lib/services/picklistSortingPolicy.service.ts#L8

for body locations we want ALPHABETICAL. For the other ones NONE.

PR #815, opened 2021-08-16
DDP-6767: Fix shifted footer border for screen 1024px [PANCAN]

alextanlan:
**_Before_**

**_After_**

### DIFF
--- a/ddp-workspace/projects/ddp-pancan/src/app/components/footer/footer.component.scss
+++ b/ddp-workspace/projects/ddp-pancan/src/app/components/footer/footer.component.scss
@@ -61,7 +61,7 @@
     min-width: fit-content;
 }
 
-@media screen and (min-width: $desktop-breakpoint) {
+@media screen and (min-width: $desktop-breakpoint + 1) {
     .actions {
         height: 8.625rem;
 


ddp-workspace/projects/ddp-pancan/src/app/components/footer/footer.component.scss
```diff
@@ -61,7 +61,7 @@
     min-width: fit-content;
 }
 
-@media screen and (min-width: $desktop-breakpoint) {
+@media screen and (min-width: $desktop-breakpoint + 1) {
```
aglushko-broad, line 64: what do you think if we replace mobile first(min-width) with desktop first (max-width) here?

alextanlan: Not sure if I get the point.
we already have 
```  
@media screen and (max-width: $desktop-breakpoint) {  // line 79
   ...change flex direction inside of some footer elements
}
 ```


The layout issue appears for the exact value `1024px`.

aglushko-broad: I mean this
https://github.com/broadinstitute/ddp-angular/pull/816/files
what do you think?

alextanlan: Deal, merged.
Will close this PR as redundant.

PR #826, opened 2021-08-17
DDP-6682: Force scroll to top after submit an activity

alextanlan:
Can't reproduce any issue from [DDP-6682](https://broadinstitute.atlassian.net/browse/DDP-6682), [DDP-6813](https://broadinstitute.atlassian.net/browse/DDP-6813).

Suggest to call explicitly the `scrollToTop` method when an activity is submited.


MocanaAtBroad (commented):
This is a tough one because can't provide reproducible steps. But does not make sense to me that scrolling to top after submit on current activity helps us with scroll position for the next activity.
Not saying this is wrong. Just saying I am not seeing it. If you have a theory for how this might solve problem, please write it. If not, and this is just an educated guess, let's just skip this issue for now.

PR #842, opened 2021-08-19
DDP-6752: *OLD* [Pancan] picklist autocomplete : treat the "-", '/' characters as a space

alextanlan:
The ticket: [DDP-6752](https://broadinstitute.atlassian.net/browse/DDP-6752)
Also see [the slack discussion](https://broadinstitute.slack.com/archives/C02AAP2AM98/p1628262486037600).

Implemented: 
symbols `dash` ('-'), `slash`('/') are also treated as a `space` (' ') in Pancan cancers list autocomplete search.
b1
b2
b3

Known issue (extra benefit?) :
Any symbols from `substitutableSymbols` are interchangeable.
Typing 'B/cell' will find 'B/cell' / 'B-cell' / 'B cell' etc.
Requiring strict independent matching is puzzling some kind of generic implementation.

### DIFF
--- a/ddp-workspace/projects/ddp-pancan/src/app/app.module.ts
+++ b/ddp-workspace/projects/ddp-pancan/src/app/app.module.ts
@@ -92,6 +92,8 @@ sdkConfig.doGcpErrorReporting = DDP_ENV.doGcpErrorReporting;
 sdkConfig.tooltipIconUrl = 'assets/images/info.png';
 sdkConfig.useStepsWithCircle = true;
 sdkConfig.picklistsWithNoSorting = ['PRIMARY_CANCER_SELF', 'PRIMARY_CANCER_CHILD'];
+sdkConfig.picklistsWithSubstitutionSymbols = ['PRIMARY_CANCER_SELF', 'PRIMARY_CANCER_CHILD'];
+sdkConfig.picklistsSubstitutionSymbols = ['-', '/'];
 
 export function translateFactory(translate: TranslateService,
     injector: Injector,
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
@@ -23,13 +23,13 @@ import { RegExpHelper } from '../../../utility/RegExpHelper';
 
         <mat-autocomplete #autoCompleteFromSource="matAutocomplete" class="autoCompletePanel" [displayWith]="displayAutoComplete">
             <mat-optgroup *ngFor="let group of filteredGroups">
-                <strong [innerHtml]="group.name | searchHighlight: userQueryRegExp$.getValue()"></strong>
+                <strong [innerHtml]="group.name | searchHighlight: userQueryRegExp"></strong>
                 <ng-container *ngTemplateOutlet="generalOptionsList; context: {list: group.options}"></ng-container>
             </mat-optgroup>
             <ng-container *ngTemplateOutlet="generalOptionsList; context: {list: filteredOptions}"></ng-container>
             <ng-template #generalOptionsList let-list="list">
                 <mat-option *ngFor="let suggestion of list" class="autoCompleteOption" [value]="suggestion">
-                    <span [innerHtml]="suggestion.optionLabel | searchHighlight: userQueryRegExp$.getValue()"></span>
+                    <span [innerHtml]="suggestion.optionLabel | searchHighlight: userQueryRegExp"></span>
                 </mat-option>
             </ng-template>
         </mat-autocomplete>
@@ -55,7 +55,7 @@ export class AutocompleteActivityPicklistQuestion extends BaseActivityPicklistQu
     // options w/o a group
     filteredOptions: ActivityPicklistOption[] = [];
     inputFormControl = new FormControl();
-    userQueryRegExp$ = new BehaviorSubject<RegExp>(null);
+    userQueryRegExp: RegExp;
     private substitutableSymbols: string[] = [];
 
     constructor(
@@ -91,9 +91,7 @@ export class AutocompleteActivityPicklistQuestion extends BaseActivityPicklistQu
             .subscribe((value: string | ActivityPicklistOption) => {
                 const query = typeof value === 'string' ? value : value.optionLabel;
                 if (query) {
-                    this.userQueryRegExp$.next(
-                        RegExpHelper.createSubstitutableSymbolsRegExp(query, this.substitutableSymbols)
-                    );
+                    this.userQueryRegExp = RegExpHelper.createSubstitutableSymbolsRegExp(query, this.substitutableSymbols);
                     this.filteredGroups = this.filterOutEmptyGroups(this.filterGroups(sortedPicklistGroups));
                     this.filteredOptions = this.filterOptions(sortedPicklistOptions);
                 } else {
@@ -173,13 +171,14 @@ export class AutocompleteActivityPicklistQuestion extends BaseActivityPicklistQu
         return groups
             .map(group => ({
                 name: group.name,
-                options: this.userQueryRegExp$.getValue().test(group.name) ?
-                    group.options : this.filterOptions(group.options)
+                options: this.userQueryRegExp.test(group.name)
+                    ? group.options
+                    : this.filterOptions(group.options)
             }));
     }
 
     private filterOptions(options: ActivityPicklistOption[]): ActivityPicklistOption[] {
-        return options.filter(option => this.userQueryRegExp$.getValue().test(option.optionLabel));
+        return options.filter(option => this.userQueryRegExp.test(option.optionLabel));
     }
 
     private get shouldBeSorted(): boolean {
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
@@ -1,6 +1,6 @@
 import { Component, Inject, OnChanges, OnDestroy, OnInit, SimpleChanges } from '@angular/core';
 import { FormControl } from '@angular/forms';
-import { BehaviorSubject, Subject } from 'rxjs';
+import { Subject } from 'rxjs';
 import { debounceTime, distinctUntilChanged, map, startWith, takeUntil } from 'rxjs/operators';
 
 import { BaseActivityPicklistQuestion } from './baseActivityPicklistQuestion.component';
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
@@ -11,8 +11,6 @@ import { PicklistSortingPolicy } from '../../../services/picklistSortingPolicy.s
 import { ConfigurationService } from '../../../services/configuration.service';
 import { RegExpHelper } from '../../../utility/RegExpHelper';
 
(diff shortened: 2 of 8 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
```diff
@@ -54,6 +55,8 @@ export class AutocompleteActivityPicklistQuestion extends BaseActivityPicklistQu
     // options w/o a group
     filteredOptions: ActivityPicklistOption[] = [];
     inputFormControl = new FormControl();
+    userQueryRegExp$ = new BehaviorSubject<RegExp>(null);
```
aglushko-broad: we never subscribe on userQueryRegExp$, I guess we can do not use BehaviorSubject here and just set the RegExp value into the userQueryRegExp

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
```diff
@@ -147,19 +169,17 @@ export class AutocompleteActivityPicklistQuestion extends BaseActivityPicklistQu
...
-        const filterValue = name.toLowerCase();
+    private filterGroups(groups: ActivityPicklistNormalizedGroup[]): ActivityPicklistNormalizedGroup[] {
         return groups
             .map(group => ({
                 name: group.name,
-                options: group.name.toLowerCase().includes(filterValue)
-                    ? group.options
-                    : this.filterOptions(filterValue, group.options)}));
+                options: this.userQueryRegExp$.getValue().test(group.name) ?
+                    group.options : this.filterOptions(group.options)
```
aglushko-broad: #minor let's keep the previous wrapping: each value on a new line

ddp-workspace/projects/ddp-sdk/src/lib/utility/RegExpHelper.ts
```diff
@@ -0,0 +1,50 @@
...
+    *  regExp.test('asd -    zxc') = true;
+    */
+    static createSubstitutableSymbolsRegExp(text: string, substitutableSymbols: string[] = [], replaceValue = ' '): RegExp {
+        const anyAmountOfSpacesRegExp = '\\s*';
+        if (substitutableSymbols.includes(text) || !substitutableSymbols.some(s => text.includes(s))) {
+            // return same text but space/case insensitive
+            return new RegExp(
+                RegExpHelper.replaceEmptySpaces(
+                    RegExpHelper.escapeRegExp(text),
+                    anyAmountOfSpacesRegExp
```
aglushko-broad: why do we replace empty spaces with anyAmountOfSpacesRegExp?

ddp-workspace/projects/ddp-sdk/src/lib/utility/RegExpHelper.ts
```diff
@@ -0,0 +1,50 @@
...
+                ),
+                'gi'
+            );
+        }
+
+        const escapedSubstitutableSymbols = substitutableSymbols.map(symbol => RegExpHelper.escapeRegExp(symbol));
```
aglushko-broad: if we know that we want to escape all symbols from the list we can simplify
`const escapedSubstitutableSymbols = substitutableSymbols.map(symbol => '\\${symbol}');`

ddp-workspace/projects/ddp-sdk/src/lib/utility/RegExpHelper.ts
```diff
@@ -0,0 +1,50 @@
...
+        const substitutableSymbolsRegExp = new RegExp(escapedSubstitutableSymbols.join('|'), 'g');
+
+        const pattern = RegExpHelper.escapeRegExp(
+                RegExpHelper.replaceEmptySpaces(text, ''),
```
aglushko-broad, line 26: do we want to remove all spaces between words? not sure it would work if the option has multiple words + substitutable symbols

ddp-workspace/projects/ddp-sdk/src/lib/utility/RegExpHelper.ts
```diff
@@ -0,0 +1,50 @@
...
+                // have issues with incorrect handling it in text, escaping substitutableSymbols twice due to the regex creating.
+                // Let's avoid escaping them here.
+                substitutableSymbols
+            )
+            .replace(
+                substitutableSymbolsRegExp,
+                [anyAmountOfSpacesRegExp, '[', ...escapedSubstitutableSymbols, replaceValue, ']', anyAmountOfSpacesRegExp].join('')
```
aglushko-broad, line 37: this line might need a comment, or described test cases

ddp-workspace/projects/ddp-sdk/src/lib/utility/RegExpHelper.ts
aglushko-broad: 'brain stem glioma d' string doesn't find 'Brain stem glioma / Diffuse Intrinsic Pontine Glioma (DIPG)'
maybe it's related

alextanlan: The example above should not be working (from required behavior).
e.g.
`B-cell` should match `B cell`
`B/cell` should match `B cell`
but
`B cell` should not match neither `B-cell` nor `B/cell`

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
alextanlan: Right, thanks.
Just was thinking about using an Observable somehow and didn't manage it, forgot to remove.
Fixed.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
alextanlan: fixed

ddp-workspace/projects/ddp-sdk/src/lib/utility/RegExpHelper.ts
alextanlan: just for more handy search (not strict matching).
 Like: `the  (some mistyped spaces here)      brain` will match `the brain`


But, you are right. 
Changed a bit (peplaced any amount of spaces with one space) as to avoid such cases:
 `the  rwise` will match `otherwise`

ddp-workspace/projects/ddp-sdk/src/lib/utility/RegExpHelper.ts
alextanlan: Done

ddp-workspace/projects/ddp-sdk/src/lib/utility/RegExpHelper.ts
alextanlan: Added both.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
```diff
@@ -1,6 +1,6 @@
 import { Component, Inject, OnChanges, OnDestroy, OnInit, SimpleChanges } from '@angular/core';
 import { FormControl } from '@angular/forms';
-import { Subject } from 'rxjs';
+import { BehaviorSubject, Subject } from 'rxjs';
```
aglushko-broad: let's remove BehaviorSubject import

alextanlan: Sure, removed.

ddp-workspace/projects/ddp-pancan/src/app/app.module.ts
```diff
@@ -92,6 +92,8 @@ sdkConfig.doGcpErrorReporting = DDP_ENV.doGcpErrorReporting;
 sdkConfig.tooltipIconUrl = 'assets/images/info.png';
 sdkConfig.useStepsWithCircle = true;
 sdkConfig.picklistsWithNoSorting = ['PRIMARY_CANCER_SELF', 'PRIMARY_CANCER_CHILD'];
+sdkConfig.picklistsWithSubstitutionSymbols = ['PRIMARY_CANCER_SELF', 'PRIMARY_CANCER_CHILD'];
+sdkConfig.picklistsSubstitutionSymbols = ['-', '/'];
```
MocanaAtBroad, line 96: I don't see a reason to make this a configurable option. I was thinking that this would be generic/every time we include the functionality we should treat these characters same as white space.
I suppose best of all worlds would be to, by default treat these symbols as whitespace and allow a particular study or question override that behavior.

ddp-workspace/projects/ddp-sdk/src/lib/services/configuration.service.ts
```diff
@@ -73,4 +73,6 @@ export class ConfigurationService {
     prismDashboardRoute: string;
     prismRoute: string;
     picklistsWithNoSorting: string[] = [];
+    picklistsWithSubstitutionSymbols: string[] = [];
+    picklistsSubstitutionSymbols: string[] = [];
```
MocanaAtBroad, line 77: Comments with descriptions of these are needed. (Including picklistsWithNoSorting). Still thinking about the names....

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
```diff
@@ -90,14 +87,20 @@ export class AutocompleteActivityPicklistQuestion extends BaseActivityPicklistQu
             sortedPicklistOptions = this.block.picklistOptions;
         }
```
MocanaAtBroad, line 89: I am unable to select here lines 80 through 89, but I believe this can be refactored to simplify.
Sort Policy already has a NO SORT option, so you can remove the conditional:
```
if (this.shouldBeSorted)
```
and instead construct a PicklistSortingPolicy based on your config `picklistsWithNoSorting` option (we should probably rename to `picklistAutocompleteOptionsWithNoSorting` ). So if config option set, then PicklistSortingPolicy use SortOrder=NONE .

ddp-workspace/projects/ddp-sdk/src/lib/utility/RegExpHelper.ts
MocanaAtBroad: Actually, what we are looking for here is is to include more matches, not less.
So if option label is B-Cell
And user enters B Cell or B-Cell or B/Cell or b-cell or B-CELL or B CELL, etc, I want to see the existing option B-Cell.
And same with the reverse: i.e., if we have B Cell in option list, I want to a match with all these weird combinations.
I think this can be achieved without regex (not saying not to use regex, just saying that the matching we are talking about does not require it).
So it would looks like something like:
1. Normalize all the options labels, converting everything to uppercase and replace all "substitutable" symbols to spaces OR convert the labels to list of words.
2. Apply same conversion to the entered search query.
3. Look for matches.

One thing to emphasize is that the ticket is more about addressing CMI's generic concern that users will enter something and they will miss some of the cancer options that ARE in the option list, but because they used or did not use a given separator or used a given capitalization or even the word order is different, user did not get a match.

I would add that as I look at the cancer list, we should expand functionality here to have words to ignore, e.g., I would like configure the cancer list so that the word "cancer" is ignored.

I take responsibility for not speccing this is in greater detail. But at same time, I am asking for engineer to put forth a solution that addresses the user's concern.

alextanlan:
@MocanaAtBroad , @aglushko-broad 
Finally, I have created a new [PR#872](https://github.com/broadinstitute/ddp-angular/pull/872) with the suggested approach. 
Also fixed needed comments there.

Will close this PR soon.

PR #844, opened 2021-08-20
Add delay of 2 secs delay after all busyness has ceased before proceeding with activity PUT

MocanaAtBroad:
This is the other part of fix discussed in pull request https://github.com/broadinstitute/ddp-study-server/pull/982

The code appeared to be doing the right thing, even doing manual tests locally (withholding response from server when mailing address was saved to ensure it final PUT/activity submit did not execute early).

But server data shows that the strict serialization we were going for was somehow failing. So added a couple of secs buffer on top of retries on the server waiting for mailing address when doing snapshots.

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/baseActivity.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/baseActivity.component.ts
@@ -203,6 +203,8 @@ export abstract class BaseActivityComponent implements OnChanges, OnDestroy {
                     filter(isPageBusy => !isPageBusy),
                     take(1))
                 ),
+                // extra buffer to allow for asynchronous saves like mailing address to complete
+                delay(2000),
                 // if we can't save we are not going past filter, so do necessary cleanup here
                 tap(() => !this.isAllFormContentValid.value && (this.dataEntryDisabled = false)),
                 filter(() => this.isAllFormContentValid.value),


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/baseActivity.component.ts
```diff
@@ -203,6 +203,8 @@ export abstract class BaseActivityComponent implements OnChanges, OnDestroy {
                     filter(isPageBusy => !isPageBusy),
                     take(1))
                 ),
+                // extra buffer to allow for asynchronous saves like mailing address to complete
+                delay(2000),
```
aglushko-broad, line 207: would be nice to have some explanation why 2 seconds exactly. Might be done separately

aglushko-broad (approved):

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/baseActivity.component.ts
MocanaAtBroad: Well, before change, code appeared to do right thing and wait for mail address to complete saving/wait for response from server before proceeding to submitting the activity.
But server logs confirm that this was not happening every time: mail address save is still going on when PUT for activity starts processing.
At this point I simply do not know why code Angular code was not enforcing serialization of requests.
2 seconds just seems like a safe margin to wait for mail address. The differences the differences in mail address save end time and PUT processing start time where in the range of .25 to .5. I am counting on 2 secs to provide enough margin.
I don't think this is a final solution to problem. I rather deal with the more fundamental issue of request serialization, but that is not going to happen with our available time to launch.
So 2 seconds delay on client AND server retries waiting for address to get there should cover problem to get us to launch. A more comprehensive/fundamental fix is TBD.

alextanlan: Agree. Seems like a hack which helps us to decrease a probability of error case (with low efforts by the way).  
Can be merged. 

For future, imo, we should think about it/investigate: 
- what if 2s is not enough ? (some connections delay)
- may be, it makes sense to make `activity submit PUT ` to wait till `address save POST` is done (keep `AddressEmbeddedComponent` busy). 
Something like:
 ```
 const savedAddress$: Observable<Address | null> = saveRealAddressAction$.pipe(
            tap(() => busyCounter$.next(1)),
            filter(savedAddressVal => !!savedAddressVal),
            share()
        );

        savedAddress$
            .subscribe((a) => {
                busyCounter$.next(-1);
            });
```
On the other hand, in some activities the mail address is not obligatory.

MocanaAtBroad: Right. This IS a hack which combined with a hack on the server side should virtually guarantee the error does not occur.
On the busy counter idea: code already does something like this. There are busy counters around all the HTTP requests including saving AND there is a debounce to wait to report that component is no longer busy.
We definitely want to come back at this. The asynchronous/independent operation of the mail address component (and to a lesser extent, the institution components) has been a source of headaches for a long time.
As I have mentioned before (and have JIRA ticket somewhere for it) I want to refactor the activity component. There are so many things going on and we have continued to add more things (asynchronous saves, validation of form data and component data, resetting of validation messages, embedded activities, etc.etc.).
At any rate.Yes: this is a hack, but gets us to launch. Can look at a correct solution without Pan Can over our heads.

PR #859, opened 2021-08-25
DDP-6343 unique validation rule

aglushko-broad:


ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityUniqueValidationRule.ts
```diff
@@ -0,0 +1,25 @@
...
+        }
+
+        // compare only single select picklist questions
+        if (answers.find(([{value}]) => !Array.isArray(value) || value.length > 1)) {
+            return true;
+        }
+
+        const answersIds = answers.map(([{value}]) => value[0].stableId);
+        this.result = this.message;
+        return (new Set(answersIds)).size === answersIds.length;
```
alextanlan: What about the case when several `Other` options selected ? 
They have the same `stableId` (e.g. `stableId: "OTHER_CANCER"`)

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityUniqueValidationRule.ts
```diff
@@ -0,0 +1,25 @@
...
+            return true;
+        }
+
+        // compare only single select picklist questions
+        if (answers.find(([{value}]) => !Array.isArray(value) || value.length > 1)) {
+            return true;
+        }
+
+        const answersIds = answers.map(([{value}]) => value[0].stableId);
+        this.result = this.message;
```
alextanlan: May be, it makes sense to set `this.result = null` for valid cases.

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityUniqueValidationRule.ts
aglushko-broad: very good point! fixed

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityUniqueValidationRule.ts
aglushko-broad: done

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityUniqueValidationRule.ts
```diff
@@ -0,0 +1,34 @@
...
+import { PicklistSelectMode } from '../../../models/activity/picklistSelectMode';
+
+export class ActivityUniqueValidationRule extends ActivityAbstractValidationRule {
+    constructor(question: ActivityQuestionBlock<any>) {
+        super(question);
+    }
+
+    public recalculate(): boolean {
+        const {answer: answers} = this.question;
+        if (this.question.questionType !== QuestionType.Composite || answers.length === 1) {
```
MocanaAtBroad: Maybe:
```suggestion
        if (this.question.questionType !== QuestionType.Composite || answers.length <= 1) {
```

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityUniqueValidationRule.ts
```diff
@@ -0,0 +1,34 @@
...
+            this.result = null;
+            return true;
+        }
+
+        const [firstChild] = (this.question as ActivityCompositeQuestionBlock).children;
+
+        if (firstChild.questionType !== QuestionType.Picklist
+            || (firstChild as ActivityPicklistQuestionBlock).selectMode !== PicklistSelectMode.SINGLE) {
+            this.result = null;
+            return true;
```
MocanaAtBroad, line 36: I was expecting that our implementation could deal with with cases beyond picklist.
e.g., so if we have a composite that the child is just a text question,I would like to use the UNIQUE rule to flag that somebody typed "FOO" twice/in two children.

MocanaAtBroad: To support all the different non-composite question comparisons, it might make sense to factor out to the individual answers or blocks the comparison logic, e.g., add an abstract method like `isEqual(other: T)` to ActivityQuestionBlock . Leave it unimplemented in abstract class.
This makes it easier to keep Unique rule up to date over time/as we add more question types; if a new question type is added, you are forced to create a new question block sub type. And then the compiler remind you that you have to implements isEqual. Otherwise we always have to remember that UNIQUE rule code needs to be updated whenever a new question type comes along. I don't want to have to remember that.

aglushko-broad: totally agree, done!

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityUniqueValidationRule.ts
aglushko-broad: done

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityUniqueValidationRule.ts
MocanaAtBroad: If I am reading your solution, I think it works if we have the one child question (like in cancer/single picklist) but I think it won't work if a composite is made up of multiple questions, e.g., a date and a number.

I think the use of Sets to find duplicates is generally a good idea though I personally am disappointed that out-of-the box implementation does not allow you in Javascript to define your one equality/depends on the == operator (maybe there is a way to override prototype implementation?

It occurred to me that you can PROBABLY get away with a single implementation of compare regardless of the answer type. Would look something like this in the AbstractActivityQuestionBlock (using the Underscore isEqual implementation)

```typescript
public answersAreEqual(answer1: T, answer2: T): boolean {
        return _.isEqual(answer1, answer2);
    }
```
Can potentially be overridden if for some reason we thought for a particular answer type` _.isEqual` would not work

So then trick is to go through the composite answers and see if something in one of the answer columns is different. Looks something like this (not tested! just for discussion purposes)

```typescript
const childQuestions: ActivityQuestionBlock<any>[] = (this.question as ActivityCompositeQuestionBlock).children;
const rowsOfAnswers: AnswerContainer[][] = (this.question as ActivityCompositeQuestionBlock).answer;
        childQuestions.forEach((currentChildQuestion, questionIdx) => {
            let valueInFirstRow: any;
            rowsOfAnswers.forEach((rowOfAnswers: AnswerContainer[], answerRowIdx) => {
                const currentAnswerValue = rowOfAnswers[questionIdx].value;
                if (answerRowIdx === 0) {
                    valueInFirstRow = currentAnswerValue;
                } else if (!currentChildQuestion.answersAreEqual(valueInFirstRow, currentAnswerValue)) {
                    return false;
                }
            });
        });
        return true;
```

And if we confident that `_isEqual` always works for all our answers, then we could just use it directly in the rule and not bother implementing answersAreEqual() .

At any rate, just illustration of what was in my head. But just make sure that your implementation works with composites that have multiple child questions.

aglushko-broad: 1. Is it possible composite question has questions of multiple types? like number and date? I thought the implementation allows us to add only one type on question to composite question, no?
2. In general if we want to compare 2 different types it's always a custom logic. So I'm not sure what's your suggestion here. Do you want to have implementation to compare each type of question with each type of questions? it sounds like plenty of logic and if there is no the case but just thoughs how to make it more abstract I'd suggest to keep for now the implementation of comparing the answers of only one type.

MocanaAtBroad: > 1. Is it possible composite question has questions of multiple types? like number and date? I thought the implementation allows us to add only one type on question to composite question, no?
Composites questions are designed to have one or more child questions. In the case of the cancer question, it is one child picklist question.
But here is one with a text question and a date as children:
Screen Shot 2021-09-02 at 12 25 27 PM
The allowMultiple=true|false determines whether we can have multiple rows of answers. 


> 2. In general if we want to compare 2 different types it's always a custom logic. So I'm not sure what's your suggestion here. Do you want to have implementation to compare each type of question with each type of questions? it sounds like plenty of logic and if there is no the case but just thoughs how to make it more abstract I'd suggest to keep for now the implementation of comparing the answers of only one type.
Well, custom logic might not be necessary.
For the composite we DO KNOW what are the child question types. And each question type has a specific answer type, e.g.:
```typescript
export class ActivityTextQuestionBlock extends ActivityQuestionBlock<string>
```
For composite, the answers are arranged in rows of answers:
```typescript
export class ActivityCompositeQuestionBlock extends ActivityQuestionBlock<AnswerContainer[][]>
```
```typescript
export interface AnswerContainer {
    stableId: string;
    value: any;
}
```
the `value: any` is any of the answer types that we support (this definition could be more specific at this point).
So in the 2 dimensional array, each row of AnswerContainer has the answers for a row composite answers. And that row of answers is in the order of that child questions have been arranged in
So in the example I pasted above, you will have multiple rows of answers, where the first column will always be the text question answer and the second column will always be a date question answer.

In the sample code I posted in the previous comment, I am proposing that, for any single child question, we can PROBABLY get away with calling _.isEqual to compare the answers. 

And actually, now that we are talking about it, there is a very good chance that you can just call _.isEqual to compare 2 rows of composite answers, so that would simplify the solution even further!

At any rate, the two important points are that composites can have multiple child questions AND that we can probably get away with a generic comparison to deal with all kinds of answers. The comparison can single answer to single answer or, as it just occurred to me here, the comparison can be on a row of answers by row of answers basis.

(I have typed a lot of stuff, but I am not sure if I have not made things more confusing. If yes we can talk. I know you are about to take off on vacation so if totally confusing, maybe best we pick it up when you back)

aglushko-broad: ooohhh I see now, thanks a lot for that example with multiple question blocks per one composite question! I didn't thought about this case actually... But it's fixed now, please take a look. I hope I took into account all cases :) but feel free to add more comments if you feel something is missing

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityUniqueValidationRule.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public recalculate(): boolean {
+        const {answer: answers} = this.question;
+        if (this.question.questionType !== QuestionType.Composite || answers.length <= 1) {
+            this.result = null;
+            return true;
+        }
+
+        const {children} = (this.question as ActivityCompositeQuestionBlock);
```
MocanaAtBroad: Can we rename to variable `children` to `childQuestionBlocks` ? I found myself inspecting ActivityCompositeQuestionBlock code trying to figure out what this was

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityUniqueValidationRule.ts
```diff
@@ -0,0 +1,40 @@
...
+
+        const compositeAnswersStrings: string[] = answers.map(values => {
+            let result = '';
+            for (const [i, value] of values.entries()) {
+                const block = children[i];
+                // can't calculate a file's uniqueness
+                if (block.questionType !== QuestionType.File) {
+                    result += value.value ? `${value.stableId}:${children[i].convertToString(value.value)};` : '';
```
MocanaAtBroad: The one question special case? file answers seem to include fileName and fileSize for file answers. I would say that if those properties have same values in to different answers, we have duplicate, no?

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityUniqueValidationRule.ts
```diff
@@ -0,0 +1,40 @@
...
+import { QuestionType } from '../../../models/activity/questionType';
+import { ActivityCompositeQuestionBlock } from 'ddp-sdk';
+
+export class ActivityUniqueValidationRule extends ActivityAbstractValidationRule {
+    constructor(question: ActivityQuestionBlock<any>) {
+        super(question);
+    }
+
+    public recalculate(): boolean {
+        const {answer: answers} = this.question;
```
MocanaAtBroad: Can we add the type for `answers`. Pretty important detail in this chunk of code. I get very confused looking at two dimensional arrays. Please make it super-obvious what I am looking at (in the sample code I provided in comment below I named variables and added types because I (myself) was having a hard time tracking what was what in my own snippet).

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityUniqueValidationRule.ts
MocanaAtBroad: Could you double-check this? My reading of your code is that if we have multiple answer rows and they are unique.

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityUniqueValidationRule.ts
aglushko-broad: no no, we don't go to the condition block id questionType is a file. The code is confusing, I'll update it and replace `block.questionType !== QuestionType.File` to `block.questionType === QuestionType.File` and add break

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityUniqueValidationRule.ts
aglushko-broad: oh yea, you're right, I need to update the result property depending on the validation return. But in fact I'm not really sure there is a sence to assign the null to the result property if validation is passed. I know it's done in the other validators but it seems like that result property is just not used if validator returns true. So I see no sence to calculate it to be honest

aglushko-broad: but I'll double check

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityUniqueValidationRule.ts
aglushko-broad: fileName and fileSize is not enough to say that files are unique. file name might be the same, file size might be the same but files might be different

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityUniqueValidationRule.ts
aglushko-broad: done!

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityUniqueValidationRule.ts
aglushko-broad: done

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityUniqueValidationRule.ts
aglushko-broad: fixed

MocanaAtBroad (approved):

PR #864, opened 2021-08-27
Circadia: Sleep Log integration

aleksey-kuznetsov:
This PR introduces integration between Circadia study and BWH product - Sleep Log.
All requests are first issued to the GCP cloud function proxy which in turn makes request to Sleep Log API.
A simple link is displayed on participants dashboard which opens a new window with daily questionnaire.

Note: a cloud function used for this feature expects to receive Auth `clientId`. Since we plan to use this function locally AND on `dev` we have to pass `clientId` for `dev` env even if app is running locally.

### DIFF
--- a/ddp-workspace/projects/ddp-circadia/src/app/components/pages/dashboard/dashboard.component.ts
+++ b/ddp-workspace/projects/ddp-circadia/src/app/components/pages/dashboard/dashboard.component.ts
@@ -1,4 +1,4 @@
-import { Component, Inject, OnInit } from '@angular/core';
+import { Component, Inject, OnDestroy, OnInit } from '@angular/core';
 import { Router } from '@angular/router';
 import { forkJoin, Observable, of } from 'rxjs';
 import { concatMap, map, take, tap } from 'rxjs/operators';
@@ -7,6 +7,7 @@ import {
   ActivityInstance,
   AnnouncementMessage,
   AnnouncementsServiceAgent,
+  CompositeDisposable,
   ConfigurationService,
   UserActivityServiceAgent,
 } from 'ddp-sdk';
@@ -20,9 +21,10 @@ import { SleepLogService } from '../../../services/sleep-log.service';
   templateUrl: './dashboard.component.html',
   styleUrls: ['./dashboard.component.scss'],
 })
-export class DashboardComponent implements OnInit {
+export class DashboardComponent implements OnInit, OnDestroy {
   loading = true;
   activities: ActivityInstance[] = [];
+  private subs = new CompositeDisposable();
 
   constructor(
     private router: Router,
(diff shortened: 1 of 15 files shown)


ddp-workspace/projects/ddp-circadia/src/app/components/pages/dashboard/dashboard.component.ts
```diff
@@ -23,11 +27,29 @@ export class DashboardComponent implements OnInit {
...
+        concatMap(({ activities, sleepLogCodes }) =>
+          this.sleepLogService.runSleepLogActions(
+            sleepLogCodes,
+            !!activities.find(
+              activity => activity.activityCode === ActivityCode.MiniSleepLog,
+            ),
+          ),
+        ),
+      )
+      .subscribe();
```
MocanaAtBroad, line 54: I think you want to make sure you to unsubscribe or do take(1) to terminate subscription.

ddp-workspace/projects/ddp-circadia/src/app/services/sleep-log.service.ts
```diff
@@ -0,0 +1,193 @@
...
+        return of(null);
+    }
+  }
+
+  private createUser(): Observable<CreateUserResponse | null> {
+    const payload: CreateUserPayload = {
+      ...this.getCommonProxyPayload(HttpMethod.Post),
+      start: this.datePipe.transform(Date.now(), SleepLogService.DATE_FORMAT),
+      end: this.datePipe.transform(
+        Date.now() +
```
MocanaAtBroad: You sure you want to get time from user's browser instead of from server? Worrying about being dependent on local computer time being accurate/and or consistent across all users.

aleksey-kuznetsov: I guess we don't have an alternative unless an API changes what it accepts in request's payload.

aleksey-kuznetsov:
Sorry about this base branch changes, I was a bit confused due to lots of studies and branches 😓

ddp-workspace/projects/ddp-circadia/src/app/services/sleep-log.service.ts
MocanaAtBroad: Do we control the back-end service? Or is this an external service provided by our collaborators? To follow up on the date business, if we implement the back-end, we should change the back-end to get its own date.

ddp-workspace/projects/ddp-circadia/src/app/components/pages/dashboard/dashboard.component.ts
PavelYakupov: Above you can see this fragments of code that terminate subs:
```
...
this.subs.addNew(sub);
...
ngOnDestroy(): void {
    this.subs.removeAll();
}
```
In this case I 100% sure that this Observables emit values only one time, because they call services methods.

ddp-workspace/projects/ddp-circadia/src/app/services/sleep-log.service.ts
PavelYakupov: I removed this.

MocanaAtBroad (approved):

PR #870, opened 2021-08-31
DDP-6924: Fix a bug when a composite list is not refreshed after some child deletion [Pancan]

alextanlan:
Note: the Not updating on the block data model caused the issue.


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-composite-answer/activityCompositeAnswer.component.ts
```diff
@@ -151,7 +151,9 @@ export class ActivityCompositeAnswer implements OnChanges {
 
     public removeRow(index: number): void {
         this.childQuestionBlocks.splice(index, 1);
-        this.valueChanged.emit(this.buildComponentAnswers());
+        const childAnswers = this.buildComponentAnswers();
+        this.block.setAnswer(childAnswers, false);
```
alextanlan, line 155: Update the data model.

aglushko-broad (approved):

PR #876, opened 2021-09-02
DDP-6927: Updates to `About Us` page [Pancan]

alextanlan:
Applied updates, mentioned in [DDP-6927](https://broadinstitute.atlassian.net/browse/DDP-6927)


ddp-workspace/projects/ddp-pancan/src/app/components/page-with-sections/page-with-sections.component.scss
```diff
@@ -5,6 +5,7 @@
 
     &.about-us {
         .section-0 {
+            min-height: 705px;
```
alextanlan: In order to the background image fits the updated section (amount of text changed).

aglushko-broad: maybe background-size: contain; would be a better choice here: we won't depend on the text size

aglushko-broad: fixed

aglushko-broad (approved):

PR #869, opened 2021-08-30
DDP-6918 [Pancan] Fix autocomplete scrolling

alextanlan:
Fix a case when an autocomplete list is scrolled above the header.

**_Before:_**
image

**_After:_**
image

PR #878, opened 2021-09-03
DDP-6918: Fix Autocomplete list scrolling [Pancan] 

alextanlan:
The previous fix [PR#869](https://github.com/broadinstitute/ddp-angular/pull/869) causes to another non caught bug. 
We put autocomplete scroll under app header by scrolling. However other components (e.g. language selector) uses the same overlay container and also appear under the header. 

_See screenshots:_ (currently on dev, after the previous fix)
image
image

It is quite tough (if possible?) to tweak it by css. And it is a [known Angular materials issue](https://github.com/angular/components/issues/1432).

So it should be done in another way. 
The simplest one is just closing any autocomplete suggestions list like it was done firstly (0171b33b052ea6806f296abc3ff2a9f5b5fd7079). 
On the other hand it's not a user friendly behavior.
Thus I have fixed it a bit tricky, but it is working. 

_See screenshots after the fix:_
image
image

### DIFF
--- a/ddp-workspace/projects/ddp-pancan/src/app/components/header/header.component.scss
+++ b/ddp-workspace/projects/ddp-pancan/src/app/components/header/header.component.scss
@@ -5,7 +5,7 @@
     top: 0;
     left: 0;
     width: 100%;
-    z-index: 2000;
+    z-index: 1000;
     height: $header-height;
     transition: all 200ms linear;
     background-color: white;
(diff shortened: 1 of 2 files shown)


alextanlan:
Discussed with Marco that it is a corner case.
Not worth complicating code in order to fix rare cases. 
Closing the PR.

PR #879, opened 2021-09-03
DDP-6908 Fixed typos, changed event detection for AddressInputService

PavelYakupov:


PavelYakupov:
@MocanaAtBroad I tested that component, it's ok for my opinion.

aleksey-kuznetsov:
@MocanaAtBroad just a quick note about this fix: address form fields would show asterisk (indicating that this field is required) only after focus event on an input. Emitting an event fixes this problem.

aleksey-kuznetsov (approved):

PR #884, opened 2021-09-06
Fix default value for date question answer

aleksey-kuznetsov:
Background: if user tried to clear out pre-populated value in date question or type something invalid, save request would fail with 400 and user would be redirected to `/error` page.
With this change a user can freely clear answers and get no error if a given answer is invalid but a red validation input underline.

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activityDateAnswer.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activityDateAnswer.component.ts
@@ -33,7 +33,7 @@ export class ActivityDateAnswer {
     constructor(private logger: LoggingService) { }
 
     public handleChange(value: DatePickerValue | null): void {
-        let dateValue: DatePickerValue | null = {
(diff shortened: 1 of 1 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activityDateAnswer.component.ts
```diff
@@ -33,7 +33,11 @@ export class ActivityDateAnswer {
     constructor(private logger: LoggingService) { }
 
     public handleChange(value: DatePickerValue | null): void {
-        let dateValue: DatePickerValue | null = null;
+        let dateValue: DatePickerValue | null = {
```
aglushko-broad: `let dateValue: DatePickerValue = {...};`
can't be a null type anymore

aleksey-kuznetsov: Correct! I'll fix that.

aglushko-broad (approved):

PR #887, opened 2021-09-09
DDP-6938 [A-T] Added disabling for Add a Participant button while wai…

PavelYakupov:
…ting for the participant to be created


MocanaAtBroad (approved):
This looks good! If you have not done so, I would inspect code in A-T for other  places that is possible to emit duplicate requests with consecutive clicks.

PR #888, opened 2021-09-09
DDP-6472 clean up participant in the session right after we build the url for activities request

aglushko-broad:


ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/userActivityServiceAgent.service.ts
```diff
@@ -21,11 +21,18 @@ export class UserActivityServiceAgent extends UserServiceAgent<Array<ActivityIns
...
+        return studyGuid.pipe(mergeMap(x => {
+            if (x) {
+                if (participantGuid) {
+                    this.session.setParticipant(participantGuid);
+                }
+                const activitiesObservable = this.getObservable(`/studies/${x}/activities`);
+                if (participantGuid) {
+                    this.session.setParticipant(null);
+                }
+                return activitiesObservable;
```
MocanaAtBroad, line 38: I don't think this is good place to change any non-local state, particularly on the session, which is kind of global. Please find an alternative fix to problem.

MocanaAtBroad (changes requested):

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/userActivityServiceAgent.service.ts
MocanaAtBroad: public getActivities(studyGuid: Observable<string | null>, participantGuid?: string): Observable<Array<ActivityInstance> | null> {
        return studyGuid.pipe(mergeMap(x => {
            if (x) {
                const previousParticipantGuid = this.session.participantGuid;
                if (participantGuid) {
                    this.session.setParticipantGuid(participantGuid);
                }
                try {
                const activitiesObservable = this.getObservable(`/studies/${x}/activities`);
               } finally {
                 this.session.setPartcipantGuid(previousParticipantGuid);
                }
                return activitiesObservable;
}
            return of(null);
        }));

aglushko-broad: done! thanks for the help!

MocanaAtBroad (approved):

PR #907, opened 2021-09-20
Pancan: styling support for composite grid questions

aleksey-kuznetsov:
This PR adds styles to arrange radio picklist questions in a grid using composite question as a group.
On mobile questions are arranged like before in a list one by one.

### DIFF
--- a/ddp-workspace/projects/ddp-pancan/src/styles/activities.scss
+++ b/ddp-workspace/projects/ddp-pancan/src/styles/activities.scss
@@ -512,6 +512,7 @@ select:disabled {
                                 .mat-radio-group {
                                     display: flex !important;
                                     flex-flow: row nowrap !important;
+                                    align-items: center;
 
                                     height: 100%;
 
@@ -528,6 +529,8 @@ select:disabled {
                                                 margin: 0;
 
                                                 .mat-radio-label-content {
+                                                    display: none;
+
                                                     padding: 0;
 
                                                     font-size: 1rem;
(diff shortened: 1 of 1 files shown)


ddp-workspace/projects/ddp-pancan/src/styles/activities.scss
```diff
@@ -473,4 +473,142 @@ select:disabled {
...
+    [class*="composite-answer"][class*="GRID"] {
+        .ddp-answer-container {
+            margin-bottom: 2.5rem;
+
+            .ddp-answer-field {
+                display: block;
+
+                ddp-activity-picklist-answer {
+                    display: flex;
+
```
aglushko-broad, line 486: #minor not sure we need to have empty lines between styles for the same selector, same for other places

ddp-workspace/projects/ddp-pancan/src/styles/activities.scss
```diff
@@ -473,4 +473,142 @@ select:disabled {
     }
 }
 
-
+@media (min-width: 768px) {
```
aglushko-broad: you can use variable $tablet-breakpoint

ddp-workspace/projects/ddp-pancan/src/styles/activities.scss
```diff
@@ -473,4 +473,142 @@ select:disabled {
...
+
+                ddp-activity-picklist-answer {
+                    display: flex;
+
+                    padding: 1rem 0;
+
+                    ddp-question-prompt {
+                        width: 10rem;
+
+                        > div {
```
aglushko-broad: can we skip this part of nesting and use 
```
ddp-question-prompt {
    width: 10rem;

    .ddp-question-prompt {
        ....
    }
}
```
same for all other places with `> div`

I see that the styling is not straightforward here but would be good to try to decrease the level of nesting in general everywhere

ddp-workspace/projects/ddp-pancan/src/styles/activities.scss
aleksey-kuznetsov: I just try to group css properties and separate them with empty lines.

ddp-workspace/projects/ddp-pancan/src/styles/activities.scss
aleksey-kuznetsov: Thank you! I'll fix that.

ddp-workspace/projects/ddp-pancan/src/styles/activities.scss
aleksey-kuznetsov: I tried to replicate the DOM structure for this type of question in CSS so that it's easier to style each possible DOM node.

aglushko-broad: that's interesting approach :) I'd like to share that very big nesting leads to maintanability issues, performance issues and too high specificity of styles which can be unpredicatble to work with/overwrite

aglushko-broad (approved):

PR #910, opened 2021-09-23
DDP-6992 error logging when no temp user at signup

aglushko-broad:


ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
```diff
@@ -142,6 +142,9 @@ export class Auth0AdapterService implements OnDestroy {
      */
     public signup(additionalParams?: Record<string, string>): void {
         const temporarySession = this.session.isTemporarySession() ? this.session.session : null;
+        if (!temporarySession || !temporarySession.userGuid) {
+            this.log.logError(`${this.LOG_SOURCE}.signup.No temporal user guid`);
+        }
         const params = {
```
MocanaAtBroad, line 148: Could you also a logInfo with the contents of `params`?

MocanaAtBroad (approved):

ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
aglushko-broad: loginfo data won't be saved at google cloud logs. Is it what we want? We want be able to get this information if the same errir happens again for some user

PR #912, opened 2021-09-29
DDP-6897 add ability to edit mailing address from dashboard

aglushko-broad:
image
image

### DIFF
--- a/ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard-redesigned.component.ts
+++ b/ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard-redesigned.component.ts
@@ -15,12 +15,12 @@ import {
     UserProfileServiceAgent,
     WorkflowServiceAgent,
     UserManagementServiceAgent,
-    LoggingService
+    LoggingService,
+    UserPreferencesComponent
 } from 'ddp-sdk';
 import { map, take, filter, switchMap, tap, mergeMap, finalize, catchError, mapTo } from 'rxjs/operators';
 import { combineLatest, forkJoin, Observable, of, Subject } from 'rxjs';
 import { TranslateService } from '@ngx-translate/core';
-import { UserPreferencesComponent } from '../../../../../ddp-sdk/src/lib/components/user/userPreferences.component';
 import { MatDialog } from '@angular/material/dialog';
 
 interface DashboardParticipant {
@@ -139,6 +139,9 @@ interface DashboardParticipant {
     styles: [`
         .full-width {
             width: 100%;
+        }
+        .mat-expansion-panel-header-title {
+            align-items: center;
         }
     `]
 })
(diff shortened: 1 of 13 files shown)


ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard-redesigned.component.ts
```diff
@@ -20,6 +20,8 @@ import {
 import { map, take, filter, switchMap, tap, mergeMap, finalize, catchError, mapTo } from 'rxjs/operators';
 import { combineLatest, forkJoin, Observable, of, Subject } from 'rxjs';
 import { TranslateService } from '@ngx-translate/core';
+import { UserPreferencesComponent } from '../../../../../ddp-sdk/src/lib/components/user/userPreferences.component';
```
MocanaAtBroad:
```suggestion
import { UserPreferencesComponent } from 'ddp-sdk';
```

MocanaAtBroad:
We probably should resize things so if there is a Suggestion from EasyPost, its display is easily visible:
Screen Shot 2021-09-30 at 11 13 43 AM

MocanaAtBroad:
(Let's not forget styling the buttons either.

MocanaAtBroad:
Also, we should play a little bit with indentation/font size/maybe even background color to establish a little more contrast between the dialog title and the mail address component. Maybe Erin can provide some help here.
Screen Shot 2021-09-30 at 11 22 24 AM
.

MocanaAtBroad:
Alignment seems off in mobile (I don't think you have had much of a chance to do styling but flagging it here for now
Screen Shot 2021-09-30 at 11 25 43 AM
)

MocanaAtBroad (commented):
Functionality looks good. Need to nail down styling. As I mentioned in Slack, the position of the "Edit" button seems confusing to me/doesn't give any indication of what will be edited. Sizing and styling need to be straightened out too.

ddp-workspace/projects/toolkit/src/lib/components/dashboard/dashboard-redesigned.component.ts
aglushko-broad: thanks! done

aglushko-broad:
> We probably should resize things so if there is a Suggestion from EasyPost, its display is easily visible: Screen Shot 2021-09-30 at 11 13 43 AM

resize wouldn't help mich here, though I increased the width in 100px. I implemented scrolling to the bottom once suggestion or form error is appeared

aglushko-broad:
> (Let's not forget styling the buttons either.

done!

aglushko-broad:
> Alignment seems off in mobile (I don't think you have had much of a chance to do styling but flagging it here for now Screen Shot 2021-09-30 at 11 25 43 AM )

thanks! fixed

MocanaAtBroad (approved):

PR #924, opened 2021-10-19
Initial commit of logging function

MocanaAtBroad:
Implements Google Cloud Function that allows for browser to write to GCP Logging Service.
Initial version deployed to dev environment. A line the one below from Curl should reach the service.
```bash
curl -v -X POST  https://us-central1-broad-ddp-dev.cloudfunctions.net/LoggingService  -H "Content-Type:application/json"  -d '{"logName": "angular-THESTUDY", "severity": "INFO","textPayload": "A test of the real deployed thing!","httpRequest": {"requestUrl": "http://dotdot.com", "userAgent": "An AgentStringFrom the client"}, "labels": {"tempUserGuid": "XXXXX"}}'
```
The command to build locally:
```bash
npm run build
```
The command used to deploy:
```bash
gcloud functions deploy LoggingService --entry-point logMessage --runtime nodejs12  --trigger-http --project broad-ddp-dev
```


logging-cloud-function/src/model/index.ts
```diff
@@ -0,0 +1,37 @@
...
+            return;
+        }
+        log.write(new Entry(entry.toGoogleEntryMetadata()))
+            .catch((error: any) => {
+                console.error("Could not write to log: %o", error);
+        });
+        res.status(204).send();
+
+    }
+
```
MocanaAtBroad: All there is to the function. Complexity is in calling the API correctly.

logging-cloud-function/src/model/logEntry.ts
```diff
@@ -0,0 +1,42 @@
...
+
+
+export class LogEntry {
+    logName: string;
+    severity: LogSeverity;
+    textPayload: string | Object;
+    trace?: string;
+    httpRequest?: LogEntryHttpRequest;
+    labels?: Record<string,string>;
+
```
MocanaAtBroad, line 12: This a subset of the fields that can be passed for a log entry. Wanted to pare it down to stuff that would be useful.

logging-cloud-function/src/model/index.ts
```diff
@@ -0,0 +1,37 @@
...
+const {Logging} = require('@google-cloud/logging');
+const DEFAULT_PROJECT_ID ='broad-ddp-dev';
+
+const projectId = process.env.GCP_PROJECT ? process.env.GCP_PROJECT : DEFAULT_PROJECT_ID;
+
+// Creates a client
+const logging = new Logging({projectId});
+
+// Selects the log to write to
+const log = logging.log('something');
```
aglushko-broad: should we use smth more specific here, e.g. 'nodejs-client'

aglushko-broad (dismissed):

logging-cloud-function/src/model/index.ts
MocanaAtBroad: How about loggingClient?

aglushko-broad (dismissed):

PR #926, opened 2021-10-19
DDP-7060 Fixed reload issue

PavelYakupov:


ddp-workspace/projects/ddp-singular/src/app/constants/route.ts
```diff
@@ -6,7 +6,7 @@ export enum Route {
   About = 'about',
   Login = 'login',
   Error = 'error',
-  Survey = 'survey',
+  Survey = 'survey/:id',
```
aleksey-kuznetsov, line 9: It's better to set this url to `activity/:id` since all reminder emails that contain links to associated activities have this in their `href` attribute: `{{ baseWebUrl }}/activity/:id`

aleksey-kuznetsov (changes requested):

ddp-workspace/projects/ddp-singular/src/app/constants/route.ts
aleksey-kuznetsov: Nevermind, I saw that we already have `activity/:id` route in our application. I'd only suggest to clean everything up and have only `activity/:id` route for filling activities.

aleksey-kuznetsov (approved):

PR #927, opened 2021-10-19
DDP-7053 integrate nodejs logging service into angular logging service

aglushko-broad:


ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
```diff
@@ -36,4 +41,24 @@ export class LoggingService {
     private stringify(obj: object): string {
         return Object.keys(obj).map(key => `${key}: ${obj[key]}`).join(', ');
     }
+
+    public logToCloud(payload: string, labels?: {[key: string]: string}): void {
+        const url = `${this.config.cloudLoggingUrl}/LoggingService`;
```
MocanaAtBroad: Let's move the "/LoggingService" to be part of the cloudLoggingUrl value. It is pretty arbritary/can change.

ddp-workspace/projects/ddp-pancan/src/app/app.module.ts
```diff
@@ -71,6 +71,7 @@ toolkitConfig.dashboardDisplayedColumns = ['name', 'summary', 'status', 'actions
 
 export const sdkConfig = new ConfigurationService();
 sdkConfig.backendUrl = DDP_ENV.basePepperUrl;
+sdkConfig.cloudLoggingUrl = 'https://us-central1-broad-ddp-dev.cloudfunctions.net';
```
ssettipalli: Don't this vary by env like dev/test/staging/prod !! 
May be it should come from DDP_ENV like other values that depend on Env !!

ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
```diff
@@ -36,4 +41,24 @@ export class LoggingService {
...
+        const body = {
+            logName: `angular-${this.config.studyGuid}`,
+            severity: 'INFO',
+            textPayload: payload,
+            labels
+        };
```
MocanaAtBroad, line 59: Let's add requestURL (whatever the current URL is) and the User Agent (I think it obtainable via browser API).
curl -v -X POST  https://us-central1-broad-ddp-dev.cloudfunctions.net/LoggingService  -H "Content-Type:application/json"  -d '{"logName": "angular-THESTUDY", "severity": "INFO","textPayload": "A test of the real deployed thing!","httpRequest": {"requestUrl": "http://dotdot.com", "userAgent": "An AgentStringFrom the client"}, "labels": {"tempUserGuid": "XXXXX"}}'

MocanaAtBroad: Also, let the severity be argument/we don't have to hard-code severity. Check the model for acceptable values:
https://github.com/broadinstitute/ddp-angular/tree/8d49f8e756b4dc51566eeb423a859a80fdfa4537/logging-cloud-function/src/model

config/pepperConfig.js.ctmpl
```diff
@@ -11,11 +11,13 @@ var DDP_ENV = {
     mapsApiKey: "{{.Data.mapsApiKey}}",
     errorReportingApiKey: "{{.Data.errorReportingApiKey}}",
     projectGcpId: "{{.Data.googleProjectId}}",
+    cloudLoggingUrl:  "{{.Data.cloudLoggingUrl}}",
```
MocanaAtBroad, line 14: @aglushko-broad @ssettipalli I modified pepperConfig.js so we can get the cloud url from vault. I have modified vault in dev environment. We'll update vault for test, staging, etc when we deploy.

ddp-workspace/projects/ddp-pancan/src/app/app.module.ts
MocanaAtBroad: See commit and comment above.

ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
MocanaAtBroad: I would also recommend that we "hardcode" sending the user guid as a label using always the same label name (userGuid)?
Just so that, on the log side, we have a "standard" way to see and retrieve it. If we inject the session into this service, the user guid.

MocanaAtBroad (commented):
A few suggested changes. See if they make sense.

config/pepperConfig.js.ctmpl
aglushko-broad: thanks!

MocanaAtBroad (approved):
Looks good! Tried it and was logging the Auth0 stuff, which is what we need.
One thing that did not make it (I swear I thought I had made comment but I must have not saved it) is that I think we should extend `logDebug` `logWarning` and `logError` in `LoggerService` https://github.com/broadinstitute/ddp-angular/pull/927/files#diff-baa6cf9bf74f6989daf5da26f9a4c007e2d39041085b29ef3a9d73fccb851c96R16

to call `logToCloud` if we have the `doCloudLogging` flag in pepperConfig.js set to `true`.
This is not urgent/ can wait. But let's merge this baby as it is now and see if we can get to the bottom of the 🐛 .

PR #929, opened 2021-10-20
DDP-7072: FIx a bug when a user can register without entering diagnosis

alextanlan:
**_The idea_** - not validate composite question answers with null values.
And set null values explicitly if it is needed,
 e.g. in case when a user can enter/press the spacebar in the diagnosis field and it is accepted as input
(Pancan - Prequal)

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityQuestion.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityQuestion.component.ts
@@ -78,6 +78,7 @@ export class ActivityQuestionComponent implements OnInit, OnDestroy {
         ]).pipe(
             // not displaying any local validations until a validation is requested
             // except some unique cases
+            // TODO: try to fix/refactor the local validation globally as to avoid exceptions
             filter(([enteredValue, validationRequested]) => {
                 return (enteredValue && this.isExceptionToEnforceLocalValidation(this.block))
                     || !!validationRequested;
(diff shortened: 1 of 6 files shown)


MocanaAtBroad (commented):
The fix seems to work/not allow submission of prequal if cancer is missing.
But there is a sequence that does not quite work as expected:
1. Go to page where you enter the cancer.
2. Press Next leaving the cancer empty. An error message is displayed
3. Select or type in a cancer
4. **The error message does not go away** (though you can click on next)

By contrast, compare to next page: enter age of 90000 and try to Submit.
You are blocked to proceed and error message for age is displayed. Now change age to 50. The age error is cleared once age is changed.

could you look into why the picklist error is not automatically cleared?

alextanlan:
Marco, thanks for noticing! 

===

So, my thoughts about 
> The error message does not go away 
- it is another issue that we have in dev branch as well
- we used to skip a local validation for any questions before click on `Next`/`Prev` button in an activity. 
  Now we have a special case when we should force the local validation:
   an autocomplete picklist "cancer select" in Pancan study.
   
  Can you please take a look at 44548f204587a62fa412249e1c623e1a14bab12e.
  Seems like a crutch, but the case is out of our old validation logic, in my opinion.

MocanaAtBroad (dismissed):

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityQuestion.component.ts
```diff
@@ -76,12 +77,17 @@ export class ActivityQuestionComponent implements OnInit, OnDestroy {
             this.validationRequested$
         ]).pipe(
             // not displaying any local validations until a validation is requested
-            filter(([_, validationRequested]) => !!validationRequested),
+            // except some uniq cases
```
aglushko-broad: uniq -> unique

alextanlan: fixed

PR #932, opened 2021-10-22
DDP-7072: Fix a bug when a user can register without entering diagnosis

alextanlan:


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityQuestion.component.ts
```diff
@@ -69,14 +69,19 @@ export class ActivityQuestionComponent implements OnInit, OnDestroy {
...
             this.validationRequested$
         ]).pipe(
-            // not displaying any local validations until a validation is requested
-            filter(([_, validationRequested]) => !!validationRequested),
+            // not displaying any local validations until
+            // - either a validation is requested
+            // - or the previous answer was invalid and then the answer is changed
+            filter(([enteredNewValue, validationRequested]) => {
+                let isInvalidPreviousAnswer: boolean;
+                this.errorMessage$.subscribe(x => isInvalidPreviousAnswer = !!x);
```
aglushko-broad: I think I might know where the problem is. 
 When click Next we set validationRequested first to true and when right after that to false even if validation state is false.
Open ActivityComponent and check incrementStep function in the second tap the line `this.validationRequested = false;` needs to me moved inside the condition block `if (this.currentSection.valid) {`

aglushko-broad: this also will fix the same validation issue with checkboxes "I've been diagnosed/my child's been diagnosed" on the very first step

alextanlan: Good catch, thank you! 
Seems reasonable for me.
Fixed in 3baa927e5a1c1273183e7005715a47ff4bfd933a.

aglushko-broad (approved):

PR #935, opened 2021-10-27
DDP-7082: Refactor the required validation rule

alextanlan:
Moved the a question required validation logic into the question itself. 
Now `ActivityRequiredValidationRule` has no need to import specific question types.

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityAgreementQuestionBlock.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityAgreementQuestionBlock.ts
@@ -13,4 +13,8 @@ export class ActivityAgreementQuestionBlock extends ActivityQuestionBlock<boolea
     public canPatch(): boolean {
         return true;
     }
+
+    public hasRequiredAnswer(): boolean {
+        return this.answer === true;
+    }
 }
--- a/ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityAgreementQuestionBlock.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityAgreementQuestionBlock.ts
@@ -14,7 +14,7 @@ export class ActivityAgreementQuestionBlock extends ActivityQuestionBlock<boolea
         return true;
     }
 
-    public hasRequiredAnswer(): boolean {
+    public hasAnswer(): boolean {
         return this.answer === true;
     }
 }
--- a/ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityCompositeQuestionBlock.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityCompositeQuestionBlock.ts
@@ -25,7 +25,10 @@ export class ActivityCompositeQuestionBlock extends ActivityQuestionBlock<Answer
     }
 
     public hasRequiredAnswer(): boolean {
-        return this.answer != null
-            && this.children.every(childQuestion => childQuestion.hasRequiredAnswer());
+        const requiredChildren = this.children.filter(child => child.isRequired);
+
+        return requiredChildren ?
+            requiredChildren.every(childQuestion => childQuestion.hasRequiredAnswer())
+            : !!this.answer?.length;
     }
 }
--- a/ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityCompositeQuestionBlock.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityCompositeQuestionBlock.ts
@@ -23,4 +23,9 @@ export class ActivityCompositeQuestionBlock extends ActivityQuestionBlock<Answer
     public get questionType(): QuestionType {
         return QuestionType.Composite;
     }
+
+    public hasRequiredAnswer(): boolean {
+        return this.answer != null
+            && this.children.every(childQuestion => childQuestion.hasRequiredAnswer());
+    }
 }
--- a/ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityCompositeQuestionBlock.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityCompositeQuestionBlock.ts
@@ -24,11 +24,11 @@ export class ActivityCompositeQuestionBlock extends ActivityQuestionBlock<Answer
         return QuestionType.Composite;
     }
 
-    public hasRequiredAnswer(): boolean {
+    public hasAnswer(): boolean {
         const requiredChildren = this.children.filter(child => child.isRequired);
 
         return requiredChildren ?
-            requiredChildren.every(childQuestion => childQuestion.hasRequiredAnswer())
+            requiredChildren.every(childQuestion => childQuestion.hasAnswer())
             : !!this.answer?.length;
     }
 }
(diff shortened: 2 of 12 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityCompositeQuestionBlock.ts
```diff
@@ -23,4 +23,9 @@ export class ActivityCompositeQuestionBlock extends ActivityQuestionBlock<Answer
     public get questionType(): QuestionType {
         return QuestionType.Composite;
     }
+
+    public hasRequiredAnswer(): boolean {
+        return this.answer != null
+            && this.children.every(childQuestion => childQuestion.hasRequiredAnswer());
```
alextanlan: It is just my guess. 
Marco, am I right that if a composite question has a required validator 
then it is valid when every child question of it matches their own required validation rule?

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityDateQuestionBlock.ts
```diff
@@ -62,4 +62,8 @@ export class ActivityDateQuestionBlock extends ActivityQuestionBlock<DatePickerV
     public convertToString(value: DatePickerValue): string {
         return `day:${value.day || ''};month:${value.month || ''};year:${value.year || ''}`;
     }
+
+    public hasRequiredAnswer(): boolean {
+        return (this.answer != null) && this.isSpecifiedFieldsPresent(this.answer);
```
aglushko-broad, line 67: wondering if we want to use here `this..isSpecifiedFieldsPresent(this.answer) || this.hasRequiredFields(this.answer)`
based on the task
https://broadinstitute.atlassian.net/browse/DDP-4600
sometimes some fields might be optional

aglushko-broad (dismissed):

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityCompositeQuestionBlock.ts
MocanaAtBroad: OK. I interestingly, I am not seeing no the server side an implementation of required rule that works on composites!

But let's define it here.
If we say a composite question is required, what we are saying is that we don't want the user to leave the question unanswered.
So that tells us at least #1: **that there would be at least one row of answers for a composite question.**

Now currently as far as I understand (and hope!) it is not possible to submit an empty row of composite answers, e.g., if a composite question is composed of NAME and DOB, it is possible to submit answer that contains only NAME, only DOB, both NAME and DOB, but not a row with no answers. You can "Delete" a row by PATCHing all the rows without the deleted value.

OK. So then, what does it mean if, say NAME in example above is required but not DOB?
I think it means that if there is a row of composite question answers, we are expecting that the row will have a NAME value but it would be OK to leave DOB blank.

So a missing NAME in a row should prevent the activity SUBMIT/final activity PUT. And this would be the case whether or not the parent composite question is required.

**So: a required composite means that we have at least one row of answers.
And a required child of a composite means that, if there is a row of answers, there needs to be a value for that child question.**

Does this makes sense to you? Are we missing some detail here?

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityAgreementQuestionBlock.ts
```diff
@@ -13,4 +13,8 @@ export class ActivityAgreementQuestionBlock extends ActivityQuestionBlock<boolea
     public canPatch(): boolean {
         return true;
     }
+
+    public hasRequiredAnswer(): boolean {
```
MocanaAtBroad: The naming of the method suggests that an answer is always required (i.e., this method will return true if the agreement question was checked even though there is no required rule associated with it). 
I would suggest renaming to simply: `hasAnswer()` (or something else that addresses concern expressed above).

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityRequiredValidationRule.ts
```diff
@@ -1,50 +1,15 @@
...
-        } else if (this.question.questionType === QuestionType.File) {
-            const value = this.question.answer as ActivityFileAnswerDto;
-            valid = value && !!value.fileName && (value.fileSize != null);
-        }
+        const isValid = this.question.hasRequiredAnswer();
+        this.result = isValid ? null : this.message;
 
-        this.result = (valid ? null : this.message);
-        return valid;
+        return isValid;
```
MocanaAtBroad, line 13: Excellent! If new question is added to system, as long as `hasRequiredAnswer()` is implemented correctly for that question, the rule will work without modification!

MocanaAtBroad (commented):
I think this looks good to me (except probably the composite question. Comments in body). 
I think it is important to, in addition to tests you have added, that we validate the code changes against actual real existing activities that have required questions as a sanity check. I'll scan the projects to see if we can identify some....

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityRequiredValidationRule.ts
alextanlan: That was the idea, suggested by you.

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityAgreementQuestionBlock.ts
alextanlan: I suppose that the correct name would be like `hasValidAnswerInCaseThereIsRequiredValidationRule`.
`hasAnswer` is not quite clear in this case, in my opinion, 
i.e. a text question might have '   ' as an answer, but the answer is not valid for `Required` validator. 

May be, `hasValidAnswer` ?

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityDateQuestionBlock.ts
alextanlan: In scope of the improvement I have moved the validation logic from `activityRequiredValidationRule.ts` to any question itself. Nothing was changed in the validation. 
In my understanding, if a date question has a "mandatory" validation rule, it should have all fields (year, month, day) filled out.

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityAgreementQuestionBlock.ts
MocanaAtBroad: Well, `hasValidAnswer()` might suggest that it passes validation rules.
Other options: `answerHasBeenEntered()` ? `hasCompleteAnswer()` I think they are all going to be a bit ambiguous (ok `hasValidAnswerInCaseThereIsRequiredValidationRule()` is not ambiguous and I do like descriptive names but...)
Just make sure abstract definition has clear jsdoc explaining...

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityCompositeQuestionBlock.ts
alextanlan: Yes, sounds pretty logical. 
- if a composite question itself **has** a required validation rule, the question must have at least one answer.
- if a composite question itself **does not have** a required validation rule, but some of its children have - 
then those children questions must have valid answers (must pass their own validation rules for their answers). 

Fixed in 7844c4ba3f448d3b3ce921278bdf099be6922be2

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityAgreementQuestionBlock.ts
alextanlan: Ok, let's go with `hasAnswer` as the best compromise option.

Fixed in dbb57e6f61eabc3bfe2c2e66c56fbc39e43b3959

MocanaAtBroad (approved):
Looks good!

PR #938, opened 2021-11-01
Update block visibility in section when creating / deleting nested activity

aleksey-kuznetsov:
This PR adds functionality to recalculate section blocks visibility when nested activity instance is created or deleted.
Backend reference PR: https://github.com/broadinstitute/ddp-study-server/pull/1050.

11/15 update: add support for enabling nested activities via `enabled` flag on `ActivityBlock` [(DDP-7085)](https://broadinstitute.atlassian.net/browse/DDP-7085). Backend: https://github.com/broadinstitute/ddp-study-server/pull/1064.


alextanlan (dismissed):

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityBlock/activityBlock.component.ts
```diff
@@ -29,12 +29,14 @@ import { ModalActivityBlockComponent } from '../modalActivityBlock/modalActivity
 })
 export class ActivityBlockComponent implements OnInit, OnDestroy {
     @Input() block: ActivityActivityBlock;
+    @Input() enabled: boolean;
```
aglushko-broad, line 32: we introduced new input "enabled" in multiple components.But is it relly neccesary? Maybe we can reuse "readonly" input? Because in fact they are synonims and it's confusing when look at both of them about what's the responcibility area of each of them. My suggestions either use readonly input, either rename enabled input to smth more specific, either add a comment to distinguish between then

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/createActivityInstanceResponse.ts
```diff
@@ -0,0 +1,9 @@
+import { BlockVisibility } from './blockVisibility';
+
+export interface CreateActivityInstanceResponse {
```
aglushko-broad: my suggestion here would be:

interface NestedActivityInstance {
  // couple of words about this property
  blockVisibility?: BlockVisibility[];
}
export interface CreateActivityInstanceResponse extends ActivityInstanceGuid, NestedActivityInstance {}

public createInstance(): Observable<CreateActivityInstanceResponse | null> 
public deleteActivityInstance(): Observable<NestedActivityInstance | null> {

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityBlock/activityBlock.component.ts
```diff
@@ -63,8 +65,13 @@ export class ActivityBlockComponent implements OnInit, OnDestroy {
     createChildInstance(): void {
         this.activityServiceAgent.createInstance(this.studyGuid, this.block.activityCode, this.parentActivityInstanceGuid)
             .pipe(
-                concatMap((instanceGuid: ActivityInstanceGuid) => {
-                    return this.activityServiceAgent.getActivitySummary(this.studyGuid, instanceGuid.instanceGuid);
+                tap(response => {
+                    if (response?.blockVisibility) {
+                        this.blockVisibilityChanged.emit(response.blockVisibility);
```
aglushko-broad: should we emit this even if error was catched? if no - I'd suggest to move it to subscription (you can use forJoin operator for example to return multiple observables as a result in concatMap)

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activitySection.component.ts
```diff
@@ -71,6 +75,9 @@ export class ActivitySectionComponent implements OnInit, OnDestroy {
                         block.shown = element.shown;
                         blockVisibilityChanged = true;
                     }
+                    if (block.enabled !== element.enabled) {
+                        block.enabled = element.enabled;
```
aglushko-broad, line 79: should we rerender components as well as we do when "shown" is changed?

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/activityServiceAgent.service.ts
```diff
@@ -126,9 +131,12 @@ export class ActivityServiceAgent extends UserServiceAgent<any> {
             map(httpResponse => httpResponse));
     }
 
-    public deleteActivityInstance(studyGuid: string, activityGuid: string): Observable<any> {
+    public deleteActivityInstance(studyGuid: string, activityGuid: string): Observable<DeleteActivityInstanceResponse | null> {
         const baseUrl = this.getBaseUrl(studyGuid, activityGuid);
-        return this.deleteObservable(baseUrl, null, true);
+
+        return this.deleteObservable(baseUrl, null, true).pipe(
+            map(response => !!response ? response.body as DeleteActivityInstanceResponse : null)
```
aglushko-broad, line 138: #minor  map(response => response?.body as DeleteActivityInstanceResponse || null)

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityBlock/activityBlock.component.ts
aleksey-kuznetsov: I understand, it can be confusing at first. The main difference between `enabled` and `readonly` properties is that `enabled` is reevaluated (for each block) each time we make a `PATCH` request (just like with block visibility).

aglushko-broad: I see, that's clear now, thanks!

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityBlock/activityBlock.component.ts
aleksey-kuznetsov: I believe `tap` operator is not going to be executed if an error occurred and it will go straight to `catchError` operator. Am I wrong?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activitySection.component.ts
aleksey-kuznetsov: Correct! Thank you, I'll fix this part.

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/activityServiceAgent.service.ts
aleksey-kuznetsov: I just wanted to mirror how `createActivity` method handles this part.

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/createActivityInstanceResponse.ts
aleksey-kuznetsov: Wouldn't we get a TSLint error `no-empty-interface`?

aglushko-broad: we shouldn't. interface which extends multiple interfaces is new one

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityBlock/activityBlock.component.ts
aglushko-broad: depends on where the error is thrown. If createInstance method throws the error the tap won't be executed, but if getActivitySummary throws the error I believe we execture the tap (before), then go to catchError and then skipping subscription handler

aleksey-kuznetsov: Hmm, that's right, if first request succeeds and second one fails the `tap` operator will be executed and it might be confusing for users to see partial updates. I'll rearrange this so that block visibility is emitted if both requests were successful. Thank you!

aglushko-broad (approved):

PR #944, opened 2021-11-08
DDP-7103 picklist question for study builder (options editor w/o groups)

aglushko-broad:
image
image
image


ddp-workspace/projects/study-builder-ui/src/app/components/picklist-question-block/picklist-question-block.component.ts
```diff
@@ -0,0 +1,77 @@
...
+    ngOnInit(): void {
+        this.angularClientBlock$ = this.definitionBlock$.pipe(
+            tap(defBlock => console.log('getting defblock: %o', defBlock)),
+            filter(block => !!block),
+            map(defBlock => this.buildFromDef(defBlock)));
+    }
+
+    private buildFromDef(defBlock: QuestionBlockDef<PicklistQuestionDef>): ActivityPicklistQuestionBlock {
+        // todo replace mock options with custom options once options control would be added
+        const text1 = new SimpleTemplate(this.factory.createBlankTemplate());
```
alextanlan: Suggest extracting  options set logic (line 36-42) into a separate method, 
like `const options = this.getOptions()`.
What do you think?

ddp-workspace/projects/study-builder-ui/src/app/components/picklist-question-block/picklist-question-block.component.ts
```diff
@@ -0,0 +1,77 @@
...
+        const questionDef = defBlock.question;
+        newClientBlock.selectMode = questionDef.selectMode;
+        newClientBlock.renderMode = questionDef.renderMode;
+        newClientBlock.isRequired = questionDef.validations.some(val => val.ruleType === 'REQUIRED');
+        newClientBlock.stableId = questionDef.stableId;
+        newClientBlock.question = new SimpleTemplate(questionDef.promptTemplate).getTranslationText(this.config.defaultLanguageCode);
+        newClientBlock.picklistLabel =
+            new SimpleTemplate(questionDef.picklistLabelTemplate).getTranslationText(this.config.defaultLanguageCode);
+        newClientBlock.picklistOptions =
+            // questionDef.picklistOptions.map(picklistOption => this.convertOptionDefToActivityOption(picklistOption));
```
alextanlan: Is the commented line still needed?

ddp-workspace/projects/study-builder-ui/src/app/components/picklist-question-block/picklist-question-block.component.ts
```diff
@@ -0,0 +1,77 @@
...
+        text2.setTranslationText(this.config.defaultLanguageCode, 'test2');
+        const options: PicklistOptionDef[] =
+            [{stableId: 'test1', optionLabelTemplate: text1}, {stableId: 'test2', optionLabelTemplate: text2}];
+
+
+        const newClientBlock = new ActivityPicklistQuestionBlock();
+        const questionDef = defBlock.question;
+        newClientBlock.selectMode = questionDef.selectMode;
+        newClientBlock.renderMode = questionDef.renderMode;
+        newClientBlock.isRequired = questionDef.validations.some(val => val.ruleType === 'REQUIRED');
```
alextanlan, line 40: _**#optional**_
Perhaps, it is worth keeping all ruleTypes as an enum in order not to have lots of string values in code. 
(`'REQUIRED'` here)

ddp-workspace/projects/study-builder-ui/src/app/components/picklist-question-block/picklist-question-block.component.ts
aglushko-broad: I'm going to remove this mock logic in the next PR :)

alextanlan (dismissed):

ddp-workspace/projects/study-builder-ui/src/app/components/picklist-question-block/picklist-question-block.component.ts
aglushko-broad: that looks like you're checking the outdated version...

ddp-workspace/projects/study-builder-ui/src/app/components/picklist-question-block/picklist-question-block.component.ts
aglushko-broad: that can be a good place for improvement. I'll create enum in the next PR

MocanaAtBroad:
This looks really good! Seeing the picklist question change from a checkbox list to a dropdown to checkbox list with just single clicks is really cool. The dragging of the options to change the order also works very nicely!

MocanaAtBroad:
One improvement we should pursue here is to prevent or flag entry errors. I'll put them in separate comments
The stable ids for the options. For them to be "correct" they need to be unique within the question, i.e. if there is stableid "one" you can not have another option with stableid "one".
Don't think we need to do anything fancy to show the validation beyond whatever Material gives us . Turning one (or both. Does not matter red with a short message is plenty).

MocanaAtBroad:
Another improvement is to filter out "Render Modes" that are not supported by the selected "Select Mode", e.g., if "Multiple" is the "Select Mode", we should hide "Autocomplete list" as we only support selection one option. (just side comment: I was surprised that "Dropdown" supports  Multiple Select Mode!).
Do something reasonable if they do something like click "Single" , click "Autocomplete list" and then click "Multiple" (and therefore creating a combination we are saying is invalid). Just default to one of the Render Modes that work Single and Multiple. Or something else that makes sense.

MocanaAtBroad:
**We might have to bring over some translations to the study builder project?  (note the label the "Other" option in preview panel:**


Screen Shot 2021-11-16 at 6 07 35 PM
l)

aglushko-broad:
> Another improvement is to filter out "Render Modes" that are not supported by the selected "Select Mode", e.g., if "Multiple" is the "Select Mode", we should hide "Autocomplete list" as we only support selection one option. (just side comment: I was surprised that "Dropdown" supports Multiple Select Mode!). Do something reasonable if they do something like click "Single" , click "Autocomplete list" and then click "Multiple" (and therefore creating a combination we are saying is invalid). Just default to one of the Render Modes that work Single and Multiple. Or something else that makes sense.

I'm thinking about not filtering out but disabling select mode/render mode if selected render mode/select mode doesn't support it. If we need to disabled mode is currently selected we just unselect it.

aglushko-broad:
> **We might have to bring over some translations to the study builder project? (note the label the "Other" option in preview panel:**
> 
> Screen Shot 2021-11-16 at 6 07 35 PM l)

I tried it but in order internatialization factory has a dependency logging service which requires stackDriver which requires vault configuration...

MocanaAtBroad (approved):

PR #955, opened 2021-11-20
Convert logToCloud to observable. Change calls to logToCloud

MocanaAtBroad:
We were getting a lot of errors trying to write to LoggingService cloud function.
Noticed that most (all?) seem to be calls before we would either be redirected to Auth0 to login or when login expired (I'll point to it in comments) we set window.location.reload() so my theory is that the logging service calls were getting interrupted when switching domains (I have seen some http status code 0 that suggest that).
So idea of change is that instead of just having to call logToCloud to execute you need to subscribe as function now returns an Observable.


ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
```diff
@@ -134,8 +134,9 @@ export class Auth0AdapterService implements OnDestroy {
             }),
             language: this.language.getCurrentLanguage()
         };
-        this.log.logToCloud(`Auth0 login modal is open: ${JSON.stringify(params)}`, { auth0Mode: Auth0Mode.LoginOnly });
-        this.showAuth0Modal(Auth0Mode.LoginOnly, params);
+        this.log.logToCloud(`Auth0 login modal is open: ${JSON.stringify(params)}`, { auth0Mode: Auth0Mode.LoginOnly })
+            .subscribe(() => this.showAuth0Modal(Auth0Mode.LoginOnly, params)
```
MocanaAtBroad: I think I got all the calls. This is representative of change: before you just call the method. Now you need to subscribe.
And this is a good example of the problem I was describing. Imagine the original version: it initiates an http call to logging service and the request is in flight. 
But because it app is asynchronous, while waiting for response to server, we go to "showAuth0Modal" which results in redirecting to Auth0 domain to display its dialog.
So if the logging service call is not done by time of redirect, the browser will cancel the original call and (probably) generating an error.

ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
```diff
@@ -158,8 +159,8 @@ export class Auth0AdapterService implements OnDestroy {
             // @todo : hack delete when done
             serverUrl: this.configuration.backendUrl
         };
-        this.log.logToCloud(`Auth0 signup modal is open for user: ${JSON.stringify(params)}`, { auth0Mode: Auth0Mode.SignupOnly });
-        this.showAuth0Modal(Auth0Mode.SignupOnly, params);
+        this.log.logToCloud(`Auth0 signup modal is open for user: ${JSON.stringify(params)}`, { auth0Mode: Auth0Mode.SignupOnly })
+            .subscribe(() => this.showAuth0Modal(Auth0Mode.SignupOnly, params));
```
MocanaAtBroad: Same as above.

ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
```diff
@@ -180,7 +181,8 @@ export class Auth0AdapterService implements OnDestroy {
                 this.windowRef.nativeWindow.location.hash = '';
                 this.setSession(authResult);
                 this.log.logEvent(`${this.LOG_SOURCE}.handleAuthentication`, authResult);
-                this.log.logToCloud(`${this.LOG_SOURCE}.handleAuthentication, authResult: ${JSON.stringify(authResult)}`);
+                this.log.logToCloud(`${this.LOG_SOURCE}.handleAuthentication, authResult: ${JSON.stringify(authResult)}`)
+                    .subscribe();
```
MocanaAtBroad: In this portion of code I am not worried about a redirect, so just .subscribe() right away.

ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
```diff
@@ -238,14 +240,14 @@ export class Auth0AdapterService implements OnDestroy {
     public setSession(authResult, isAdmin: boolean = false): void {
         const decodedJwt = this.jwtHelper.decodeToken(authResult.idToken);
         this.log.logEvent(this.LOG_SOURCE, `authResult: ${decodedJwt}`);
-        this.log.logToCloud(`${this.LOG_SOURCE} authResult: ${JSON.stringify(decodedJwt)}`);
+        this.log.logToCloud(`${this.LOG_SOURCE} authResult: ${JSON.stringify(decodedJwt)}`).subscribe();
```
MocanaAtBroad: Same as above

ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
```diff
@@ -263,33 +265,34 @@ export class Auth0AdapterService implements OnDestroy {
...
 
     public logout(returnToUrl: string = ''): void {
         const baseUrl = this.configuration.baseUrl;
-        this.log.logToCloud(`${this.LOG_SOURCE} logout for user`);
-        // Remove tokens and expiry time from localStorage
-        this.session.clear();
-        this.log.logEvent(this.LOG_SOURCE, 'logout');
-        this.analytics.emitCustomEvent(AnalyticsEventCategories.Authentication, AnalyticsEventActions.Logout);
+        this.log.logToCloud(`${this.LOG_SOURCE} logout for user`).subscribe(() => {
+            // Remove tokens and expiry time from localStorage
```
MocanaAtBroad, line 275: Not sure if needed here to subscribe, but can't hurt.

ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
```diff
@@ -353,9 +356,11 @@ export class Auth0AdapterService implements OnDestroy {
     }
 
     private handleExpiredTemporarySession(): void {
-        this.log.logToCloud(`${this.LOG_SOURCE} expired temporary session`);
-        this.session.clear();
-        window.location.reload();
+        this.log.logToCloud(`${this.LOG_SOURCE} expired temporary session`).subscribe(() => {
+                this.session.clear();
+                window.location.reload();
```
MocanaAtBroad, line 362: This is the other case that I worried about. a location.reload will not result in changing domains but (maybe) could result in interrupting logToCloud in flight request.

ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
```diff
@@ -44,9 +45,9 @@ export class LoggingService {
         return Object.keys(obj).map(key => `${key}: ${obj[key]}`).join(', ');
     }
 
-    public logToCloud(payload: string, labels?: {[key: string]: string}, severity = 'INFO'): void {
+    public logToCloud(payload: string, labels?: {[key: string]: string}, severity = 'INFO'): Observable<void> {
```
MocanaAtBroad, line 47: We changed the signature. Observable!

ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
```diff
@@ -58,14 +59,15 @@ export class LoggingService {
...
             url,
             body,
             { headers: new HttpHeaders({ 'Content-Type': 'application/json' }) }).pipe(
             catchError((error: any) => {
                 this.logError(this.LOG_SOURCE, `HTTP POST: ${url}. Error:`, error);
                 return of(null);
-            })
-        ).subscribe();
+            }),
+            map(() => void 0)
```
MocanaAtBroad, line 69: Learned that this is how you  create a function that returns void!

ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
```diff
@@ -12,6 +12,7 @@ type Logger = (message?: any, ...optionalParams: any[]) => void;
 @Injectable()
 export class LoggingService {
     private readonly LOG_SOURCE = 'Logging';
+    hello: Observable<void> = of();
```
alextanlan: I guess, that was for debug purposes only.

ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
```diff
@@ -134,8 +134,9 @@ export class Auth0AdapterService implements OnDestroy {
             }),
             language: this.language.getCurrentLanguage()
         };
-        this.log.logToCloud(`Auth0 login modal is open: ${JSON.stringify(params)}`, { auth0Mode: Auth0Mode.LoginOnly });
-        this.showAuth0Modal(Auth0Mode.LoginOnly, params);
+        this.log.logToCloud(`Auth0 login modal is open: ${JSON.stringify(params)}`, { auth0Mode: Auth0Mode.LoginOnly })
```
alextanlan, line 137: Suggest adding `.pipe(takeUntil(this.ngUnsubscribe))` here and for other subscriptions.

ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
alextanlan: It is a good catch!  

Just wondering whether using a `webworker` would help us to make the logging calls in a separate tread. 
But please don't go so crazy, it can be just an idea for research time. )

ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
```diff
@@ -44,9 +45,9 @@ export class LoggingService {
         return Object.keys(obj).map(key => `${key}: ${obj[key]}`).join(', ');
     }
 
-    public logToCloud(payload: string, labels?: {[key: string]: string}, severity = 'INFO'): void {
+    public logToCloud(payload: string, labels?: {[key: string]: string}, severity = 'INFO'): Observable<void> {
         if (!this.config.doCloudLogging) {
-            return;
+            return of();
```
alextanlan: Seems, this (`of()`) will never emit. The page is becoming "frozen".
I believe, using `of(null)` would be proper there.

alextanlan: P.S. After debugging I see that `of()` completes the Observable even though [Rxjs documentation](https://rxjs.dev/api/index/function/of) teels us it shoud behave like `Never`

image

ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
MocanaAtBroad: Interesting idea! I am going to postpone pursuing that for now, but we might come back to this. Have had the sense that webworker could have a role in our apps and this might be it. But for now I just want to get rid fo the errors.

ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
MocanaAtBroad: Ah! Garbage made it into the commit (was trying to figure out what type of() returns). Removed.

ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
```diff
@@ -44,9 +45,9 @@ export class LoggingService {
         return Object.keys(obj).map(key => `${key}: ${obj[key]}`).join(', ');
     }
 
-    public logToCloud(payload: string, labels?: {[key: string]: string}, severity = 'INFO'): void {
+    public logToCloud(payload: string, labels?: {[key: string]: string}, severity = 'INFO'): Observable<void> {
         if (!this.config.doCloudLogging) {
```
alextanlan: I would also add check `!this.config.cloudLoggingUrl` here. 

Because when I was debugging the logic I caught a corner case when url was undefined and it was not easy to get what is wrong.

alextanlan: created  [PR#956](https://github.com/broadinstitute/ddp-angular/pull/956) to fix that ^

ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
MocanaAtBroad: Interesting!  Found same problem. I think I fixed this now.
bce5f34e028a5e449101e7d37f61baf52c85e266

ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
MocanaAtBroad: ef13432b0

ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
MocanaAtBroad: Went around and added .pipe(take(1)) everywhere, which I think now that I think about, is probably a better way to do it (everytime we subscribe we unsubscribe immediately).
cbb033159c64faf427db8e73c65dc00401a09d9a

ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
MocanaAtBroad: 1190fe8b1fb47ead674ad5e4c9bee8d8b544602e

ddp-workspace/projects/ddp-pancan/src/app/app.module.ts
```diff
@@ -90,8 +90,8 @@ sdkConfig.projectGAToken = DDP_ENV.projectGAToken;
 sdkConfig.errorReportingApiKey = DDP_ENV.errorReportingApiKey;
 sdkConfig.projectGcpId = DDP_ENV.projectGcpId;
 sdkConfig.doGcpErrorReporting = DDP_ENV.doGcpErrorReporting;
+sdkConfig.doCloudLogging = DDP_ENV.doCloudLogging;
 sdkConfig.cloudLoggingUrl = DDP_ENV.cloudLoggingUrl;
-sdkConfig.doCloudLogging = DDP_ENV.doGcpErrorReporting;
 sdkConfig.tooltipIconUrl = 'assets/images/info.png';
```
MocanaAtBroad, line 95: Notice we were not setting these correctly in pancan.

alextanlan (approved):

aglushko-broad:
that's a great catch! thanks!

aglushko-broad (approved):

PR #960, opened 2021-11-23
DDP-7103 add i18n for study builder

aglushko-broad:
add i18n for study builder UI


ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
```diff
@@ -69,41 +69,52 @@ export class Auth0AdapterService implements OnDestroy {
...
+                clientID: clientId,
+                studyGuid: this.configuration.studyGuid,
+                responseType: 'code',
+                audience,
+                scope: 'offline_access openid profile',
+                redirectUri: this.configuration.auth0CodeRedirect
+            });
+        } catch (e) {
+            console.error(e);
+            throw new Error(e);
```
MocanaAtBroad: Why was this necessary?

aglushko-broad: it wasn't but while I was working on i18n I faced again that old error about absence of adminClientId. The system throws  just a strange message about http interceptors in the console and that's it. Remember how we spent almost all day (or two days?) for pancan in the past and you realized it's related to adminClientId absence eventually and we agreed it would be better to improve the error logging in the future. So I decided to do it.

I can do it in a separate PR

aglushko-broad: created separate PR https://github.com/broadinstitute/ddp-angular/pull/966

alextanlan (approved):

PR #987, opened 2021-12-16
DDP-6203 Added success message for ActivityFileAnswer component

PavelYakupov:


PavelYakupov:
In order not to drag the property through the entire component tree from the client application, or in order not to depend on the config, I created a separate component that simply checks the support of the key in the language dictionary, and if it is, this feature becomes active. Clients without this key will remain unchanged.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer-success/activityFileAnswerSuccess.component.ts
```diff
@@ -0,0 +1,25 @@
...
+  constructor(
+    private readonly translateService: TranslateService
+  ) {}
+
+  showMessage(): boolean {
+    return this.hasTranslation(this.SUCCESS_KEY);
+  }
+
+  private hasTranslation(key: string): boolean {
+    const translation: string = this.translateService.instant(key);
```
aleksey-kuznetsov: I think we shouldn't use `instant` here since it won't reevaluate this when participant changes language.

MocanaAtBroad (dismissed):

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-file-answer-success/activityFileAnswerSuccess.component.ts
PavelYakupov: You are right. I've added this support.

aleksey-kuznetsov (approved):

PR #993, opened 2021-12-23
DDP-7335

aleksey-kuznetsov:
- Show modal upon clicking on "Sign Up" button
- Prevent users from navigating to `/join-us`

### DIFF
--- a/ddp-workspace/projects/ddp-circadia/src/app/components/header/header.component.ts
+++ b/ddp-workspace/projects/ddp-circadia/src/app/components/header/header.component.ts
@@ -6,6 +6,9 @@ import { CompositeDisposable, SessionMementoService } from 'ddp-sdk';
 
 import { Route } from '../../constants/route';
 import { AppAuth0AdapterService, SDKAuth0AdapterService } from '../../services/auth0Adapter.service';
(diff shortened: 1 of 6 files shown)


ddp-workspace/projects/ddp-circadia/src/app/components/header/header.component.ts
```diff
@@ -21,12 +24,14 @@ import { AppAuth0AdapterService, SDKAuth0AdapterService } from '../../services/a
 export class HeaderComponent implements OnInit, OnDestroy {
   Route = Route;
   isSticky = false;
+  isEnrollmentPaused = DDP_ENV.enrollmentPaused;
```
aglushko-broad, line 27: that's interesting. I thought we use ConfigurationService in order to read values from DDP_ENV. What do you think?

aglushko-broad (approved):

ddp-workspace/projects/ddp-circadia/src/app/components/header/header.component.ts
aleksey-kuznetsov: That's right but I don't think that we should introduce this new property in `ConfigurationService` since it's the first time we are pausing the whole study.

PR #1007, opened 2022-01-07
[DDP-SDK] Fix linter warnings for ddp-sdk

alextanlan:
Fixed new errors/warnings for ddp-sdk.
(previously some new rules were added/tweaked for eslint in [PR#1006](https://github.com/broadinstitute/ddp-angular/pull/1006)).

Also the linter warns about lots of deprecated 'toPromise'/'forkJoin'/'throwError' stuff.
Suggesting to fix them later in a separate PR (in order to decrease code-review effort).
(Temporarily[ switched off the deprecation rule](https://github.com/broadinstitute/ddp-angular/pull/1007/files#:~:text=%22deprecation/deprecation%22%3A%20%22off%22%2C%20//%20Temporarily%20changed%20from%20%27warn%27%20to%20%27off%27))


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-text-input/activityTextInput.component.ts
```diff
@@ -89,11 +89,11 @@ export class ActivityTextInput implements OnInit, OnChanges, OnDestroy {
         const adjustedOffset = tagsAddedLength + match.offset;
 
         taggedValue =
-          taggedValue.substr(0, adjustedOffset) +
+          taggedValue.substring(0, adjustedOffset) +
           openTag +
-          taggedValue.substr(adjustedOffset, match.length) +
+          taggedValue.substring(adjustedOffset, adjustedOffset + match.length) +
```
MocanaAtBroad, line 94: I think this might be wrong. Please check.

MocanaAtBroad (commented):
Please double check the conversions from `substr()` to `substring()`.
Also when running tests it looks like a bunch might be missing `logger` initialization.
Screen Shot 2022-01-17 at 3 31 03 PM

alextanlan:
> when running tests it looks like a bunch might be missing logger initialization.

Thanks for the catch. Fixed in 22a00706c4f04bb740df5a1a5f4d7b8f4c6b77a2

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-text-input/activityTextInput.component.ts
alextanlan: Alright. 
The `substr()` method begins at a specified position, and returns a specified number of characters, e.g. `substr(startIndex, length)`.
The `substring()` method extracts characters from start to end (endIndex is not included), e.g. `substring(startIndex, endIndex)`.
Both methods don't change the original string.

case 1. Cut from the string beginning:
```
'abcdefg'.substr(0, 3) -> 'abc'
'abcdefg'.substring(0, 0+3) -> 'abc'
```

case 2. Cut from the string middle:
```
'abcdefg'.substr(2, 3) -> 'cde'
'abcdefg'.substring(2, 2+3) -> 'cde'
```

case 3. Cut from the string middle to the string end (second argument is undefined):
```
'abcdefg'.substr(4) -> 'efg'
'abcdefg'.substring(4) -> 'efg'
```

case 4. Cut from the string middle to the string end (second argument is bigger then endIndex/string length):
```
'abcdefg'.substr(3, 58) -> defg'
'abcdefg'.substring(3, 3+58) -> 'defg'
```

So, we can replace `substr(startIndex, length)` with `substring(startIndex, startIndex + length).`

MocanaAtBroad: OK. Thanks for looking into this! Was doing the math in my head and got it wrong.

MocanaAtBroad (approved):

PR #1022, opened 2022-01-21
[DDP-SDK] Fix `ddp-sdk` build

alextanlan:
There are some issues during the `ddp-sdk` build.

Done in the PR: 
- fixed several circular dependencies in `sdk`
- added missed classes in `public-api.ts`
- refactored providing of `PicklistSortingPolicy` service

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/services/picklistSortingPolicy.service.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/services/picklistSortingPolicy.service.ts
@@ -1,12 +1,28 @@
-import { Injectable } from '@angular/core';
+import { Inject, Injectable, InjectionToken } from '@angular/core';
 import { ActivityPicklistNormalizedGroup } from '../models/activity/activityPicklistNormalizedGroup';
 import { SortOrder } from './sortOrder';
 import { ActivityPicklistOption } from '../models/activity/activityPicklistOption';
 import { FuncType } from '../models/funcType';
 
+export const PICKLIST_SORTING_POLICY_MAIN_SORT_ORDER = new InjectionToken<string>(
+    'mainSortOrder for PicklistSortingPolicy configuration', {
+        factory: () => SortOrder.NONE // default value for DI
+    }
+);
+export const PICKLIST_SORTING_POLICY_LAST_STABLE_ID = new InjectionToken<string>(
+    'lastStableId (optional) for PicklistSortingPolicy configuration', {
+        factory: () => undefined // default value for DI
+    }
+);
+
 @Injectable()
 export class PicklistSortingPolicy {
-    constructor(readonly mainSortOrder: SortOrder = SortOrder.NONE, readonly lastStableId?: string) {}
+    constructor(
+        @Inject(PICKLIST_SORTING_POLICY_MAIN_SORT_ORDER)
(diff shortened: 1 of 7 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/services/picklistSortingPolicy.service.ts
```diff
@@ -1,12 +1,28 @@
...
+);
+
 @Injectable()
 export class PicklistSortingPolicy {
-    constructor(readonly mainSortOrder: SortOrder = SortOrder.NONE, readonly lastStableId?: string) {}
+    constructor(
+        @Inject(PICKLIST_SORTING_POLICY_MAIN_SORT_ORDER)
+            readonly mainSortOrder: SortOrder = SortOrder.NONE, // default value for new PicklistSortingPolicy()
+        @Inject(PICKLIST_SORTING_POLICY_LAST_STABLE_ID) readonly lastStableId?: string
+    ) {
```
alextanlan, line 24: Fixed a [dependency injection (DI) error](https://angular.io/errors/NG2003).
 Angular DI Provider can't be used with primitive types (string in our case).
 Thus I used [InjectionToken](https://angular.io/api/core/InjectionToken) which helps us injecting a string.

For now:
1) If we create a new instance of **_PicklistSortingPolicy_** without any parameters (`new PicklistSortingPolicy()`), 
there is a default value for mainSortOrder (`mainSortOrder = SortOrder.NONE`) in the constructor.

2) When we use an instance of **_PicklistSortingPolicy_** from Angular DI Provider _**without parameters**_, 
like e.g. 
```
@NgModule({
...
  providers: [
    ...
    PicklistSortingPolicy 
  ]
})
```
   it is created with default values ([ defined in InjectionToken-s](https://github.com/broadinstitute/ddp-angular/pull/1022/files#:~:text=from%20%27../models/funcType%27%3B-,export%20const%20PICKLIST_SORTING_POLICY_MAIN_SORT_ORDER%20%3D%20new%20InjectionToken%3Cstring%3E(,)%3B,-%40Injectable())).

3) In case we need an instance of **_PicklistSortingPolicy_** from Angular DI Provider **_with parameters,_** we should specify their values, providing the mentioned injections tokens, like:
```
@NgModule({
...
  providers: [
    ...
    {
      provide: PICKLIST_SORTING_POLICY_MAIN_SORT_ORDER,
      useValue: SortOrder.ALPHABETICAL
    },
    {
      provide: PICKLIST_SORTING_POLICY_LAST_STABLE_ID,
      useValue: 'UNSURE'
    },
    PicklistSortingPolicy
  ]
})
```

MocanaAtBroad: This is better than what I had done originally!
Have we verified this is doing the right thing in, say, the Pancan cancer list in prequal activity?

MocanaAtBroad (commented):
Looks good, cleared the circular dependency errors, and liked what you did with the sorting policy. Requesting we do a quick sanity check with pancan if not done so already.

ddp-workspace/projects/ddp-sdk/src/lib/services/picklistSortingPolicy.service.ts
alextanlan: yes, I have been checking it by Pancan study (cancers autocomplete list).

MocanaAtBroad (approved):

PR #1034, opened 2022-02-01
fix: added content sanitization on initialization

grigol-brd:
https://broadinstitute.atlassian.net/browse/DDP-7470

Fix for failing test


nsakande (approved):
LGTM

PR #1036, opened 2022-02-02
[ATCP] DDP-7500: fix failed `atcp/atcp-auth` build

alextanlan:
1) **Initial issue** - 
ATCP study: login page fails.  `ddp-atcp-auth.umd.js` file can't be downloaded (404). 
The file is used in `ddp-study-server\study-builder\tenants\atcp\pages\login.html` as a JS script (<script>).

image

2) Seems like after Angular update to v.13 an Angular builder `ng-packagr` does not create a build/bundle in **umd** format by default at present, but only as ES modules. Moreover, I was not able to configure it (see [ng-packagr doc](https://github.com/ng-packagr/ng-packagr/blob/master/docs/DESIGN.md#:~:text=For%20example%2C%20the,JSON%20configuration%20property)).
Thus I decided to add[ an additional step to convert ](https://github.com/broadinstitute/ddp-angular/pull/1036/files#:~:text=npm%20run%20convert%2Datcp%2Dauth%2Dbuild%2Dto%2DUMD)the created (ES2015-module) build into umd one by using [Rollup library](https://www.npmjs.com/package/rollup) (the library is a peer dependency of ng-packagr).

3) Checked on dev env - it works properly.
atcp_after

4) May be it makes sense to refactor the `ddp-atcp-auth` library in ES-modules style in future.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -315,6 +315,7 @@ commands:
           command: |
             set -u
             ng build ddp-atcp-auth
+            npm run convert-atcp-auth-build-to-UMD
       - run:
           name: build atcp auth0 styles
           command: |
--- a/ddp-workspace/package.json
+++ b/ddp-workspace/package.json
@@ -7,7 +7,8 @@
     "build": "ng build",
     "test": "ng test",
     "lint": "ng lint",
-    "e2e": "ng e2e"
+    "e2e": "ng e2e",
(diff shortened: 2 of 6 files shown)


josemanuel167 (approved):

MocanaAtBroad (commented):
Looks good! Nice job figuring out a solution for this one!

PR #1043, opened 2022-02-10
[DSS] Fix failed tests 

alextanlan:
Try to fix (flacky) failed tests.
We notice them periodically on CircleCI pipeline.

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/submissionManager.service.spec.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/submissionManager.service.spec.ts
@@ -1,6 +1,6 @@
 import { ActivityServiceAgent } from './activityServiceAgent.service';
 import { Observable, of, from, interval, timer, throwError } from 'rxjs';
-import { delay, delayWhen, take, tap } from 'rxjs/operators';
+import { delayWhen, take, tap } from 'rxjs/operators';
 import { TestBed } from '@angular/core/testing';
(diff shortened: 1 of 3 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/submissionManager.service.spec.ts
```diff
@@ -90,11 +91,14 @@ describe('SubmissionManagerTest',() => {
...
         // Fake the call to service agent, including the response delay
         const httpCallDelay = 500;
+
         serviceAgent.saveAnswerSubmission.and.callFake((): Observable<PatchAnswerResponse> => {
             return of({
                 answers: [],
                 blockVisibility: []
-            }).pipe(delay(httpCallDelay));
+            }).pipe(
+                delayWhen(() => timer(httpCallDelay))
```
alextanlan, line 100: The only thing I changed in this file:
`delay(httpCallDelay)` => `delayWhen(() => timer(httpCallDelay)`.
Just a guess, because it is practically the same, 
but other tests in the file have `delayWhen` and don't fail.
[This particular test (line 145)](https://github.com/broadinstitute/ddp-angular/pull/1043/files#:~:text=expect(returnedInProgressValues%5B2%5D.delay).toBeGreaterThanOrEqual(httpCallDelay)%3B) fails from time to time on CircleCI (but never locally).

Let's try. It it does not help, may be it makes sense to avoid/comment/remove this test/line as we are going to have a required tests check passed in every PR in future.

josemanuel167 (approved):

PR #1045, opened 2022-02-10
DDP-7115: Add a new api call in order to run tests in CircleCI

alextanlan:
Modified a script that can be used by developers to trigger the execution of all Angular tests in their PR branch
(for the pointed project/study (**`study_key`**) and also **`ddp-sdk`** and **`toolkit`** as common libraries).

To run test use the following command (from `ddp-angular` folder): 

 ### ` ./build-utils/run_ci.sh run-tests [study_key] [branch_name] [target_env] `

  e.g.: 
### ` ./build-utils/run_ci.sh run-tests pancan develop dev `

---
**_Note:_**
If we are going to set "**_all tests have passed_**" as a required rule before a PR merge 
we should remove/comment out flacky tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -538,6 +538,13 @@ jobs:
       - deploy-stored-build:
           study_key: << parameters.study_key >>
 
+  app-run-tests-job:
+    working_directory: *ng_workspace_path
+    executor:
+      name: build-executor
+    steps:
+      - run-tests
+
 parameters:
   study_key:
     type: string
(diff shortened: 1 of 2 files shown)


josemanuel167 (approved):

MocanaAtBroad:
Looks good!
One comment: It is not necessary to specify the deployment enviroment to to run tests, so you can just type:
`./build-utils/run_ci.sh run-tests pancan develop`

PR #1058, opened 2022-02-15
DDP-7307: Implement decimal data type

alextanlan:
Implemented UI (Angular) part of decimal data type [(DDP-7184 epic)](https://broadinstitute.atlassian.net/browse/DDP-7184).

**_Notes:_** 
 - used separate data models (**NUMERIC** - ActivityNumericQuestionBlock/ **DECIMAL** - ActivityDecimalQuestionBlock) as to apply **"REQUIRED"** validator in a generic way (will be tweaked in DDP-7573)
 - used the same component ActivityNumericAnswer for both **NUMERIC** / **DECIMAL** answers
     in order to avoid lots of duplication (logic / styles)
- set `valueChangeStep` property calculated from a question scale. It is used as an input step (the input value can be changed by ArrowUp/ArrowDown)
- default value for a question scale ("_the number of digits allowed to the right of the decimal point_") is 0 (zero). In this case the input value does not have any decimal part (the number is integer). _**See the first screenshot below.**_

**_Screenshots:_**

1) Display a saved decimal answer (question scale is not defined in the server response)
q1

2) Patch a decimal answer (question scale = 2, cut all extra digits from the right, e.g. `3.141516` => `3.14` )
q2

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-answer/activityAnswer.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-answer/activityAnswer.component.ts
@@ -3,7 +3,6 @@ import { AbstractActivityQuestionBlock } from '../../../../models/activity/abstr
 import { AnswerValue } from '../../../../models/activity/answerValue';
 import { QuestionType } from '../../../../models/activity/questionType';
 import { BlockType } from '../../../../models/activity/blockType';
-import { NumericType } from '../../../../models/activity/numericType';
 import { ActivityNumericQuestionBlock } from '../../../../models/activity/activityNumericQuestionBlock';
 
 @Component({
@@ -21,7 +20,7 @@ import { ActivityNumericQuestionBlock } from '../../../../models/activity/activi
                                       [readonly]="readonly"
                                       (valueChanged)="onChange($event)">
             </ddp-activity-text-answer>
-            <ddp-activity-numeric-answer *ngIf="isNumericQuestion(block)"
+            <ddp-activity-numeric-answer *ngIf="isNumericQuestion(block) || isDecimalQuestion(block)"
                                          [class]="'numeric-answer-' + block.stableId"
                                          [block]="block"
                                          [valueChangeStep]="numericValueStep"
@@ -94,9 +93,8 @@ export class ActivityAnswerComponent {
     @Output() componentBusy = new EventEmitter<boolean>();
 
     public get numericValueStep(): number {
-        if (this.isNumericQuestion(this.block)) {
-            const block = this.block as ActivityNumericQuestionBlock;
-            return (block.numericType === NumericType.Integer) ? 1 : Math.pow(10, -(block.scale));
+        if (this.isDecimalQuestion(this.block)) {
+            return Math.pow(10, -( (this.block as ActivityNumericQuestionBlock).scale) );
         }
     }
 
@@ -114,6 +112,10 @@ export class ActivityAnswerComponent {
 
     public isNumericQuestion(block: AbstractActivityQuestionBlock): boolean {
         return this.isQuestion(block) && block.questionType === QuestionType.Numeric;
+    }
+
+    public isDecimalQuestion(block: AbstractActivityQuestionBlock): boolean {
+        return this.isQuestion(block) && block.questionType === QuestionType.Decimal;
     }
 
     public isPicklistQuestion(block: AbstractActivityQuestionBlock): boolean {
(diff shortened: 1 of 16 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activityNumericAnswer.component.ts
```diff
@@ -60,8 +72,65 @@ export class ActivityNumericAnswer implements OnInit, OnChanges, OnDestroy {
...
+            return null;
+        }
+
+        return this.isIntegerQuestion ?
+            parseInt(answerValue, 10) :
+            this.formatDecimalAnswer(answerValue);
+    }
+
+    private formatDecimalAnswer(answerValue: string): DecimalAnswer {
+        const [integerPart = '', decimalPart = ''] = answerValue.split('.');
```
josemanuel167, line 112: Not sure if new ESLINT rule but does it have to be initializaed like that?

josemanuel167 (dismissed):

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activityNumericAnswer.component.ts
alextanlan: It is not eslint, just a custom logic.
I used default values ('') for cases when `integerPart ` or `decimalPart ` is `undefined` (most likely `decimalPart `, but put two initializations just for readability/consistency).
Because in the next lines there is a transformation a string to a number:
`value: +integerPart.concat(decimalPart)`. Having  `undefined` would lead to an error:

e.g.
```
answerValue = '12345';
  => integerPart = '12345', decimalPart = undefined;
    => value = NaN
```

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-answer/activityAnswer.component.ts
```diff
@@ -90,6 +92,12 @@ export class ActivityAnswerComponent {
     @Output() valueChanged: EventEmitter<AnswerValue> = new EventEmitter();
     @Output() componentBusy = new EventEmitter<boolean>();
 
+    public get numericValueStep(): number {
+        if (this.isDecimalQuestion(this.block)) {
+            return Math.pow(10, -( (this.block as ActivityDecimalQuestionBlock).scale) );
```
MocanaAtBroad: would this calculation be better inside the ddp-activity-numeric-answer ?

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityDecimalQuestionBlock.ts
```diff
@@ -0,0 +1,27 @@
+import { ActivityQuestionBlock } from './activityQuestionBlock';
+import { QuestionType } from './questionType';
+
+export interface DecimalAnswer {
+    value: number;
+    scale: number;
+}
+export type NumericAnswerType = number | DecimalAnswer;
+
```
MocanaAtBroad, line 4: I think we should break these out into their own separate files; NumericAnswerType and DecimalAnswer get used in different in code outside of this file.
(At some point we should figure out how to reorganize all the code that is not services or components!)

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/activityValidatorBuilder.service.ts
```diff
@@ -28,6 +28,21 @@ import { ActivityPicklistQuestionBlock } from '../../models/activity/activityPic
...
+    NumOptionsSelected = 'NUM_OPTIONS_SELECTED',
+    YearRequired = 'YEAR_REQUIRED',
+    MonthRequired = 'MONTH_REQUIRED',
+    DayRequired = 'DAY_REQUIRED',
+    DateRange = 'DATE_RANGE',
+    AgeRange = 'AGE_RANGE',
+    IntRange = 'INT_RANGE',
+    Unique = 'UNIQUE'
+}
+
```
MocanaAtBroad, line 31: I think giving its own type is a good idea, but if we are going to use from multiple places we should give its own file.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-answer/activityAnswer.component.ts
alextanlan: Good point!
Fixed in 74260431b038b532f0e15b359baca147a9e925ba

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityDecimalQuestionBlock.ts
alextanlan: Fixed, 192bd238fb65f174a489e6d5a45bd51c0beda71b

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/activityValidatorBuilder.service.ts
alextanlan: Fixed in efd6c439616357c0a0c1aac65b1fe0e765a687c8

MocanaAtBroad (approved):

PR #1059, opened 2022-02-16
DDP-7573: Add decimal answer validators on FE (`REQUIRED` / `DECIMAL_RANGE`)

alextanlan:
Implemented `REQUIRED` and `DECIMAL_RANGE` (front-end side) validators for decimal answers.

- **REQUIRED**

s2

- **DECIMAL_RANGE**

s1

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityDecimalRangeValidationRule.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityDecimalRangeValidationRule.ts
@@ -1,13 +1,12 @@
 import { ActivityAbstractValidationRule } from './activityAbstractValidationRule';
 import { ActivityQuestionBlock } from '../../../models/activity/activityQuestionBlock';
-import { DecimalAnswer } from '../../../models/activity/activityDecimalQuestionBlock';
 import { DecimalHelper } from '../../../utility/decimalHelper';
 
 export class ActivityDecimalRangeValidationRule extends ActivityAbstractValidationRule {
     constructor(
         public question: ActivityQuestionBlock<any>,
-        public min: DecimalAnswer | null = null,
-        public max: DecimalAnswer | null = null) {
+        public min: number | null = null,
+        public max: number | null = null) {
         super(question);
     }
 
@@ -16,20 +15,16 @@ export class ActivityDecimalRangeValidationRule extends ActivityAbstractValidati
 
         if (DecimalHelper.isDecimalAnswerType(this.question.answer)) {
             const answer = DecimalHelper.mapDecimalAnswerToNumber(this.question.answer);
-            const min = this.min && DecimalHelper.mapDecimalAnswerToNumber(this.min);
-            const max = this.max && DecimalHelper.mapDecimalAnswerToNumber(this.max);
-
-            if (min && max === null) {
-                valid = answer >= min;
+            if (this.min && this.max === null) {
+                valid = answer >= this.min;
             }
-            if (max && min === null) {
-                valid = answer <= max;
+            if (this.max && this.min === null) {
+                valid = answer <= this.max;
             }
-            if (min && max) {
-                valid = answer >= min && answer <= max;
+            if (this.min && this.max) {
+                valid = answer >= this.min && answer <= this.max;
             }
         }
-
         this.result = valid ? null : this.message;
         return valid;
     }
--- a/ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityDecimalRangeValidationRule.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityDecimalRangeValidationRule.ts
@@ -1,6 +1,7 @@
 import { ActivityAbstractValidationRule } from './activityAbstractValidationRule';
 import { ActivityQuestionBlock } from '../../../models/activity/activityQuestionBlock';
 import { DecimalHelper } from '../../../utility/decimalHelper';
+import * as _ from 'underscore';
 
 export class ActivityDecimalRangeValidationRule extends ActivityAbstractValidationRule {
     constructor(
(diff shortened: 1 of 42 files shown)


josemanuel167 (dismissed):

ddp-workspace/projects/ddp-sdk/src/lib/utility/decimalHelper.ts
```diff
@@ -0,0 +1,12 @@
+import { DecimalAnswer } from '../models/activity/decimalAnswer';
+import * as _ from 'underscore';
+
+export  class DecimalHelper {
+    static mapDecimalAnswerToNumber(answer: DecimalAnswer ): number {
+        return answer.value * Math.pow(10, -(answer.scale));
+    }
+
+    static isDecimalAnswerType(obj: any): boolean {
+        return obj && _.isNumber(obj.value) && _.isNumber(obj.scale);
```
MocanaAtBroad, line 10: Will this work correctly if the answer is 0 ?

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityDecimalRangeValidationRule.ts
```diff
@@ -0,0 +1,31 @@
...
+        public max: number | null = null) {
+        super(question);
+    }
+
+    public recalculate(): boolean {
+        let valid = true;
+
+        if (DecimalHelper.isDecimalAnswerType(this.question.answer)) {
+            const answer = DecimalHelper.mapDecimalAnswerToNumber(this.question.answer);
+            if (this.min && this.max === null) {
```
MocanaAtBroad: Will this work if `min` is 0?

MocanaAtBroad: Same question for `max`

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityDecimalRangeValidationRule.ts
```diff
@@ -0,0 +1,31 @@
...
+                valid = answer >= this.min;
+            }
+            if (this.max && this.min === null) {
+                valid = answer <= this.max;
+            }
+            if (this.min && this.max) {
+                valid = answer >= this.min && answer <= this.max;
+            }
+        }
```
MocanaAtBroad, line 25: Consider refactoring so you don't have to do the min max comparison more than once, e.g.:
let valid = true;
if (min value is set)
   valid = valid && (answer >= this.min)
if (max value is set)
   valid = valid && (answer <= this.max)

You can then skip the case where both are set in line 24.

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityNumericRangeValidationRule.ts
```diff
@@ -15,10 +15,10 @@ export class ActivityNumericRangeValidationRule extends ActivityAbstractValidati
         const answer = this.question.answer;
```
MocanaAtBroad, line 15: for the question type, couldn't we use `ActivityQuestionBlock<number>` instead of any?

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityNumericRangeValidationRule.ts
```diff
@@ -15,10 +15,10 @@ export class ActivityNumericRangeValidationRule extends ActivityAbstractValidati
...
             if (_.isNumber(this.min) && this.max === null) {
-                valid = answer < this.min ? false : true;
+                valid = answer >= this.min;
             }
             if (_.isNumber(this.max) && this.min === null) {
-                valid = answer > this.max ? false : true;
+                valid = answer <= this.max;
             }
             if (_.isNumber(this.max) && _.isNumber(this.min)) {
                 valid = answer >= this.min && answer <= this.max ? true : false;
```
MocanaAtBroad: Multiple comparisons refactor mentioned above applies here too.

ddp-workspace/projects/ddp-sdk/src/lib/utility/decimalHelper.ts
alextanlan: This method checks whether an input object is decimal answer ("duck typing").
According to the [API](https://pepper-dev.datadonationplatform.org/docs#operation/getActivity):

ans1

a decimal answer should be an object with required `value` and `scale` fields that are numbers.
Thus
>  if the answer is 0 ?

in this case it is just a not decimal answer.

alextanlan: If you meant an answer like `{value: 0, scale: 0}`  -
  there is a test for such a case 
` expect(DecimalHelper.isDecimalAnswerType({value: 0, scale: 0})).toBeTrue();`
(decimalHelper.spec.ts, line 23)

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityDecimalRangeValidationRule.ts
alextanlan: Good catch, thanks!
Missed that corner case during some refactoring.
Fixed in f0376703b3a1f27b4b84377329da579612051ea8

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityDecimalRangeValidationRule.ts
alextanlan: Cool! 
I'd even replace 
`if (min value is set)
valid = valid && (answer >= this.min)
`
with
`if (min value is set)
valid = answer >= this.min
`
 because **valid** is always **true** there. 

 Fixed in f0376703b3a1f27b4b84377329da579612051ea8

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityNumericRangeValidationRule.ts
alextanlan: Fixed in 679a4f0def91d8a3207d8232645e926a299f3c1e

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/activityNumericRangeValidationRule.ts
alextanlan: Fixed in 679a4f0def91d8a3207d8232645e926a299f3c1e

MocanaAtBroad (approved):

PR #1077, opened 2022-02-24
DDP-7424: removing display name

SimoneMaiwald:
study staff can not give display name anymore
moved fieldSettings model into sub folder


ddp-workspace/projects/ddp-dsm-ui/src/app/field-settings/field-settings.component.ts
```diff
@@ -166,53 +166,41 @@ export class FieldSettingsComponent implements OnInit {
...
     if (this.realm != null) {
       this.saving = true;
-      let foundError = false;
-      let cleanedFieldSettings: Array<FieldSettings>;
-      if (this.settingsOfSelectedType != null && this.settingsOfSelectedType.length > 0) {
-        cleanedFieldSettings = FieldSettings.removeUnchangedFieldSettings(this.settingsOfSelectedType, this.selectedType);
-        for (const setting of cleanedFieldSettings) {
-          if (setting.notUniqueError || setting.spaceError) {
-            foundError = true;
+      // let cleanedFieldSettings: Array<FieldSettings>;
```
alextanlan, line 169: Do you need that commented code ?

alextanlan (approved):

ddp-workspace/projects/ddp-dsm-ui/src/app/field-settings/field-settings.component.html
```diff
@@ -43,9 +43,6 @@ <h4>{{selectedType.name}} Settings</h4>
           </th>
         </tr>
         <tr>
-          <th>
```
gmakharat, line 46: @SimoneMaiwald is this PR related to avoid inconvenient `column_name` in `field_settings` from study stuff?

SimoneMaiwald: yes. that was the reason for making that change

gmakharat: Since we need `variable_name` in any case, I think it is better not to remove it from `frontend`, instead we may keep it but add check on the input that they could only enter values which follow `camelCase` or `camel_case` pattern

PR #1084, opened 2022-02-28
DDP-7525: Validation for Unique Values

alextanlan:
Implemented front-end (Angular) handler for a new (back-end) validation type **`UNIQUE_VALUE`**  [(DDP-7182)](https://broadinstitute.atlassian.net/browse/DDP-7182).
It works generally for all failed answer **PATCH** requests with **422** error:

scrn2

scrn1


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-text-input/activityTextInput.component.ts
```diff
@@ -104,12 +104,17 @@ export class ActivityTextInput implements OnInit, OnChanges, OnDestroy {
...
-    });
+    const formGroup = new FormGroup(
+      {
+        [this.controlName]: new FormControl({
+          value: this.block.answer,
+          disabled: this.readonly,
+        }),
+      },
+      {
+        updateOn: this.block.hasUniqueValueValidator ? 'blur' : 'change'
```
alextanlan, line 115: Invoke the input validation on blur for questions with **UNIQUE_VALUE** validation (for optimal performance)

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/activityValidatorBuilder.service.ts
```diff
@@ -52,6 +52,7 @@ export class ActivityValidatorBuilder {
             { type: ValidationRuleType.AgeRange, factory: (x, y) => this.buildAgeRangeValidator(x, y) },
             { type: ValidationRuleType.IntRange, factory: (x, y) => this.buildNumericRangeValidator(x, y) },
             { type: ValidationRuleType.Unique, factory: (x, y) => new ActivityUniqueValidationRule(y) },
+            { type: ValidationRuleType.UniqueValue },
```
alextanlan, line 58: In order to prevent a logger error "**_Received unknown type of validation rule.._**" (see [lines 73-74](https://github.com/broadinstitute/ddp-angular/pull/1084/files#:~:text=this.logger.logError(this,rule%20named%3A%20%24%7BvalidationJson.rule%7D%60)%3B))

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -400,4 +405,30 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
...
+            const violatedQuestions = questionBlocks.filter(qBlock => qBlock.stableId === violation.stableId);
+            this.handleViolation(violatedQuestions, violation.rules);
+        }
+    }
+
+    private handleViolation(violatedQuestions: AbstractActivityQuestionBlock[], validationRules = []): void {
+        for (const violatedQuestion of violatedQuestions) {
+            violatedQuestion.serverValidationMessages = [];
+            validationRules.map(rule => rule.message)
+                .filter(Boolean)
```
josemanuel167, line 423: Not sure what is going with this filter, could we make it more readable?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-text-input/activityTextInput.component.ts
josemanuel167: Seems good, what is this used for? Does it blur the question?

josemanuel167 (dismissed):

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
alextanlan: It is kind of an alias for ` filter(x => !!x) `.

The **Boolean** is a function, actually a global reference points to the constructor function (`Boolean.prototype.constructor`) that returns a boolean value (true / false) from the first argument. 

E.g. `Boolean('some text' | 123 | [] | {}) => true ; Boolean(null | undefined | 0 | '') => false`

Thus:
`.filter(Boolean)` === `.filter(x => Boolean(x))` === `.filter(x => !!x)` 


[docs link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean#:~:text=to%20convert%20a%20non%2Dboolean%20value%20to%20a%20boolean%20value.%20To%20perform%20this%20task%2C%20instead%2C%20use%20Boolean%20as%20a%20function%2C%20or%20a%20double%20NOT%20operator%3A)

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-text-input/activityTextInput.component.ts
alextanlan: > Does it blur the question?

No. It means that the `FormControl` [is updated on the input **blur** event](https://angular.io/api/forms/FormControl#configure-the-control-to-update-on-a-blur-event) only 
(not for every symbol typing, but only when the input field loses focus).

 We should apply the `onBlur` update for some inputs that have **UNIQUE_VALUE** validation rule 
(in order to decrease an amount of requests to the server side - because this particular validation is quite a costly operation on BE side)

```mermaid
flowchart LR
B(ActivityTextAnswer \n component)
B --> C{has \n *UNIQUE_VALUE* \n validation rule}
C -->|Yes| D[Send requests on *blur* event \n   = when the text input loses a focus ]
C -->|No| E[Send requests by every change \n of the input value \n using a debounceTime ]
```

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/activityValidatorBuilder.service.ts
alextanlan: Since we have more server-side validations defined in a question definition (e.g. **unique_value** /  **cross-form** validations)
there is a question: 
  Should we use such approach for those validations 
    or  
  we can just write the error (mentioned above) in console (but in Stackdriver errors as well).

What do you think, @MocanaAtBroad ?

MocanaAtBroad: Writing an error to stackdriver is appropiate.
Even if all the client can do is to show the error returned by the server, I think that it would be a system error condition if Angular code encounters a validation type in the activity instance JSON that it does not recognize.

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/submissionManager.service.ts
```diff
@@ -194,11 +211,21 @@ export class SubmissionManager implements OnDestroy {
                 });
             }));
 
+        // remove submissions that include failed answers PATCH with 422 error
+        const submissionRemovalOnPatch422Error$: QueueOpObservable = this.answerSubmissionPatch422ErrorSubject.pipe(
+            map((stableIds: string[]) => (queue: ActivityInstanceAnswerSubmission[]) => {
+                return queue.filter(item => {
+                    return !stableIds.includes(item.stableId);
```
MocanaAtBroad: I don't see why this is necessary. 
Say if question with stableId QUESTION_1 has answer e.g. "REPEAT" that violates the unique rule on the server
then I expect the server would return a 422 with the error violation.
But it could be that in the queue, the user has entered another e.g. "GOOD VALUE" that does not violate the rule.
I think in that case, we should just proceed to clear the error, submit the new answer and look again to see if there is an error or not in the response.
In the code above, if I am reading it correctly, "GOOD VALUE" would be removed from the queue and would not be sent to server and I think this would not be right.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -114,9 +116,13 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
         });
 
         // Get notified of failure patching. Submission Manager has given up.
-        const subErrSub = this.submissionManager.answerSubmissionFailure$.subscribe((error) => {
-            this.logger.logError('ActivityComponent', 'There was an error during submission:', error);
-            return this.navigateToErrorPage();
+        const subErrSub = this.submissionManager.answerSubmissionFailure$.subscribe((error: Error | HttpErrorResponse) => {
```
MocanaAtBroad: Better if we don't have references to lower level objects like HttpErrorResponse in components. At this level we should prefer to deal with domain model objects (e.g. "question" or just a string or something of the sort).
Maybe we need to create a ValidationError or something like that? GUI components should not be aware of http codes.

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/submissionManager.service.ts
```diff
@@ -122,13 +123,23 @@ export class SubmissionManager implements OnDestroy {
         const submitAnswerWithRetry: (submission: ActivityInstanceAnswerSubmission) => Observable<PatchAnswerResponse> =
             (submission) => {
                 return executeAnswerSubmission(submission).pipe(
+                    catchError(err => {
+                        if (err?.status === 422) {
+                            this.answerSubmissionFailureSubject.next(err);
+                            this.answerSubmissionPatch422ErrorSubject.next(this.getViolatedStableIds(err?.error?.violations));
+                            return EMPTY;
+                        } else {
```
MocanaAtBroad, line 132: Consider creating a public observable, similar to answerSubmissionResponse$ that provides a stream of objects that represent the 422 body responses ($answerDataErrors: Observable<PatchDataError>  ?). Maybe you can come up with something better.
I think these sort of errors are different than the ones that we record in answerSubmissionFailureSubject. (Check,but probably OK that the 422 errors are not fed to answerSubmissionFailureSubject. Was looking and appears that all we ever do answerSubmissionFailureSubject emits is to direct users to error page, like an unrecoverable error.)

Then , in the activity, you can just subscribe to $answerDataErrors and do whatever you got to do with validation violations/don't have to worry about sorting between communication errors and validation errors.

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/submissionManager.service.ts
alextanlan: > "GOOD VALUE" would be removed from the queue and would not be sent to server and I think this would not be right.

It works as expected for now.

https://user-images.githubusercontent.com/7396837/157223554-60cedaa9-d8ef-471f-b32e-a7ec42213127.mov

alextanlan: The thing is that in the current implementation we just push every submission in the queue, not checking if there is already the same submission with another (valid / invalid) value (submissionManager.service.ts, line 191). 

However, even if we check and push only a new (non-existing) submission in the queue - we will have at least one submission (with that invalid value) in the queue until we change the value. 
        
  In this case we have a freezing activity (`isPageBusy = true`) till the answer is valid. Because we rely to the condition
 ` this.isAnswerSubmissionInProgress$ = this.pendingAnswerSubmissionQueue$.pipe(map(queue => queue.length > 0), share());`  (submissionManager.service.ts, line 243)

See screenshot:
scr11

MocanaAtBroad: I don't think I am following your explanation here. Maybe we can talk about it.
The thing that still does not look right to me is that the QueueOp you implemented can potentially remove more than one item from the queue. I think it is correct that the queue gets updated, but my expectation is that it would be just one the one item that caused the error in the first place. And I think it would be the the last item in the queue. Or even more specific, the last item in the queue that has the corresponding question stable id.
Current use of `filter` can potentially remove 2 or more items and I don't think that is what we want. I think we just want to remove one. The last one. But if you think I am misundertanding things, just reach out.

alextanlan: Fixed as discussed in c9623f9f44b3a0765edefdae5ee74da9c6c6f706

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
alextanlan: Fixed

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/submissionManager.service.ts
alextanlan: Fixed.

MocanaAtBroad (approved):

PR #1087, opened 2022-03-01
changing service

SimoneMaiwald:
Context
DSM from the "old" repo can be stopped and DSM from the "new" repo should get that service name now


SimoneMaiwald:
@MocanaAtBroad how do I change the URL of the new DSM back to the old/normal one?

MocanaAtBroad (approved):

MocanaAtBroad:
> @MocanaAtBroad how do I change the URL of the new DSM back to the old/normal one?

I THINK that if we deploy the app to the old service name, the old URL should just work.
But we shall find out!

PR #1092, opened 2022-03-03
Ddp 7548/parental medical release form fixes

GiCharkviani:
https://broadinstitute.atlassian.net/jira/software/c/projects/DDP/boards/815?modal=detail&selectedIssue=DDP-7548

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
@@ -20,7 +20,7 @@ import { AnalyticsEventsService } from '../../services/analyticsEvents.service';
 import { SubmissionManager } from '../../services/serviceAgents/submissionManager.service';
 import { PatchAnswerResponse } from '../../models/activity/patchAnswerResponse';
 import { ActivitySection } from '../../models/activity/activitySection';
-import { NavigateFields } from '../../models/activity/navigateFields';
+import { NavigateFieldsOptionsModel} from '../../models/activity/navigateFields';
 import { AnalyticsEventCategories } from '../../models/analyticsEventCategories';
 import { CompositeDisposable } from '../../compositeDisposable';
 import { BehaviorSubject, combineLatest, Observable, Subscription, timer } from 'rxjs';
@@ -62,7 +62,7 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit, 
     private readonly LOG_SOURCE = 'ActivityComponent';
     private shouldSaveLastStep = false;
     private isAdminEditing = false;
-    private carouselRevert:boolean = false;
+    private carouselRevert: boolean;
 
     constructor(
         private logger: LoggingService,
@@ -88,38 +88,6 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit, 
         } else if (scrolledPixels < this.HEADER_HEIGHT) {
             this.isScrolled = false;
         }
-    }
-    // Navigate through fields
-    protected nextField(event: Event, {elements, preventDefault = false, carousel = false}: NavigateFields) {
-        preventDefault && event.preventDefault();
-        if(elements && elements.length > 0) {
-            // All provided elements array, excluding those, which doesn't have id
-            const elementsArray: Element[] = Array.from(document.querySelectorAll(elements.join(','))).filter(input => input.id && (input as HTMLElement)?.style.display !== 'none');
-            // Map data structure for elements in order to easily find them
-            const elementsObjects: Map<Element, number> = new Map(elementsArray.map((input, i) => [input, i]));
-            // Pressed element in a DOM
-            const pressedElement = event.target as Element;
-            // One-sided or Carousel behaviour while moving through fields
-            this.switchBehaviour(elementsArray, elementsObjects.get(pressedElement), carousel)
-        }
-    }
-    // Decide behaviour while navigation
-    private switchBehaviour(elements: Element[] | HTMLElement, elementIndex: number | undefined, carousel:boolean) {
-        if(carousel){
-            if(((elements as Element[]).length - 1 === elementIndex)) this.carouselRevert = true;
-            if(elementIndex === 0) this.carouselRevert = false;
-
-            this.carouselRevert ? this.moveBack(elements, elementIndex) : this.moveToNextField(elements, elementIndex);
-        } else {
-            this.moveToNextField(elements, elementIndex)
-        }
-    }
-
-    private moveToNextField(elements: Element[] | HTMLElement, elementIndex: number | undefined) {
-       elementIndex > -1 && elementIndex !== undefined ? elements[elementIndex + 1]?.focus() : elements[0]?.focus();
-    }
-    private moveBack(elements: Element[] | HTMLElement, elementIndex: number | undefined) {
-       elementIndex > -1 && elementIndex !== undefined ? elements[elementIndex - 1]?.focus() : elements[0]?.focus();
     }
 
     public ngOnInit(): void {
@@ -352,6 +320,47 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit, 
             }
         }
         return -1;
+    }
+
+    // Navigate through fields
+    protected nextField(event: Event, {elements, preventDefault = false, carousel = false}: NavigateFieldsOptionsModel): void {
+        preventDefault && event.preventDefault();
+        if(elements && elements.length > 0) {
+            // All provided elements array, excluding those, which doesn't have id or style.display is 'none'
+            const elementsArray: Element[] = Array.from(document.querySelectorAll(elements.join(',')))
+                .filter(input => input.id && (input as HTMLElement)?.style.display !== 'none');
+            // Map data structure for elements in order to easily find them
+            const elementsObjects: Map<Element, number> = new Map(elementsArray.map((input, i) => [input, i]));
+            // Pressed element in a DOM
+            const pressedElement = event.target as Element;
+            // One-sided or Carousel behaviour while moving through fields
+            this.switchBehaviour(elementsArray, elementsObjects.get(pressedElement), carousel);
+        }
+    }
+    // Decide behaviour while navigation
+    private switchBehaviour(elements: Element[], elementIndex: number | undefined, carousel: boolean): void {
+        if(carousel){
+            if(elements.length - 1 === elementIndex) {
+                this.carouselRevert = true;
+            }
+            if(elementIndex === 0) {
+                this.carouselRevert = false;
+            }
+
+            this.carouselRevert ? this.moveBack(elements, elementIndex) : this.moveToNextField(elements, elementIndex);
+        } else {
+            this.moveToNextField(elements, elementIndex);
+        }
+    }
+
+    private moveToNextField(elements: Element[], elementIndex: number | undefined): void {
+        const nextElement = elementIndex >= 0 ? elements[elementIndex + 1] : elements[0];
+        nextElement && (nextElement as HTMLElement).focus();
+    }
+
+    private moveBack(elements: Element[], elementIndex: number | undefined): void {
+        const prevElement = elementIndex >= 0 ? elements[elementIndex - 1] : elements[0];
+        prevElement && (prevElement as HTMLElement).focus();
     }
 
     private smoothScrollToTop(): void {
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
@@ -20,6 +20,7 @@ import { AnalyticsEventsService } from '../../services/analyticsEvents.service';
 import { SubmissionManager } from '../../services/serviceAgents/submissionManager.service';
 import { PatchAnswerResponse } from '../../models/activity/patchAnswerResponse';
 import { ActivitySection } from '../../models/activity/activitySection';
+import { NavigateFields } from '../../models/activity/navigateFields';
 import { AnalyticsEventCategories } from '../../models/analyticsEventCategories';
 import { CompositeDisposable } from '../../compositeDisposable';
 import { BehaviorSubject, combineLatest, Observable, Subscription, timer } from 'rxjs';
@@ -61,6 +62,7 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit, 
     private readonly LOG_SOURCE = 'ActivityComponent';
     private shouldSaveLastStep = false;
     private isAdminEditing = false;
+    private carouselRevert:boolean = false;
 
     constructor(
         private logger: LoggingService,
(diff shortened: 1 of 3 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -61,6 +62,7 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
     private readonly LOG_SOURCE = 'ActivityComponent';
     private shouldSaveLastStep = false;
     private isAdminEditing = false;
+    private carouselRevert:boolean = false;
```
alextanlan: Just wondering what is `carousel` that should be reverted in some cases ?

alextanlan, line 65: Just wondering what is `carousel` that should be reverted in some cases ?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -87,6 +89,38 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
             this.isScrolled = false;
         }
     }
+    // Navigate through fields
+    protected nextField(event: Event, {elements, preventDefault = false, carousel = false}: NavigateFields) {
```
alextanlan: **_#minor_**

It would be nice to move **_protected_** and **_private_** methods to the end of component as to have an order: 
public / protected / private

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -87,6 +89,38 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
...
+            // Pressed element in a DOM
+            const pressedElement = event.target as Element;
+            // One-sided or Carousel behaviour while moving through fields
+            this.switchBehaviour(elementsArray, elementsObjects.get(pressedElement), carousel)
+        }
+    }
+    // Decide behaviour while navigation
+    private switchBehaviour(elements: Element[] | HTMLElement, elementIndex: number | undefined, carousel:boolean) {
+        if(carousel){
+            if(((elements as Element[]).length - 1 === elementIndex)) this.carouselRevert = true;
```
alextanlan: Why it is typed like `elements: Element[] | HTMLElement` ? 
Seems like `elements` is always an array, isn't it ? 
if yes then you can avoid many parentheses there.

alextanlan, line 109: Why it is typed like `elements: Element[] | HTMLElement` ? 
Seems like `elements` is always an array, isn't it ? 
if yes then you can avoid many parentheses there.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -87,6 +89,38 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
...
+            if(elementIndex === 0) this.carouselRevert = false;
+
+            this.carouselRevert ? this.moveBack(elements, elementIndex) : this.moveToNextField(elements, elementIndex);
+        } else {
+            this.moveToNextField(elements, elementIndex)
+        }
+    }
+
+    private moveToNextField(elements: Element[] | HTMLElement, elementIndex: number | undefined) {
+       elementIndex > -1 && elementIndex !== undefined ? elements[elementIndex + 1]?.focus() : elements[0]?.focus();
```
alextanlan: you can skip `&& elementIndex !== undefined` condition because `undefined > -1 === false` in JS. 

The right part of the `elementIndex > -1 && elementIndex !== undefined` condition will never be checked.

alextanlan, line 119: 1. you can skip `&& elementIndex !== undefined` condition because `undefined > -1 === false` in JS. 

The right part of the `elementIndex > -1 && elementIndex !== undefined` condition will never be checked.

alextanlan: **_#optional_**

suggesting some refactoring like:

```
const nextElement = elementIndex >= 0 ? elements[elementIndex + 1] : elements[0];

if (nextElement) {
  nextElement.focus();
}
```
or something like this

alextanlan: **_#optional_**

suggesting some refactoring like:

```
const nextElement = elementIndex >= 0 ? elements[elementIndex + 1] : elements[0];

if (nextElement) {
  nextElement.focus();
}
```

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
GiCharkviani: Yes, when the last form field is reached, then it starts going back

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/navigateFields.ts
```diff
@@ -0,0 +1,5 @@
+export interface NavigateFields {
+    elements: string[] | [] | undefined,
```
alextanlan: Would not be enough just to have ` elements: string[];` ?

ddp-workspace/projects/ddp-singular/src/app/components/activity/activity.component.ts
```diff
@@ -15,6 +15,14 @@ export class ActivityComponent extends ActivityRedesignedComponent {
   Route = Route;
   isCaptchaResolved = false;
 
+  @HostListener('document:keydown.enter', ['$event']) onEnter(event:KeyboardEvent) {
+      super.nextField(event, {
```
alextanlan, line 19: **_#optional_** 
Should not the event fire on input fields only?

E.g. a corner case (rare and maybe weird, but anyway): 
 - there 10 input fields in a form
 - a user types something in 5th input field
 - the user moves focus out of the edited field (click on the document outside, press Tab and focus on a next button, 
    etc.)
- the user presses Enter by accident
- what field has focused ?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
GiCharkviani: Thank you for your feedback, I will do.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
GiCharkviani: Yes, thank you for your feedback, I will do.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
GiCharkviani: Thank you for your feedback, I will do.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
alextanlan: A corner case:
- there 10 input fields in a form
- a user types something in the last input field and presses Enter
- then the first input has focused
- the user moves focus to the next field (click on any input in the middle, press Tab and focus on a next input)
- the user presses Enter 
- what field has focused ?

ddp-workspace/projects/ddp-singular/src/app/components/activity/activity.component.ts
```diff
@@ -15,6 +15,14 @@ export class ActivityComponent extends ActivityRedesignedComponent {
...
+          elements: ['input', 'mat-select', 'select', 'textarea'],
+          preventDefault: true,
+          carousel: true
+      })
+  }
+
   get isPrequal(): boolean {
     return this.model && this.model.activityCode === ActivityCode.Prequal;
```
alextanlan: **_#minor_**

can be shortened like 
`return this.model?.activityCode === ActivityCode.Prequal;`

alextanlan:
Can you please, check and fix linter errors/warnings?

ddp-workspace/projects/ddp-singular/src/app/components/activity/activity.component.ts
GiCharkviani: Not only, but there might also be another type of Elements, like 'textarea' or 'select'.
Each time the user moves out focus, after pressing enter - the first field will be focused.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
GiCharkviani: Each time the user presses the enter, the focus will be moved to the next field, as it takes element, where enter was pressed and finds its following (the next) element and moves focus on to that element.
But we could change behavior if needed. It is just my implementation.

ddp-workspace/projects/ddp-singular/src/app/components/activity/activity.component.ts
alextanlan: Btw, what about a default behavior by pressing Enter: 
- **textarea** - go to the next line
- **select**  open a dropdown options list (even though it can be done by Space or mouse click) ?

It will be unavailable, won't it ?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -354,6 +322,47 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
...
+    }
+    // Decide behaviour while navigation
+    private switchBehaviour(elements: Element[], elementIndex: number | undefined, carousel: boolean): void {
+        if(carousel){
+            if(elements.length - 1 === elementIndex) {
+                this.carouselRevert = true;
+            }
+            if(elementIndex === 0) {
+                this.carouselRevert = false;
+            }
```
alextanlan, line 348: Just wondering whether it is enough to check for the last index only? 
Like: 
```
 if(elements.length - 1 === elementIndex) {
                this.carouselRevert = true;
            }
            if(elementIndex === 0) {
                this.carouselRevert = false;
            }
```
=> `this.carouselRevert =  elementIndex === elements.length - 1;`

ddp-workspace/projects/ddp-singular/src/app/components/activity/activity.component.ts
GiCharkviani: Yes, right, but as they asked, I had to change the default behavior of pressing Enter. I think that I probably could find some workaround, though. What do you say?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
alextanlan: I suppose in the described case above the focus will be moved **back** finally (instead of **forward**). 
Because there is a check for either the **first** or the **last** index only. 
Is it ok?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
GiCharkviani: Do you have in mind any case, where it could fail?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
GiCharkviani: Yes, only in this scenario, where the pressed element is not in the elements array (like "body"). Also, I have discussed this particular case with QA, and he says it's fine. But there is always a different way. What would you suggest?

ddp-workspace/projects/ddp-singular/src/app/components/activity/activity.component.ts
alextanlan: There are 2 arguable things:
- the implementation changes the default html elements behavior
- you placed a custom feature into the common `activity.component.ts` that is used in dozens of other studies 

May be it would be just sufficient to use a default possibility:
press **Tab** to move to the next focusable element,
press **Shift + Tab** to move to the previous one ?
[an example](https://www.a11y-collective.com/glossary/tab-order/)

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
alextanlan: https://github.com/broadinstitute/ddp-angular/pull/1092#discussion_r822547606
Can you please check the case ?

ddp-workspace/projects/ddp-singular/src/app/components/activity/activity.component.ts
GiCharkviani: I do agree that it changes the default HTML elements behavior, but is it not what was asked in this following task: 
- If you hit "return" on your keyboard, it takes you back to the dashboard and you have to start over. Hitting return should take you to the next line of the form

Yes, I did so only because I expected the same functionality would be required by other studies as well, and they could just easily call the same function. I also thought about implementing this functionality differently, though.

alextanlan: > default HTML elements behavior

I mean e.g. go to the next line in a textarea. 

>  hit "return" on your keyboard, it takes you back to the dashboard and you have to start over

Most likely it is an unknown (?) bug. At least a behavior that nobody wanted to fix before.
I don't think it is a special feature.

alextanlan: @MocanaAtBroad, what are your thoughts about it ?

GiCharkviani: > I mean e.g. go to the next line in a textarea.

Yes, I got this and I thought the same about changing default behavior. I have even a special option:
```
export interface NavigateFieldsOptionsModel {
    elements: string[];
    preventDefault?: boolean; <---- this
    carousel?: boolean;
}
```

> Most likely it is an unknown (?) bug. At least a behavior that nobody wanted to fix before. I don't think it is a special feature.

What do you think about only fixing this bug and keeping the default behavior of the enter key?

alextanlan: I guess it should be discussed with Marco, Kiara (QA) and Annie (the ticket creator)

GiCharkviani: Okay. Btw, We could also just avoid preventing default behaviour only with Textarea and Select HTML elements. I think input fields are fine. But as you said, it must be discussed.

PR #1100, opened 2022-03-10
DDP-7406: fix:  hide autocomplete options on scroll

grigol-brd:
It was requested that auto-complete dropdown options would disappear when a user scrolls up or down.

https://broadinstitute.atlassian.net/browse/DDP-7406

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
@@ -3,6 +3,7 @@ import { FormControl } from '@angular/forms';
 import { Subject } from 'rxjs';
 import { debounceTime, distinctUntilChanged, map, startWith, takeUntil } from 'rxjs/operators';
 import { MatAutocomplete, MatAutocompleteTrigger } from '@angular/material/autocomplete';
+import { boundMethod } from 'autobind-decorator';
 
 import { BaseActivityPicklistQuestion } from './baseActivityPicklistQuestion.component';
 import { ActivityPicklistNormalizedGroup } from '../../../models/activity/activityPicklistNormalizedGroup';
@@ -20,7 +21,8 @@ import { StringsHelper } from '../../../utility/stringsHelper';
                #autocompleteInput
                [formControl]="inputFormControl"
                [placeholder]="block.picklistLabel"
-               [matAutocomplete]="autoCompleteFromSource" />
+               [matAutocomplete]="autoCompleteFromSource"
+               (click)="openAutocompleteOptions()" />
 
         <mat-autocomplete #autoCompleteFromSource="matAutocomplete" class="autoCompletePanel" [displayWith]="displayAutoComplete">
             <mat-optgroup *ngFor="let group of filteredGroups">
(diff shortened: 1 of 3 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
```diff
@@ -112,8 +117,17 @@ export class AutocompleteActivityPicklistQuestion extends BaseActivityPicklistQu
         super.ngOnDestroy();
         this.ngUnsubscribe.next();
         this.ngUnsubscribe.complete();
+        window.removeEventListener('scroll', this.onScroll, true);
     }
 
+    onScroll = (event): void => {
```
MocanaAtBroad: This arrow function definition is technically correct, but looks out of place between instance methods.
Consider rewriting as `onScroll(event: WhateverTheEventType): void {`

MocanaAtBroad:
I did not see anything in the ticket about scrolling. Is this the right ticket?

MocanaAtBroad:
Tried it out and works as described in the comment in PR.
However I would suggest we drop the functionality or go further.
The intent of creating this type of autocomplete  is to try as frequently as possible to allow user to select an option from the list.
And, if needed because it is not there, capture the string entered by the user.
So before the change:
1. Enter "bra"
2. List shows options that match bra (brain category, brain options whatever)
3. User accidentally scrolls (hits arrow or touches the mouse wheel or just touches there Apple Magic mouse.

Previously, the list of options would still be shown, and hopefully, the matching cancer is there.

If you repeat the steps above AFTER the change, the list is gone. 
Now place the cursor in the autocomplete field. The list is still gone.
We have now made it MORE LIKELY that user will NOT select a matching option from the list.

So I think options are to:
1) Not do this feature at all (I really do not see this as a big deal anyways. It is common behavior. But I understand. Opinions).
2. When cursor back on field, we redisplay the matching options.

I think #2 is still not as good as keeping the list open, but it is better.
Please ping back whoever entered this issue and have talk about this. I have my misgivings.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
grigol-brd: I used an arrow function to avoid ugly `this` binding. Will see what else I can do

grigol-brd:
@MocanaAtBroad  I've had a discussion with them and they would still like to have this behavior of hiding options on scroll, but we agreed to add redisplaying options when a user clicks back on the field. Now it works as you described in `#2`.

bf287ef0-1200-4280-a93b-d69c43c38847

MocanaAtBroad:
Nice! 👍

MocanaAtBroad (approved):

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts
```diff
@@ -112,6 +119,22 @@ export class AutocompleteActivityPicklistQuestion extends BaseActivityPicklistQu
...
+        if (this.autocompleteTrigger.panelOpen && !isAutoCompleteOptionsScrolling) {
+            this.autocompleteTrigger.closePanel();
+        }
+    }
+
+    public openAutocompleteOptions(): void {
+        if (!this.autocompleteTrigger.panelOpen) {
+            this.autocompleteTrigger.openPanel();
+        }
     }
```
MocanaAtBroad, line 138: Actually, forget what I said earlier. Just use the arrow function. Looks a little odd when reading the code, but it is preferable to adding another dependency that has not been updated since 2018. 
Sorry for the flip-flopping!

grigol-brd: Sure, will do that in #1111

PR #1102, opened 2022-03-11
DDP-7646: Use DSM token in DSS

alextanlan:
This is a first step for dealing with DSS inside of DSM. The PR is kind of PoC (proof of concept).

 **_Implemented:_** 
Using DSM token for basic work with activities from DSS side. 
Used DSS SDK (Angular) library in order to display activities.

**_Examples:_**

1. **GET** an activity

https://user-images.githubusercontent.com/7396837/160459515-e15f9868-bd53-4eec-9f8d-d49c0313cd3b.mp4

2. **CREATE** an activity instance

https://user-images.githubusercontent.com/7396837/160459535-d29c085d-f711-41b9-b95e-2716dfd7d475.mp4

3. **DELETE** an activity instance

https://user-images.githubusercontent.com/7396837/160459551-dbd2c279-b1b0-45e5-ad88-2a96fdb57471.mp4


ddp-workspace/projects/ddp-dsm-ui/src/app/services/auth.service.ts
```diff
@@ -227,4 +229,13 @@ export class Auth {
...
     }
   }
+
+  private setDssSession(dsmToken: string): void {
+    const accessToken = null;
+    const userGuid = null;
+    const locale = 'en';
+    const expiresAtInSeconds = 1e10; // TODO: get rid of that by using a study config flag as to not renew a session
+    // set DSS Session partially
+    this.dssSessionService.setSession(accessToken, dsmToken, userGuid, locale, expiresAtInSeconds);
```
alextanlan, line 239: The main thing here is set `dsmToken` to DSS session.

ddp-workspace/projects/ddp-dsm-ui/src/app/services/auth.service.ts
```diff
@@ -129,9 +130,10 @@ export class Auth {
       next: response => dsmResponse = response,
       complete: () => {
         const dsmToken = dsmResponse.dsmToken;
-        localStorage.setItem(SessionService.DSM_TOKEN_NAME, dsmToken);
```
alextanlan, line 132: Moved this line to `session.service.ts` (https://github.com/broadinstitute/ddp-angular/pull/1102/files#:~:text=localStorage.setItem(SessionService.DSM_TOKEN_NAME%2C%20token)%3B)

ddp-workspace/projects/ddp-dsm-ui/src/app/test-dss/test-dss.component.html
```diff
@@ -0,0 +1,14 @@
...
+
+<button (click)="getActivity()">GET</button>
+
+<button (click)="fetchActivity()">FETCH</button>
+
+<!--<ddp-activity-redesigned [studyGuid]="studyGuid"-->
+<!--                         [activityGuid]="activityGuid"-->
+<!--                         (submit)="navigate($event)"-->
+<!--                         (activityCode)="activityCodeChanged($event)">-->
+<!--</ddp-activity-redesigned>-->
```
alextanlan, line 14: Should work now for existing `userGuid`, `studyGuid`, `activityGuid` (they are set up in test-dss.component.ts)

ddp-workspace/projects/ddp-sdk/src/lib/services/configuration.service.ts
```diff
@@ -7,6 +7,7 @@ import { MailAddressFormErrorFormatter } from '../models/mailAddressFormErrorFor
 @Injectable()
 export class ConfigurationService {
     backendUrl: string;
+    backendUrlDSS: string;
```
MocanaAtBroad, line 10: I think this is going to be confusing. I BELIEVE you are expecting that we will set one or the other. My advice is to just stick with backendUrl OR refactor the name, but not add second field that is meant to do exactly same thing.

ddp-workspace/projects/ddp-dsm-ui/src/app/services/auth.service.ts
```diff
@@ -227,4 +229,13 @@ export class Auth {
       this.router.navigate([ Statics.HOME_URL ]);
     }
   }
+
+  private setDssSession(dsmToken: string): void {
+    const accessToken = null;
+    const userGuid = null;
+    const locale = 'en';
+    const expiresAtInSeconds = 1e10; // TODO: get rid of that by using a study config flag as to not renew a session
```
MocanaAtBroad: Well, we should extract from the token itself the expiration, no?

alextanlan: I thought a DSS token should never expire for DSM operator/admin
(means that a DSM "admin" can use DSS without time limit). 
Am I wrong?

ddp-workspace/projects/ddp-sdk/src/lib/services/configuration.service.ts
alextanlan: Indeed, you are right. 
I was messed with different config files DSM/DSS.
But `ConfigurationService` is about DSS only.
Fixed in 1bd7bc225a0b38da1731a83e82a6c22b273293e4

ddp-workspace/projects/ddp-dsm-ui/src/app/services/auth.service.ts
MocanaAtBroad: The admin has more privileges to do stuff, but their tokens also expire.
I am under distinct impression that DSM token includes an expiration.
(Which brings up the question of what we should do if the token really has expired. We probably want to do whatever DSM currently does).

alextanlan: Agree, 
fixed in db71d57852287276b3fd264612ed1b2d91eaccb4

MocanaAtBroad (dismissed):

pegahtah (approved):

PR #1107, opened 2022-03-15
DDP-7645/DSM-All-emails-are-marked-invalid

GiCharkviani:
https://broadinstitute.atlassian.net/browse/DDP-7645

### DIFF
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-page/participant-page.component.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-page/participant-page.component.ts
@@ -110,7 +110,7 @@ export class ParticipantPageComponent implements OnInit, OnDestroy, AfterViewChe
   updatingParticipant = false;
   private taskType: string;
   private checkParticipantStatusInterval: any;
-  isEmailValid: boolean = true;
+  isEmailValid: boolean;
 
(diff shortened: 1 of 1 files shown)


ddp-workspace/projects/ddp-dsm-ui/src/app/participant-page/participant-page.component.ts
```diff
@@ -110,7 +110,7 @@ export class ParticipantPageComponent implements OnInit, OnDestroy, AfterViewChe
   updatingParticipant = false;
   private taskType: string;
   private checkParticipantStatusInterval: any;
-  isEmailValid: boolean;
+  isEmailValid: boolean = true;
```
gmakharat: I think it is not good idea to give value `true` directly, cause this variable is used to check whether email is valid or not, there is a special method called `validateEmailInput` in `participant-page.component.ts` which actually checks whether email is valid or not and updates value of `isEmailValid` accordingly, I think it is better to use that method in `ngOnInit`, you can read participant's email from `this.participant.data.profile`.

GiCharkviani: Thank you for your feedback. I did as you suggested.

gmakharat (approved):

PR #1121, opened 2022-03-18
add validation for initial request and display selected answer

josemanuel167:
Display selected answer on remote autocomplete picklist option.
Avoid backend request when query is empty.

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/activity-picklist-remote-auto-complete-options.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/activity-picklist-remote-auto-complete-options.component.ts
@@ -168,7 +168,7 @@ export class ActivityPicklistRemoteAutoCompleteOptionsComponent
         );
     }
 
-    onBlur(value: string, data: SplittedData) {
+    onBlur(value: string, data: SplittedData): void {
         const matchingOption = data.plOptions.find(pl => value.toLocaleLowerCase() === pl.optionLabel.toLocaleLowerCase());
         const plOption = matchingOption ? matchingOption : this.createNotListedOption(data.otherOption);
         this.onValueSelect(plOption);
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/activity-picklist-remote-auto-complete-options.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/activity-picklist-remote-auto-complete-options.component.ts
@@ -155,18 +155,6 @@ export class ActivityPicklistRemoteAutoCompleteOptionsComponent
         return '';
     }
 
-    validateIfMissingOptionShouldBeDisplayed(
-        picklistOptions: ActivityPicklistOption[]
-    ): boolean {
-        return (
-            this.searchValue$.value.length &&
-            !picklistOptions.filter((pl) =>
-                pl.optionLabel
-                    .toLocaleLowerCase()
-                    .includes(this.searchValue$.value.toLocaleLowerCase())
-            ).length
-        );
-    }
 
     onBlur(value: string, data: SplittedData): void {
         const matchingOption = data.plOptions.find(pl => value.toLocaleLowerCase() === pl.optionLabel.toLocaleLowerCase());
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/activity-picklist-remote-auto-complete-options.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/activity-picklist-remote-auto-complete-options.component.ts
@@ -1,20 +1,21 @@
-import { Component, Inject, Input, OnInit, SimpleChanges } from '@angular/core';
+import { Component, Input, OnInit } from '@angular/core';
 import {
     BehaviorSubject,
     debounceTime,
     distinctUntilChanged,
     map,
     Observable,
-    of,
     switchMap,
-    tap,
 } from 'rxjs';
-import { ActivityPicklistNormalizedGroup } from '../../../models/activity/activityPicklistNormalizedGroup';
 import { ActivityPicklistOption } from '../../../models/activity/activityPicklistOption';
-import { ActivityPicklistQuestionBlock } from '../../../models/activity/activityPicklistQuestionBlock';
 import { NGXTranslateService } from '../../../services/internationalization/ngxTranslate.service';
 import { ActivityServiceAgent } from '../../../services/serviceAgents/activityServiceAgent.service';
 import { BaseActivityPicklistQuestion } from './baseActivityPicklistQuestion.component';
+
+interface SplittedData {
+    plOptions: ActivityPicklistOption[];
+    otherOption: ActivityPicklistOption;
+}
 
 @Component({
     selector: 'ddp-activity-picklist-remote-auto-complete-options',
@@ -29,7 +30,6 @@ import { BaseActivityPicklistQuestion } from './baseActivityPicklistQuestion.com
             [placeholder]="block.picklistLabel"
             [matAutocomplete]="autoCompleteFromSource"
             (keyup)="onInput($event.target.value)"
-            [ngModel]="getInitialAnswerOptionLabel()"
         />
 
         <mat-autocomplete
@@ -86,21 +86,45 @@ export class ActivityPicklistRemoteAutoCompleteOptionsComponent
     }
 
     ngOnInit(): void {
+        const initSearchValue = this.getAnswer();
+        this.searchValue$ = new BehaviorSubject(initSearchValue);
+
         this.picklistOptions$ = this.searchValue$.pipe(
-            debounceTime(1000),
+            debounceTime(500),
             distinctUntilChanged(),
             switchMap((searchValue) =>
-                searchValue.length > 0
-                    ? this.activityService.getPickListOptions(
-                          this.block.stableId,
-                          searchValue,
-                          this.studyGuid,
-                          this.activityGuid
-                      )
-                    : of({ results: [] })
+                this.activityService.getPickListOptions(
+                    this.block.stableId,
+                    searchValue,
+                    this.studyGuid,
+                    this.activityGuid
+                )
             ),
-            map((data) => data.results)
+            map((data) => this.splitOtherOptionFromPLOptions(data.results)),
+            map((splittedData) =>
+                this.validateIfMissingOptionShouldBeDisplayed(
+                    splittedData.plOptions
+                )
+                    ? this.createNotListedOption(splittedData)
+                    : splittedData.plOptions
+            )
         );
+    }
+    splitOtherOptionFromPLOptions(
+        data: ActivityPicklistOption[]
+    ): SplittedData {
+        return {
+            plOptions: data.filter((pl) => pl.stableId !== 'OTHER'),
+            otherOption: data.find((pl) => pl.stableId === 'OTHER'),
+        };
+    }
+
+    createNotListedOption(data: SplittedData) {
+        const notListedOption = {
+            ...data.otherOption,
+            optionLabel: this.searchValue$.value.toLocaleUpperCase(),
+        };
+        return [...data.plOptions, notListedOption];
     }
 
     onInput(value): void {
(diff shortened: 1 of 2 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/activity-picklist-remote-auto-complete-options.component.ts
```diff
@@ -85,15 +87,17 @@ export class ActivityPicklistRemoteAutoCompleteOptionsComponent
 
     ngOnInit(): void {
         this.picklistOptions$ = this.searchValue$.pipe(
-            debounceTime(500),
+            debounceTime(1000),
```
MocanaAtBroad: 1 second is kind of a long time! stay with 500 or even go down to 300. But you can try it and set a value that makes sense experimentally.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/activity-picklist-remote-auto-complete-options.component.ts
```diff
@@ -114,4 +118,13 @@ export class ActivityPicklistRemoteAutoCompleteOptionsComponent
         this.block.answer = value ? [{ stableId: value.stableId, detail }] : [];
         this.valueChanged.emit([...this.block.answer]);
     }
+
+    getInitialAnswerOptionLabel(): string {
+        if (
+            this.block.selectMode === 'SINGLE' &&
+            this.block.picklistOptions.length
+        ) {
+            return this.block.picklistOptions[0].optionLabel;
```
MocanaAtBroad, line 147: Actually, it is a little more complicated than this.
Need to deal with case where the answer that was saved is just the text entered.
Remember the example:
What fruit do you like?
Apples
Oranges
Other ___________

You are not handling the case where somebody typed an option.
The text would come in that details property in the answer that we have been talking about.

From the very similar component, this is how she was initializing the answer:

https://github.com/broadinstitute/ddp-angular/blob/cfffe1b69f2f9f8c190978f3d9ba3673719d929c/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/autocompleteActivityPicklistQuestion.component.ts#L138-L147

MocanaAtBroad: I guess your code was not looking for situation where user types an option that does not exist.
Behavior we want is to include as stableid an option that has a property of allowDetails = true and include in the answer the property detail: THEVALUEENTERED.
Not seeing this at all in the code.
But I get the impression that our remote API does not account for it.
@ssettipalli Can you work with Jose to nail down these details?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/activity-picklist-remote-auto-complete-options.component.ts
```diff
@@ -28,6 +29,7 @@ import { BaseActivityPicklistQuestion } from './baseActivityPicklistQuestion.com
             [placeholder]="block.picklistLabel"
             [matAutocomplete]="autoCompleteFromSource"
             (keyup)="onInput($event.target.value)"
+            [ngModel]="getInitialAnswerOptionLabel()"
```
MocanaAtBroad: Let's try to avoid using ngModel. The rest of the code is in the style that of using component events and rxjs to handle input and output. I don't know what this is supposed to do. Please figure out how we can initialize by setting the value for component directly.

alextanlan: As an option you can init the BehaviourSubject (`searchValue$`) in `ngOnInit` method.
Like:
```
  ngOnInit(): void {
        const initSearchValue = ...;
        this.searchValue$ = new BehaviorSubject(initSearchValue);
        ...
}
```

alextanlan:
```
 template: `<mat-form-field
        class="full-width ddp-autocomplete-field"
        *ngIf="picklistOptions$ | async as picklistOptions"  
    >
```
    
Is the ***ngIf** above needed   (line 23) ? 
Seems like
- it is always true because `picklistOptions` is an empty array by default
- the question is hidden until we have picklistOptions$ (e.g. send a request to the server and wait for the response)

alextanlan:
A corner case:
- type `Alpha`
- reach any autocomplete option by keyboard arrows and select it by Enter 
- an excess request is sent (with the selected option)

scr1_1

alextanlan:
**_#minor_**

There are several unused imports in the `activity-picklist-remote-auto-complete-options.component.ts` file

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/activity-picklist-remote-auto-complete-options.component.ts
josemanuel167: Reverted to 500

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/activity-picklist-remote-auto-complete-options.component.ts
josemanuel167: Thank you

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/activity-picklist-remote-auto-complete-options.component.ts
```diff
@@ -95,9 +100,32 @@ export class ActivityPicklistRemoteAutoCompleteOptionsComponent
...
+                    : splittedData.plOptions
+            )
         );
     }
+    splitOtherOptionFromPLOptions(
+        data: ActivityPicklistOption[]
+    ): SplittedData {
+        return {
+            plOptions: data.filter((pl) => pl.stableId !== 'OTHER'),
+            otherOption: data.find((pl) => pl.stableId === 'OTHER'),
```
MocanaAtBroad: The filter criteria should be to to check that pl.allowDetails is true. I think it is OK to expect there will only be (at most) one option that has this allowDetails == true.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/activity-picklist-remote-auto-complete-options.component.ts
```diff
@@ -111,7 +139,49 @@ export class ActivityPicklistRemoteAutoCompleteOptionsComponent
...
+            this.block.selectMode === 'SINGLE' &&
+            this.block.picklistOptions.length
+        ) {
+            return this.block.picklistOptions[0].optionLabel;
+        }
+    }
+
+    getAnswer(): string {
+        return this.block.answer && this.block.answer[0]
+            ? this.block.answer[0].stableId !== 'OTHER'
```
MocanaAtBroad: stableId is not the criteria for the option. It is "allowDetails" is true.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/activity-picklist-remote-auto-complete-options.component.ts
```diff
@@ -95,9 +100,32 @@ export class ActivityPicklistRemoteAutoCompleteOptionsComponent
                     this.activityGuid
                 )
```
MocanaAtBroad, line 102: Can you change this function so that searchValue is the last argument.
Preferred order is studyGuid, activityGuid, stableId and last searchValue .
(the first two have the largest scope, the last two the least scope). It is a style thing, not a correctness issue.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/activity-picklist-remote-auto-complete-options.component.ts
```diff
@@ -95,9 +100,32 @@ export class ActivityPicklistRemoteAutoCompleteOptionsComponent
...
             ),
-            map((data) => data.results)
+            map((data) => this.splitOtherOptionFromPLOptions(data.results)),
+            map((splittedData) =>
+                this.validateIfMissingOptionShouldBeDisplayed(
+                    splittedData.plOptions
+                )
+                    ? this.createNotListedOption(splittedData)
+                    : splittedData.plOptions
+            )
```
MocanaAtBroad, line 106: You might have to explain this to me as I don't understand it: why would we not always show all the options that come from the server, just making sure we exclude those that are "allowDetails = true"?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/activity-picklist-remote-auto-complete-options.component.ts
```diff
@@ -111,7 +139,49 @@ export class ActivityPicklistRemoteAutoCompleteOptionsComponent
         value: ActivityPicklistOption,
         detail: string | null = null
     ): void {
-        this.block.answer = value ? [{ stableId: value.stableId, detail }] : [];
+        this.block.answer = value
+            ? [
+                  {
+                      stableId: value.stableId,
+                      detail:
+                          value.stableId === 'OTHER'
```
MocanaAtBroad: OnValueSelect appears to be called when an option is selected. 
I don't see why we would ever care about it's optionLabel.
For autocomplete we will either set the stableid of the option as the answer.
Or if there is no matching option, we will use the stableId of the allowDetails=true option and set the detail property to the entered string.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/activity-picklist-remote-auto-complete-options.component.ts
```diff
@@ -95,9 +100,32 @@ export class ActivityPicklistRemoteAutoCompleteOptionsComponent
...
+    splitOtherOptionFromPLOptions(
+        data: ActivityPicklistOption[]
+    ): SplittedData {
+        return {
+            plOptions: data.filter((pl) => pl.stableId !== 'OTHER'),
+            otherOption: data.find((pl) => pl.stableId === 'OTHER'),
+        };
+    }
+
+    createNotListedOption(data: SplittedData): ActivityPicklistOption[] {
```
MocanaAtBroad: I don't se why this is necessary either....

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/activity-picklist-remote-auto-complete-options.component.ts
josemanuel167: Extra code removed

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/activity-picklist-remote-auto-complete-options.component.ts
josemanuel167: Sure, on it

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/activity-picklist-remote-auto-complete-options.component.ts
```diff
@@ -1,20 +1,22 @@
...
 import { ActivityPicklistOption } from '../../../models/activity/activityPicklistOption';
 import { NGXTranslateService } from '../../../services/internationalization/ngxTranslate.service';
 import { ActivityServiceAgent } from '../../../services/serviceAgents/activityServiceAgent.service';
 import { BaseActivityPicklistQuestion } from './baseActivityPicklistQuestion.component';
 
+interface SplittedData {
+    plOptions: ActivityPicklistOption[];
+    otherOption: ActivityPicklistOption;
+}
+
```
MocanaAtBroad, line 19: Le't rename this something like CategorizedPicklistOptions
instead of plOptions, let's call it currentAutoCompleteOptions .
otherOption, let's call it userEnteredStringOption

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/activityServiceAgent.service.ts
```diff
@@ -146,14 +146,14 @@ export class ActivityServiceAgent extends UserServiceAgent<any> {
     }
 
     getPickListOptions(
-        questionStableId: string,
-        query: string = '',
         studyGuid: string,
-        activityGuid: string
+        activityGuid: string,
+        stableId: string,
```
MocanaAtBroad:
```suggestion
        questionStableId: string,
```

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/activity-picklist-remote-auto-complete-options.component.ts
josemanuel167: Sounds good, renamed

MocanaAtBroad (approved):

PR #1139, opened 2022-03-28
DDP-7684/CMI-Pancan-DSM-family-history-survey-display-family-members-…

GiCharkviani:
https://broadinstitute.atlassian.net/browse/DDP-7684


ddp-workspace/projects/ddp-dsm-ui/src/app/participant-page/participant-page.component.ts
```diff
@@ -186,6 +188,19 @@ export class ParticipantPageComponent implements OnInit, OnDestroy, AfterViewChe
     clearInterval(this.checkParticipantStatusInterval);
   }
 
+  displayAtivityOrder(): void {
+    const orderedActivities = [];
```
gmakharat, line 192: do we need an auxiliary array? I think in `this.participant,data.activities` we get only chosen study's activity definitions and what if we sort it directly?

gmakharat (approved):

PR #1140, opened 2022-03-28
DDP-7322: singular form

SimoneMaiwald:
adding some logic to display the singular form

checkboxes on the from are not working yet!


ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/participant-list.component.ts
```diff
@@ -2127,7 +2127,7 @@ export class ParticipantListComponent implements OnInit {
   private createFilter(field: any): Filter {
     let showType = field.displayType;
     let filter: Filter = new Filter(
-      new ParticipantColumn(field.columnDisplay.replace('*', ''), field.columnName, 'participantData', field.fieldType, false),
+      new ParticipantColumn(field.columnDisplay?.replace('*', ''), field.columnName, 'participantData', field.fieldType, false),
```
SimoneMaiwald, line 2120: Not sure where that `?` came from… @GiCharkviani is that ok here or will that break things?

ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/participant-list.component.html
```diff
@@ -220,9 +220,9 @@ <h3 class="Color--primary Line--Break">{{additionalMessage}}</h3>
           <br/>
         </div>
 
-        <app-search-bar [textQuery]="filterQuery" [allFieldNames]="allFieldNames" (queryText)="doFilterByQuery($event)"
-                        [showHelp]="showHelp" [filters]="sourceColumns">
-        </app-search-bar>
+<!--        <app-search-bar [textQuery]="filterQuery" [allFieldNames]="allFieldNames" (queryText)="doFilterByQuery($event)"-->
+<!--                        [showHelp]="showHelp" [filters]="sourceColumns">-->
+<!--        </app-search-bar>-->
```
pegahtah: I thought we are removing that

SimoneMaiwald: commenting it out for now, in case we want to add it later

ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/participant-list.component.ts
GiCharkviani: I have added that sign, just for any case, if columnDisplay appears undefined, not to crush the app. That is called ["Optional Chaining"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

GiCharkviani (approved):

PR #1146, opened 2022-03-30
DDP-7721: fix: family history error message re-render

grigol-brd:
https://broadinstitute.atlassian.net/browse/DDP-7721

### DIFF
--- a/ddp-workspace/projects/ddp-pancan/src/app/components/activity/activity.component.ts
+++ b/ddp-workspace/projects/ddp-pancan/src/app/components/activity/activity.component.ts
@@ -6,17 +6,16 @@ import {
     ActivityRedesignedComponent,
     SubmissionManager,
     SubmitAnnouncementService,
-    ActivitySection,
     ActivityActivityBlock,
     ActivityRenderHintType,
     LoggingService,
     WindowRef,
     AnalyticsEventsService,
-    ParticipantsSearchServiceAgent
+    ParticipantsSearchServiceAgent,
+    ActivityActionsAgent,
+    ActivityBlockType,
 } from 'ddp-sdk';
 import { Subscription } from 'rxjs';
-
-type Block = ActivitySection['blocks'][number]
 
 @Component({
     selector: 'app-activity',
@@ -24,10 +23,13 @@ type Block = ActivitySection['blocks'][number]
     providers: [SubmitAnnouncementService, SubmissionManager],
 })
 export class ActivityComponent extends ActivityRedesignedComponent implements OnInit, OnDestroy {
+    showEmptyNestedActivityError = false;
     dialogsClosedSubscription: Subscription;
+    activityInstanceDeletedSubscription: Subscription;
 
     constructor(
-        public editDialog: MatDialog,
+        private editDialog: MatDialog,
+        private activityActionsAgent: ActivityActionsAgent,
         logger: LoggingService,
         windowRef: WindowRef,
         changeRef: ChangeDetectorRef,
@@ -44,20 +46,23 @@ export class ActivityComponent extends ActivityRedesignedComponent implements On
     public ngOnInit(): void {
         super.ngOnInit();
 
-        this.dialogsClosedSubscription = this.editDialog.afterAllClosed.subscribe(() => {
-            this.showEmptyNestedActivityError = !this.hasNoEmptyNestedActivity;
-            setTimeout(() => this.showEmptyNestedActivityError = !this.hasNoEmptyNestedActivity, 200); // quick and dirty solution to re-check, in case activity was deleted a bit later
-        });
+        this.dialogsClosedSubscription = this.editDialog.afterAllClosed.subscribe(this.updateErrorMessageDisplayState);
+
+        this.activityInstanceDeletedSubscription =
+            this.activityActionsAgent.activityBlockInstancesUpdated.subscribe(this.updateErrorMessageDisplayState);
     }
 
     public ngOnDestroy(): void {
         this.dialogsClosedSubscription.unsubscribe();
+        this.activityInstanceDeletedSubscription.unsubscribe();
     }
 
-    showEmptyNestedActivityError = false;
+    public updateErrorMessageDisplayState = (): void => {
+        this.showEmptyNestedActivityError = !this.allNestedActivitiesHaveAnswers;
+    };
 
     public incrementStep(): void {
-        if (this.hasNoEmptyNestedActivity) {
+        if (this.allNestedActivitiesHaveAnswers) {
             this.showEmptyNestedActivityError = false;
             super.incrementStep();
         } else {
@@ -65,20 +70,16 @@ export class ActivityComponent extends ActivityRedesignedComponent implements On
         }
     }
 
-    get hasNoEmptyNestedActivity(): boolean {
-        try {
-            const enabledActivityBlocks = this.currentSection.blocks.filter(this.isActiveModalActivityBlock) as ActivityActivityBlock[];
+    get allNestedActivitiesHaveAnswers(): boolean {
+        const enabledActivityBlocks = this.currentSection.blocks.filter(this.isActiveModalActivityBlock) as ActivityActivityBlock[];
 
-            return enabledActivityBlocks.every(this.blockHasAnswers);
-        } catch {
-            return true;
-        }
+        return enabledActivityBlocks.every(this.blockHasAnswers);
     }
 
-    private isActiveModalActivityBlock = (block: ActivityActivityBlock): boolean =>
+    private isActiveModalActivityBlock = (block: ActivityBlockType): boolean =>
         this.isActivityBlock(block) ? block.enabled && block.renderHint === ActivityRenderHintType.Modal : false;
 
-    private isActivityBlock(block: Block): block is ActivityActivityBlock {
+    private isActivityBlock(block: ActivityBlockType): block is ActivityActivityBlock {
         return 'activityCode' in block && typeof block.activityCode === 'string';
     }
 
--- a/ddp-workspace/projects/ddp-pancan/src/app/components/activity/activity.component.ts
+++ b/ddp-workspace/projects/ddp-pancan/src/app/components/activity/activity.component.ts
@@ -1,4 +1,6 @@
-import { Component } from '@angular/core';
+import { DOCUMENT } from '@angular/common';
+import { ChangeDetectorRef, Component, Inject, Injector, OnDestroy, OnInit, Renderer2 } from '@angular/core';
+import { MatDialog } from '@angular/material/dialog';
 
 import {
     ActivityRedesignedComponent,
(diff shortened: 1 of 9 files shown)


GiCharkviani (dismissed):

ddp-workspace/projects/ddp-pancan/src/app/components/activity/activity.component.ts
```diff
@@ -16,7 +23,37 @@ type Block = ActivitySection['blocks'][number]
...
+        super.ngOnInit();
+
+        this.dialogsClosedSubscription = this.editDialog.afterAllClosed.subscribe(() => {
+            this.showEmptyNestedActivityError = !this.hasNoEmptyNestedActivity;
+            setTimeout(() => this.showEmptyNestedActivityError = !this.hasNoEmptyNestedActivity, 200); // quick and dirty solution to re-check, in case activity was deleted a bit later
+        });
+    }
+
+    public ngOnDestroy(): void {
+        this.dialogsClosedSubscription.unsubscribe();
```
alextanlan, line 54: No need to unsubscribe from afterClosed() as it auto completes itself. 
See [Angular material dialog](https://github.com/angular/components/blob/ae41a0ad69ca26c600f0f56c68dd5a1c102d4f1f/src/lib/dialog/dialog-ref.ts#L75), [stackoverflow](https://stackoverflow.com/questions/58198544/angular-dialogref-unsubscribe-do-i-need-to-unsubscribe-from-afterclosed)

ddp-workspace/projects/ddp-pancan/src/app/components/activity/activity.component.ts
```diff
@@ -29,9 +66,13 @@ export class ActivityComponent extends ActivityRedesignedComponent {
     }
 
     get hasNoEmptyNestedActivity(): boolean {
-        const enabledActivityBlocks = this.currentSection.blocks.filter(this.isActiveModalActivityBlock) as ActivityActivityBlock[];
+        try {
+            const enabledActivityBlocks = this.currentSection.blocks.filter(this.isActiveModalActivityBlock) as ActivityActivityBlock[];
 
-        return enabledActivityBlocks.every(this.blockHasAnswers);
+            return enabledActivityBlocks.every(this.blockHasAnswers);
+        } catch {
```
alextanlan: Just wondering what kind of error can be caught there ?

ddp-workspace/projects/ddp-pancan/src/app/components/activity/activity.component.ts
```diff
@@ -16,7 +23,37 @@ type Block = ActivitySection['blocks'][number]
...
+    ) {
+        super(logger, windowRef, changeRef, renderer, submitService, analytics, participantsSearch, document, injector);
+    }
+
+    public ngOnInit(): void {
+        super.ngOnInit();
+
+        this.dialogsClosedSubscription = this.editDialog.afterAllClosed.subscribe(() => {
+            this.showEmptyNestedActivityError = !this.hasNoEmptyNestedActivity;
+            setTimeout(() => this.showEmptyNestedActivityError = !this.hasNoEmptyNestedActivity, 200); // quick and dirty solution to re-check, in case activity was deleted a bit later
```
alextanlan: Seems like a weird hack.
It won't work if an activity is removed with a little delay.

ddp-workspace/projects/ddp-pancan/src/app/components/activity/activity.component.ts
```diff
@@ -16,7 +23,37 @@ type Block = ActivitySection['blocks'][number]
...
+        });
+    }
+
+    public ngOnDestroy(): void {
+        this.dialogsClosedSubscription.unsubscribe();
+    }
+
     showEmptyNestedActivityError = false;
```
alextanlan: it would be nice to keep class properties at the beginning of the class.

ddp-workspace/projects/ddp-pancan/src/app/components/activity/activity.component.ts
grigol-brd: It's an error referencing `this.currentSection` causes at times, I haven't investigated the root, but I'll look into it

ddp-workspace/projects/ddp-pancan/src/app/components/activity/activity.component.ts
grigol-brd: Yes, it is weird. In some cases removing activity happens after dialogs are closed and that's why I had to add a delay there.
In order for it to be fixed the non-hacky way I think it'll need changes in SDK

alextanlan:
**_#optional_**

Line 82:
 `return 'activityCode' in block && typeof block.activityCode === 'string';`
 can be shortened like 
`typeof block.activityCode === 'string';`

ddp-workspace/projects/ddp-pancan/src/app/components/activity/activity.component.ts
alextanlan: This trick is helpless.
For example in Pancan Family History:
- add a Parent's Sibling
- open and close without editing
- an error message appears ('Please enter information for the family member above or remove the empty entry')
- wait a second and remove the Parent's Sibling instance
- the error message is still there

alextanlan: This setTimeout does not help.
There should be more proper way to solve it.

grigol-brd: Yep, that's a con of timeout, works for me but may not for others. Anyway, I'll get rid of it

ddp-workspace/projects/ddp-pancan/src/app/components/activity/activity.component.ts
alextanlan: Using try/catch here is not a nice idea, imo.
What about just checking if `this.currentSection` exists ? (even though I can't see a case when it does not)

ddp-workspace/projects/ddp-pancan/src/app/components/activity/activity.component.ts
```diff
@@ -29,9 +66,13 @@ export class ActivityComponent extends ActivityRedesignedComponent {
     }
 
     get hasNoEmptyNestedActivity(): boolean {
```
alextanlan: Honestly, I am a bit confused with the method name.

Does it mean 'There are some **not empty** activities' or most likely 'There **are not** any empty activity' ? 

May be rename it something like 'allNestedActivitiesAreEmpty' / 'allNestedActivitiesAreNotEmpty' / 'allNestedActivitiesHaveAnswers' - or kind of ?

grigol-brd:
> **_#optional_**
> 
> Line 82: `return 'activityCode' in block && typeof block.activityCode === 'string';` can be shortened like `typeof block.activityCode === 'string';`

@alextanlan That's a type guard for `Block` since it's a union type and only one of them has `activityCode ` property.

image

ddp-workspace/projects/ddp-pancan/src/app/components/activity/activity.component.ts
grigol-brd: It means "there are not any empty nested activities". I think `allNestedActivitiesHaveAnswers` sounds better

alextanlan:
> That's a type guard for Block since it's a union type and only one of them has activityCode property.

1) Why do we need this ` type Block` at all? 

image

The method `this.isActivityBlock(block)` is invoked with `block: block: ActivityActivityBlock` (line 79).
ActivityActivityBlock class has an `activityCode` property. 



2) The method `isActivityBlock` returns **boolean** (not `block is ActivityActivityBlock`)

grigol-brd:
> > That's a type guard for Block since it's a union type and only one of them has activityCode property.
> 
> 1. Why do we need this ` type Block` at all?
> 
> image
> 
> The method `this.isActivityBlock(block)` is invoked with `block: block: ActivityActivityBlock` (line 79). ActivityActivityBlock class has an `activityCode` property.
> 
> 2. The method `isActivityBlock` returns **boolean** (not `block is ActivityActivityBlock`)

1. Well turns out that's a mistake and `isActiveModalActivityBlock` should have `(block: Block)` and not `(block: ActivityActivityBlock)`

2. As you know, it's a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) and common for user-defined type guards, it serves a different purpose than returning just a boolean. TS complains if I change it to a boolean:
image

ddp-workspace/projects/ddp-pancan/src/app/components/activity/activity.component.ts
grigol-brd: The issue was coming from the base class and I've fixed it

alextanlan:
1. > As you know, it's a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)

I might not have been aware of it. I never used it. 
Got your idea about a user-defined type guard, thank you!

2. Seems like `Block` type is a duplicate of `ActivityBlock` type (ddp-sdk\src\lib\models\activity\activitySection.ts).
 Is not it ?

grigol-brd:
> 2. Seems like `Block` type is a duplicate of `ActivityBlock` type (ddp-sdk\src\lib\models\activity\activitySection.ts).
>    Is not it ?

2. It is and I thought exposing it from SDK, but on the other hand I'm not sure how good of an idea that is because there's an abstract class with the exact same name (`ddp-workspace\projects\ddp-sdk\src\lib\models\activity\activityBlock.ts`) and I don't want them to become conflicting. That's why I've extracted it locally:
image
WDYT should I still make `ActivityBlock` (the one in `ddp-sdk\src\lib\models\activity\activitySection.ts`) public?

ddp-workspace/projects/ddp-pancan/src/app/components/activity/activity.component.ts
alextanlan: In this case using try/catch is  a hiding of that issue.
 Like: 
```
 something went wrong (potentially with `this.currentSection`) 
   -> never mind, let's decide that all blocks in the current section have answers 
     -> toggle showEmptyNestedActivityError to false in order to remove an error message
```
     
At least the issue should be logged as to fix it later. 

> The issue was coming from the base class 

It is strange. `currentSection` from the base class is used in many places and never failed before. May be it is a reason to create a bug ticket as to investigate it more ?

alextanlan:
I'd vote for using one type/interface only.
Let's use `ActivityBlock` type (from ActivitySection.ts), move it to a separate file, make it public and rename e.g. `ActivityBlockType`, shall we ?

grigol-brd:
> I'd vote for using one type/interface only. Let's use `ActivityBlock` type (from ActivitySection.ts), move it to a separate file, make it public and rename e.g. `ActivityBlockType`, shall we ?

Yea that would be safer, I believe, since it's declared locally right now.

alextanlan:
Looks cool, thank you!

alextanlan:
One more comment only: 
Tests launch is not successful in your branch. 
(take a look at `ddp-sdk/src/lib/components/activityForm/activity-blocks/activityBlock/activityBlock.component.spec.ts`)

grigol-brd:
> One more comment only: Tests launch is not successful in your branch. (take a look at `ddp-sdk/src/lib/components/activityForm/activity-blocks/activityBlock/activityBlock.component.spec.ts`)

Found it, looks like automatic import has messed up one of the paths.

Edit: But looks like there's another error

Edit: Okay all should be fixed now

GiCharkviani (approved):

PR #1157, opened 2022-04-01
DDP-7379: fix: added osteo family history version

grigol-brd:
https://broadinstitute.atlassian.net/browse/DDP-7379


archilsharashenidzebroad (approved):
LGTM

PR #1165, opened 2022-04-04
DDP-7740: Fix - INTEGER type questions should not allow decimals to be inputted

alextanlan:


ddp-workspace/projects/ddp-sdk/src/lib/directives/inputRestrictionDirective.directive.ts
```diff
@@ -82,11 +84,15 @@ export class InputRestrictionDirective {
         return (event.ctrlKey === true || event.metaKey === true);
     }
 
-    private integerOnly(e: KeyboardEvent): void {
+    private integerOnly(e: KeyboardEvent, onlyPositiveNumbers: boolean): void {
```
alextanlan, line 87: `onlyPositiveNumbers` means that digits (0-9) only allowed.

MocanaAtBroad (approved):

PR #1202, opened 2022-04-13
Feature/boolean question

josemanuel167:
Add boolean checkbox question into existing boolean type

PR #1205, opened 2022-04-14
fix: undefined `renderMode` causing boolean options not be rendered

grigol-brd:
Fixes bug introduced in #1202
The bug was caused by `block.renderMode` being `undefined` causing the default radio buttons not to be rendered.
Decided to mark `renderMode` as an optional property since it can be `undefined`.

https://broadinstitute.atlassian.net/browse/DDP-7833

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityBooleanQuestionBlock.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityBooleanQuestionBlock.ts
@@ -6,7 +6,7 @@ import { BooleanRenderMode } from './booleanRenderMode';
 export class ActivityBooleanQuestionBlock extends ActivityQuestionBlock<boolean> {
     public trueContent: string;
     public falseContent: string;
-    public renderMode: BooleanRenderMode;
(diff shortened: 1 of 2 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityBooleanQuestionBlock.ts
```diff
@@ -6,7 +6,7 @@ import { BooleanRenderMode } from './booleanRenderMode';
 export class ActivityBooleanQuestionBlock extends ActivityQuestionBlock<boolean> {
     public trueContent: string;
     public falseContent: string;
-    public renderMode: BooleanRenderMode;
+    public renderMode?: BooleanRenderMode;
```
alextanlan, line 9: the `renderMode` will be mandatory soon. 
[PR#1202](https://github.com/broadinstitute/ddp-angular/pull/1202) accidentally was merged before the back-end implementation.

alextanlan:
Thank you for this catch and quick fix!

alextanlan (approved):

PR #1221, opened 2022-04-21
|DDP-7796: feat: osteo family history shows error when fields are empty

grigol-brd:
https://broadinstitute.atlassian.net/browse/DDP-7796


nsakande (approved):
LGTM

PR #1222, opened 2022-04-25
DDP-7876: Equation question rendering

alextanlan:
Implemented new Equation (derived) questions rendering.
The question value is received from back-end side according to the related input values (see [DDP-7183 description](https://broadinstitute.atlassian.net/browse/DDP-7183))
The value is read-only and should not be changed/patched.

Screenshots:
sn2

sn1


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activitySection.component.ts
```diff
@@ -66,7 +67,17 @@ export class ActivitySectionComponent implements OnInit, OnDestroy {
...
+        this.embeddedValidationStatus.set(id, isValid);
+        const reducedValidationStatus = Array.from(this.embeddedValidationStatus.values()).every(value => value);
+        this.embeddedComponentsValidationStatus.next(reducedValidationStatus);
+    }
+
+    public isCertainTypeOfBlock(block: ActivityBlock, type: BlockType): boolean {
+        return block.blockType === type && block.shown;
+    }
+
+    private updateVisibilityAndValidation(visibility: BlockVisibility[]): void {
```
alextanlan, line 75: Changed `public` to `private` because it is used in the file only.

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/answerValue.ts
```diff
@@ -6,6 +6,7 @@ import { NumericAnswerType } from './numericAnswerType';
 export type AnswerValue =
   | boolean
   | string
+  | string[]
```
alextanlan, line 9: Fix for changed earlier` FileAnswer` (not related to the current feature).

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/activityEquationQuestionBlock.ts
```diff
@@ -0,0 +1,20 @@
...
+
+    constructor() {
+        super();
+    }
+
+    public get questionType(): QuestionType {
+        return QuestionType.Equation;
+    }
+
+    public hasAnswer(): boolean {
```
josemanuel167, line 16: Could be safe to remove it'

alextanlan: Sorry, I did not get the point. 
What do you mean, @josemanuel167 ?

`ActivityEquationQuestionBlock` class  should implement inherited abstract `hasAnswer` method 
from its parent abstract class `ActivityQuestionBlock`.

josemanuel167 (dismissed):

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityQuestion.component.ts
```diff
@@ -19,6 +19,7 @@ import { ActivityValidationResult } from '../../../models/activity/activityValid
                 [validationRequested]="validationRequested$ | async"
                 [studyGuid]="studyGuid"
                 [activityGuid]="activityGuid"
+                [triggerChanges]="triggerChanges"
```
MocanaAtBroad: Why is this needed?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activitySection.component.ts
```diff
@@ -66,7 +76,17 @@ export class ActivitySectionComponent implements OnInit, OnDestroy {
...
     }
 
-    public updateVisibilityAndValidation(visibility: BlockVisibility[]): void {
+    public updateValidationStatusInSection(id: string, isValid: boolean): void {
+        this.embeddedValidationStatus.set(id, isValid);
+        const reducedValidationStatus = Array.from(this.embeddedValidationStatus.values()).every(value => value);
+        this.embeddedComponentsValidationStatus.next(reducedValidationStatus);
+    }
+
+    public isCertainTypeOfBlock(block: ActivityBlock, type: BlockType): boolean {
```
MocanaAtBroad: Maybe rename to something like shouldBlockBeShown ? (isCertainTypeOfBlock suggests that it is only the type that matters, but block.shown matters too).

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activitySection.component.ts
```diff
@@ -90,44 +110,67 @@ export class ActivitySectionComponent implements OnInit, OnDestroy {
...
+    private updateValidationForHiddenEmbeddedActivity(block: ActivityActivityBlock): void {
+        block.instances.forEach((instance: ActivityInstance) => {
+            this.updateValidationStatusInSection(this.idPrefix.instance + instance.instanceGuid, true);
+        });
     }
 
-    public isMailAddress(block: ActivityBlock): block is ActivityInstitutionBlock {
-        return block.blockType === BlockType.MailAddress && block.shown;
+    private updateEquationQuestions(equations: AnswerResponseEquation[]): void {
+        const equationQuestionBlocks = this.section.allChildBlocks()
```
MocanaAtBroad: IMO it would be preferable if this component did not have any logic specific to type of the blocks that are contained by it. (I see we had some logic already here regarding activity blocks. Not great/we should revisit that. But this problem for now).
I think the issue is that we need to get the equation results/values down to the correct block/component, right?
This param value `equations` is being extracted from the response via a SubmissionManager subscription.
Perhaps a way to avoid having Equation specific logic in this component would be to have the Equation component or the EquationBlock subscribe the SubmissionManager directly?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activitySection.component.ts
alextanlan: Fixed in be820a94cb240af38838e9f6b81bded1d6c6d45f

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityQuestion.component.ts
alextanlan: There was an issue with updating equations inside of a composite question thus I added a hack for trigger updating.
Anyway, it is useless now. I have removed it.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activitySection.component.ts
alextanlan: My bad, I overcomplicated the logic.
Keeping an equation question tuned with submission answer response is nice idea, thank you!

In addition we should have similar logic in a composite question with some tweaks (we can't update an equation itself inside of a composite. But we should update the composite answer and re-render the composite question).
Nevertheless, it is better than my previous solution.

Fixed in f3a66d7475dd354e4bfa593f67b808a66c9dc60f

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-composite-answer/activityCompositeAnswer.component.ts
```diff
@@ -42,38 +62,65 @@ export class ActivityCompositeAnswer implements OnChanges {
...
+                        // eslint-disable-next-line arrow-body-style
+                        answerRow.map((answer: AnswerContainer) => {
+                            return (answer.stableId === question.stableId) ? {
+                                ...answer,
+                                value: [equation.values[index]]
+                            } : answer;
+                        })
+                    );
+                    this.block.setAnswer(newAnswer, false);
+                    this.childQuestionBlocks = this.rebuildChildQuestions(this.block, newAnswer);
```
MocanaAtBroad: Handling the server responses to  update equation here is definitely better than in the section component.
But could the `activityEquationAnswerComponent` itself subscribe to results from submissionmanager, see if its own stableid is referenced in the equation results and update itself if necessary? I don't think any other answer/question component cares, right?

alextanlan: Right, other questions don't care.

alextanlan: However, I can't see a decent way to update an equation only. 
Perhaps, just take a reference to a equation child of the composite question and make a force re-rendering?
Will try.

alextanlan: Fixed in that^ way (https://github.com/broadinstitute/ddp-angular/pull/1222/commits/1dd4e07203e464d261e52c842a1e8371a890319a)

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-equation-answer/activityEquationAnswer.component.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    private updateEquationQuestions(equations: AnswerResponseEquation[] = []): void {
+        for (const equation of equations) {
+            // update only single equations (not equations which are children of a composite question)
+            if (this.block.stableId === equation.stableId && equation.values.length === 1) {
+                this.block.setAnswer(equation.values, false);
+            }
+        }
+    }
```
MocanaAtBroad: This looks good! A code change suggestion to make this more compact: move the filtering (iteration through equations to find the one that matches this component) to the pipeline in the ngOnInit.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-composite-answer/activityCompositeAnswer.component.ts
```diff
@@ -42,36 +78,80 @@ export class ActivityCompositeAnswer implements OnChanges {
...
+        // which values were changed (get them from `equationsToUpdate`)
+        // in order to re-render them only
+        this.childAnswerComponents.toArray()
+            .filter((answer: ActivityAnswerComponent) => Object.keys(equationsToUpdate).includes(answer.block.stableId))
+            .map((answer: ActivityAnswerComponent) => answer.equationAnswerComponent)
+            .forEach((equationAnswerComponent: ActivityEquationAnswerComponent, index: number) => {
+                const newValue = [equationsToUpdate[equationAnswerComponent.block.stableId][index]];
+                equationAnswerComponent.block.setAnswer(newValue);
+            });
+    }
```
MocanaAtBroad: Are we sure that for composite we need to do all this? The child equation questions are listening for their changes, so this code and the code in the equation component are both updating? Or you just looking in the equation component for answers of size 1? If so maybe we should add a comment?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-composite-answer/activityCompositeAnswer.component.ts
```diff
@@ -42,36 +78,80 @@ export class ActivityCompositeAnswer implements OnChanges {
...
+                    const newAnswer: AnswerContainer[][] = prevAnswer.map((answerRow: AnswerContainer[], index) =>
+                        // eslint-disable-next-line arrow-body-style
+                        answerRow.map((answer: AnswerContainer) => {
+                            return (answer.stableId === childEquation.stableId) ? {
+                                ...answer,
+                                value: [equation.values[index]]
+                            } : answer;
+                        })
+                    );
+                    this.block.setAnswer(newAnswer, false);
```
MocanaAtBroad: Would be desirable that composite questions don't know about equations either.
This probably can be achieved if each equation.component and/or equation.block knows that it is part of a composite answer AND what its index position it is in.
With this information, by subscribing to the Patch response, the equation component can pick out its own answer ("My stableid is X, my index pos is Y, so this must be my Equation value from the response").

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-equation-answer/activityEquationAnswer.component.ts
alextanlan: Fixed in f7521bc1cc3665755cf7d381755d896d1804bf4d

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-composite-answer/activityCompositeAnswer.component.ts
alextanlan: As I wrote [here](https://github.com/broadinstitute/ddp-angular/pull/1222#discussion_r865199735), I just had no clue how to separate single equations vs composite child equations. I wasn't able to receive a reference to the parent (composite) question. And I was not so brave to create such the reference/connection. That was the root of all the mess around. 

Thank you for such an elegant idea with the index position! 
I fixed the logic in this way. And it looks miles better for now.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-composite-answer/activityCompositeAnswer.component.ts
alextanlan: Removed that, see https://github.com/broadinstitute/ddp-angular/pull/1222#discussion_r866772902

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-equation-answer/activityEquationAnswer.component.ts
```diff
@@ -0,0 +1,57 @@
...
+    get displayValue(): string {
+        return this.block.answer && this.block.answer[0] ? this.formatValue(this.block.answer[0]) : null;
+    }
+
+    private formatValue(answer: DecimalAnswer): string {
+        // align the answer decimal part
+        const fullValue = `${DecimalHelper.mapDecimalAnswerToNumber(answer)}`;
+        const [integerPart, decimalPart] = fullValue.split('.');
+        return integerPart + this.getFormattedDecimalPart(decimalPart);
+    }
```
MocanaAtBroad: We have similar conversion code elsewhere, e.g.: https://github.com/broadinstitute/ddp-angular/blob/eff7a60134e39a99462d239cb648bb8650d58de1/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activityNumericAnswer.component.ts#L123-L137
Let's consolidate into a utility. Or DecimalAnswer interface have an implementation that include toString() and does the same thing.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-composite-answer/activityCompositeAnswer.component.ts
```diff
@@ -110,19 +133,25 @@ export class ActivityCompositeAnswer implements OnChanges {
...
-                childQuestionBlockRow.map((childQuestionBlock) => {
-                    if (childQuestionBlock.validate()) {
-                        return this.buildChildAnswer(childQuestionBlock);
-                    } else {
-                        return this.buildChildAnswer(childQuestionBlock, null);
-                    }
-                }));
+                childQuestionBlockRow
+                    // we don't patch an equation question answer because it is read-only
+                    .filter(childQuestionBlock => !(childQuestionBlock instanceof ActivityEquationQuestionBlock))
```
MocanaAtBroad: I BELIEVE you can safely remove this line as I don't expect the EquationComponent will emit an answer.

alextanlan: Here we iterate through all `childQuestionBlocks` of the composite question independently of which child question answer was changed. The filtering of equations here is to exclude equation answers from the composite answer before patching. 
Otherwise, the patch of the composite answer fails.

MocanaAtBroad: Ah. OK!This something that was not really specified clearly in the API, but what you are doing is reasonable.
You have been able to verify this against the Basil example?

I still would like to get rid of the `instanceof` as it tends to bind the code to specific implementation of QuestionBlock.
The only kind of block that this code should be aware is either a generic ActivityBlock or QuestionBlock.
Would suggest then that we extend ActivityQuestionBlock to include a method like `generatesAnswers(): boolean` or something like that.
The default value can be true, but for EquationBlock it returns false. Then we can get rid of `instanceof`

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-equation-answer/activityEquationAnswer.component.ts
alextanlan: Actually, formatting of a `equation question answer` value and formatting of a `decimal question answer` differ a bit 
(even though they have the same type `DecimalAnswer`). 
In case of we don't have any decimal part scale:
 - `equation` should keep the whole decimal part of the answer value (1.23 -> 1.23)
 - `decimal` should cut the whole decimal part (1.23 -> 1)
 
 Moreover, for `decimal question answer` when the component input fires a value update, we should transform just a number, e.g. 5.27 ( not `DecimalAnswer` format).
 
 Anyway, I have raised a separate [PR#1260](https://github.com/broadinstitute/ddp-angular/pull/1260) for that improvement. It is more generic one. However it did not become easier, imo. 
 
 What  are your thoughts?

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-composite-answer/activityCompositeAnswer.component.ts
alextanlan: 1) yep, I verified the behavior on basil dev env (`Equation activity`).
2) so, you mean that check for a `questionBlock.questionType` does not make sense as well, right ? 
3) `generatesAnswers` means that only EquationQuestion does not change its answer because it is read-only. Is that the idea ?

MocanaAtBroad: Yeah. That's the idea. Essentially instead of checking whether a block is instanceof something we can just call a method that answers whether object can or cannot do something. At this time, EquationQuestionBlock (or whatever it is called) would be the only QuestionBlock that would answer the generatesAnswers question with a false.

alextanlan: 4.
 > get rid of the instanceof as it tends to bind the code to specific implementation of QuestionBlock.

I see your point here. 
However seems to me, in this particular case **maybe** we have to bind the code to `ActivityEquationQuestionBlock` (as a specific implementation of `QuestionBlock`). Because we have some special requirements for that type of questions. 
Like, don't patch the equation answer (both itself and being a composite child). 
Also we should set position indexes in a composite answer (for Equation only), but it doesn't connect to the fact that it can/can't generate answers. It's just because the Equation has quite "unnormal" behavior, we were not able to align it with other questions behavior on back-end. Thus we need to tweak it on the client/Angular side.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-equation-answer/activityEquationAnswer.component.ts
alextanlan: Merged that PR.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-composite-answer/activityCompositeAnswer.component.ts
MocanaAtBroad: Yes. Perfectly fine to subclass. It is desirable in this case.
What we want to avoid is have ActivityCompositeAnswer be aware of QuestionBlock specializations or subclasses, like `ActivityEquationQuestionBlock`. Adding the method `generatesAnswers()` to QuestionBlock allows us to not have to check the specific type of the Block. We just ask it whether it can do a thing.

alextanlan: Fixed in e3738dc34bf88d67c2585152262921740f6ab8ae

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activity-composite-answer/activityCompositeAnswer.component.ts
```diff
@@ -176,4 +211,14 @@ export class ActivityCompositeAnswer implements OnChanges {
     private showAsterisk(questionType: QuestionType): boolean {
         return !this.config.compositeRequiredFieldExceptions.includes(questionType);
     }
+
+    private resetEquationsRowIndexes(compositeChildQuestionBlocks: ActivityQuestionBlock<any>[][]): void {
+        compositeChildQuestionBlocks.forEach((row: ActivityQuestionBlock<any>[], rowIndex: number) => {
+            for (const questionBlock of row) {
+                if (!(questionBlock.generatesAnswers())) {
+                    (questionBlock as ActivityEquationQuestionBlock).compositeRowIndex = rowIndex;
```
MocanaAtBroad: OK. You ALMOST got rid of ActivityEquationQuestionBlock!
Just move `compositeRowIndex` up to ActivityQuestionBlock, set it for all the blocks and you are done.

alextanlan: Done

MocanaAtBroad (approved):

PR #1224, opened 2022-04-26
DDP-7888: abstraction

SimoneMaiwald:


pegahtah (approved):

ddp-workspace/projects/ddp-dsm-ui/src/app/abstraction-group/abstraction-group.model.ts
```diff
@@ -59,15 +60,17 @@ export class AbstractionWrapper {
     public abstraction: AbstractionGroup[],
     public review: AbstractionGroup[],
     public qc: AbstractionGroup[],
-    public finalFields: AbstractionGroup[]
+    public finalFields: AbstractionGroup[],
+    public abstractionActivities: Abstraction[]
   ) {
     this.abstraction = abstraction;
```
GiCharkviani, line 66: I think it's better to write so:
```
class TestClass {
  constructor(public prop: Prop[]) {}
}
```
rather than so:
```
class TestClass {
  constructor(public prop: Prop[]) {
    this.prop = prop
  }
}
```
[TS - Parameter Properties](https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties)

PR #1225, opened 2022-04-27
Added Dependent activities.

leonidrakitin:
Added Dependent activities.

### DIFF
--- a/ddp-workspace/projects/ddp-singular/src/app/components/progress-bar/providers.ts
+++ b/ddp-workspace/projects/ddp-singular/src/app/components/progress-bar/providers.ts
@@ -8,11 +8,31 @@ import { ActivityInstance, ActivityStatusCodes, ConfigurationService, UserActivi
 
 
 const getUpdatedActivityMap = (activities: ActivityInstance[], activeActivityId: string): Map<string, ActivityListItem> => {
-  const isConsentSelf = !!activities.find(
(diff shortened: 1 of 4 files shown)


ddp-workspace/projects/ddp-singular/src/app/components/progress-bar/providers.ts
```diff
@@ -8,11 +8,31 @@ import { ActivityInstance, ActivityStatusCodes, ConfigurationService, UserActivi
 
 
 const getUpdatedActivityMap = (activities: ActivityInstance[], activeActivityId: string): Map<string, ActivityListItem> => {
-  const isConsentSelf = !!activities.find(
-    (activity: ActivityInstance) => activity.activityCode === ActivityCode.ConsentSelf
+  const consentActivityCodes = [ActivityCode.ConsentSelf, ActivityCode.ConsentParental, ActivityCode.ConsentDependent];
+  const consentActivity = activities.find(
+    (activity: ActivityInstance) => consentActivityCodes.includes(activity.activityCode)
   );
```
grigol-brd, line 14: `find()` will return the first one that matches the condition and what if there are multiple consents in an activity?
Is it possible for an activity to have, for example, parental and dependent consent at the same time?

leonidrakitin: Nope, its not possible. There can only be one of these activities at a one time.

grigol-brd (approved):

PR #1227, opened 2022-04-27
DDP-7896/Send-symbolic-date-instead-of-NA

GiCharkviani:


ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.ts
```diff
@@ -135,7 +138,7 @@ export class FieldDatepickerComponent implements OnInit, OnChanges {
 
   public setNA(): void {
     this.isNA = true;
-    this._dateString = 'N/A';
+    this._dateString = '01-01-1000';
```
SimoneMaiwald: I think the string should be `1000-01-01` to be in the same format as the date strings currently are @gmakharat @Anzori-Ghurtchumelia

GiCharkviani: Done. Thanks

SimoneMaiwald (approved):

PR #1292, opened 2022-05-20
fix day, month, year validation rules

LKapanad:


alextanlan:
Just wondering what was the reason for changing the validation rules in the opposite direction ? 
Those rules are used (can be used) by all DSS studies. 
Something can be broken by those radical changes. 
As far as I know, we add REQUIRED validation rule for a DATE question if we should avoid an empty answer. 

At least the changes should be discussed with other teammates.

@MocanaAtBroad, what are your thoughts about it?

LKapanad:
Hi @alextanlan , thanks for noticing. I believe it was a bug these methods are used  on for example FON study to validate that birthday field value is present, but the current code returns true and validates on null (day, month, year) values which seems to me wrong

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/dateValidators/activityDayRequiredDateValidationRule.ts
```diff
@@ -14,10 +14,13 @@ export class ActivityDayRequiredDateValidationRule extends ActivityAbstractValid
...
                 return false;
+            } else {
+                this.result = null;
+                return true;
             }
         }
-        this.result = null;
-        return true;
+        this.result = this.message;
+        return false;
```
MocanaAtBroad, line 23: OK. This code looks wrong to me.
Most of our answer validations (including this one) are setup so that, if there is no answer available yet for a given field, it just return true (the explanation is that the validation in only applicable once we have a value for it). Here we are undoing this behavior.

@LKapanad Do we have a JIRA ticket for a bug that this code is fixing? This code has been in place for a very long time and since we have had many date fields validated, so it would be weird that this would become a problem all of the sudden.

LKapanad: @MocanaAtBroad I think we should close the PR, there was a bug on FON that birthday required validation didn't work correctly because there was no REQUIRED validation on the backend config, and I thought it was a frontend bug. REQUIRED rule was added yesterday and it works now without any changes on frontend.

alextanlan: Cool!

PR #1302, opened 2022-05-24
Singular-Change activity creation to just editing an instance. Disable buttons

MocanaAtBroad:
Please take a look at the attached ticket as it describes specific problem I am trying to address (users have reported others).

In review, please start with this comment as I think this is the biggest problem.
https://github.com/broadinstitute/ddp-angular/pull/1302/files/3f8e1df5440ab0dd6c9a6119ecf629dbf9f5e5dc#r880926312

After that first comment, you can take a look at changes made to disable buttons in order that they appear in File Changed page.

### DIFF
--- a/ddp-workspace/projects/ddp-singular/src/app/components/activities-list/activities-list.component.html
+++ b/ddp-workspace/projects/ddp-singular/src/app/components/activities-list/activities-list.component.html
@@ -56,20 +56,20 @@
           <ng-container *ngTemplateOutlet="hasPreviousInstance(activity) ? continue : start"></ng-container>
 
           <ng-template #start>
-            <button type="button" class="button button--primary button--sm" (click)="onStartClick(activity)">
+            <button type="button" class="button button--primary button--sm" (click)="onStartClick(activity)" [disabled]="disableButtons">
               {{ (isMedicalRecordFileUpload(activity) ? 'SDK.UploadButton' : 'SDK.StartButton') | translate }}
             </button>
           </ng-template>
 
           <ng-template #continue>
-            <button type="button" class="button button--primary button--sm" (click)="onContinueClick(activity)">
+            <button type="button" class="button button--primary button--sm" (click)="onContinueClick(activity)"  [disabled]="disableButtons">
               {{ (isMedicalRecordFileUpload(activity) ? 'SDK.UploadButton' : 'SDK.ContinueButton') | translate }}
             </button>
           </ng-template>
         </ng-container>
 
         <ng-container *ngSwitchCase="ActivityStatusCodes.IN_PROGRESS">
-          <button type="button" class="button button--primary button--sm" (click)="onContinueClick(activity)">
+          <button type="button" class="button button--primary button--sm" (click)="onContinueClick(activity)"  [disabled]="disableButtons">
             {{ (isMedicalRecordFileUpload(activity) ? 'SDK.UploadButton' : 'SDK.ContinueButton') | translate }}
           </button>
         </ng-container>
@@ -78,13 +78,13 @@
           <ng-container *ngTemplateOutlet="isActivityEditable(activity) ? update : view"></ng-container>
 
           <ng-template #update>
-            <button type="button" class="button button--outline button--sm fixed-button" (click)="onEditClick(activity)">
+            <button type="button" class="button button--outline button--sm fixed-button" (click)="onEditClick(activity)"  [disabled]="disableButtons">
               {{ (isMedicalRecordFileUpload(activity) ? 'SDK.UploadButton' : 'SDK.EditButton') | translate }}
             </button>
           </ng-template>
 
           <ng-template #view>
-            <button type="button" class="button button--outline button--sm fixed-button" (click)="onViewClick(activity)">
+            <button type="button" class="button button--outline button--sm fixed-button" (click)="onViewClick(activity)"  [disabled]="disableButtons">
               {{ 'SDK.ReviewButton' | translate }}
             </button>
           </ng-template>
--- a/ddp-workspace/projects/ddp-singular/src/app/components/activities-list/activities-list.component.ts
+++ b/ddp-workspace/projects/ddp-singular/src/app/components/activities-list/activities-list.component.ts
@@ -11,6 +11,7 @@ import { ActivityIcons } from '../../constants/activity-icons';
 })
 export class ActivitiesListComponent {
   @Input() activities: ActivityInstance[];
+  @Input() disableButtons = false;
   @Output() startActivity = new EventEmitter<ActivityInstance>();
   @Output() continueActivity = new EventEmitter<ActivityInstance>();
   @Output() editActivity = new EventEmitter<ActivityInstance>();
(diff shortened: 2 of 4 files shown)


ddp-workspace/projects/ddp-singular/src/app/components/pages/participant-list/participant-list.component.ts
```diff
@@ -131,21 +132,19 @@ export class ParticipantsListComponent implements OnInit {
...
-      .createInstance(this.config.studyGuid, activityToEdit.activityCode)
-      .pipe(take(1))
-      .subscribe(activity => {
-        this.setCurrentActivity(activity as ActivityInstance);
-        this.redirectToSurvey(activity.instanceGuid);
-      });
+    this.setCurrentActivity(activityToEdit, false);
+    this.redirectToSurvey(activityToEdit.instanceGuid);
+    this.isPageBusy = false;
   }
```
MocanaAtBroad, line 141: Creating activity instance here makes no sense to me. As you see, the method already receives an instance for editing, so why do we  need to create one?
Maybe there is a hidden reason? Take a look at JIRA ticket https://broadinstitute.atlassian.net/browse/DDP-8060.
My change was to get rid of the createInstance and just redirect to a web page where it can be edited.

ddp-workspace/projects/ddp-singular/src/app/components/activities-list/activities-list.component.html
```diff
@@ -78,13 +78,13 @@
...
+            <button type="button" class="button button--outline button--sm fixed-button" (click)="onEditClick(activity)"  [disabled]="disableButtons">
               {{ (isMedicalRecordFileUpload(activity) ? 'SDK.UploadButton' : 'SDK.EditButton') | translate }}
             </button>
           </ng-template>
 
           <ng-template #view>
-            <button type="button" class="button button--outline button--sm fixed-button" (click)="onViewClick(activity)">
+            <button type="button" class="button button--outline button--sm fixed-button" (click)="onViewClick(activity)"  [disabled]="disableButtons">
               {{ 'SDK.ReviewButton' | translate }}
             </button>
```
MocanaAtBroad, line 89: To be triple sure that we are not allowing double clicks or duplicate requests, I am disabling buttons whenever something is going on.

ddp-workspace/projects/ddp-singular/src/app/components/activities-list/activities-list.component.ts
```diff
@@ -11,6 +11,7 @@ import { ActivityIcons } from '../../constants/activity-icons';
 })
 export class ActivitiesListComponent {
   @Input() activities: ActivityInstance[];
+  @Input() disableButtons = false;
```
MocanaAtBroad, line 14: Since the "action" (e.g. getting an activity) happens in the parent component, we use @Input so parent can tell this child "disable buttons"

ddp-workspace/projects/ddp-singular/src/app/components/pages/participant-list/participant-list.component.html
```diff
@@ -104,6 +104,7 @@ <h1 class="title">
         <div class="participant__content" *ngIf="isParticipantContentExpanded(participant)">
           <app-activities-list
             [activities]="participant.activities"
+            [disableButtons]="isPageBusy || isEnrollBtnDisabled || loading"
```
MocanaAtBroad, line 107: And this is how set that disableButtons input param. Note that there were already some other flags used to disable stuff. I am using those AND I am adding a new one: isPageBusy. 
(We could probably refactor so we have only one flag or something like that, but this is a bug fix, so trying to take the safest path).

ddp-workspace/projects/ddp-singular/src/app/components/pages/participant-list/participant-list.component.ts
```diff
@@ -43,6 +43,7 @@ export class ParticipantsListComponent implements OnInit {
   participants: Participant[] = [];
   private expandedMap: Record<string, boolean> = {};
   errorMessage: string | null = null;
+  isPageBusy = false;
```
MocanaAtBroad, line 46: And here is that new flag.

ddp-workspace/projects/ddp-singular/src/app/components/pages/participant-list/participant-list.component.ts
```diff
@@ -131,21 +132,19 @@ export class ParticipantsListComponent implements OnInit {
...
-    this.activityService
-      .createInstance(this.config.studyGuid, activityToEdit.activityCode)
-      .pipe(take(1))
-      .subscribe(activity => {
-        this.setCurrentActivity(activity as ActivityInstance);
-        this.redirectToSurvey(activity.instanceGuid);
-      });
+    this.setCurrentActivity(activityToEdit, false);
+    this.redirectToSurvey(activityToEdit.instanceGuid);
+    this.isPageBusy = false;
```
MocanaAtBroad, line 139: Whenever something happens I am saying the page is busy (I think it might be overkill here, but at worse, it does not hurt anything).

ddp-workspace/projects/ddp-singular/src/app/components/pages/participant-list/participant-list.component.ts
```diff
@@ -131,21 +132,19 @@ export class ParticipantsListComponent implements OnInit {
...
   }
 
   onViewActivity(participantGuid: string, activity: ActivityInstance): void {
+    this.isPageBusy = true;
     this.setParticipant(participantGuid);
     this.setCurrentActivity(activity, true);
     this.redirectToSurvey(activity.instanceGuid);
+    this.isPageBusy = false;
```
MocanaAtBroad, line 147: The other place that I set page busy.

ddp-workspace/projects/ddp-singular/src/app/components/pages/participant-list/participant-list.component.ts
grigol-brd: I can't speak for this study, but I've seen this behavior in another study too.
My idea would be that once an activity is submitted/completed it becomes read-only? So activity edit functionality is done using kind of a workaround, that is - creating a new activity and populating fields from the old one, this way user can change whatever fields he wants in this new pre-filled activity. But this is more of a guess, I don't know why or how the whole flow goes

MocanaAtBroad: Hi Grigol. Yes. We have had that kind of interaction where a new activity is created, copying existing fields. And that seems to be what is happening. Current implementation, though, really causes a lot of problems.
On the other hand, I am not seeing that the old activity is reachable or viewable by users in any way (maybe from DSM?). Will ask Jeff and Whitney if create/copy/edit is what is needed here.

grigol-brd: I was suspicious that removing that new instance creation functionality might break activity edit functionality but looks like it doesn't. Edit still works just fine, it saves changed fields and I can see them changed, so I'm not sure why they had additional instance creation in the first place, maybe it's a remnant of a copy-pasted code? Who knows, anyway I think your changes are legit

grigol-brd (approved):
Disabling buttons makes sense and also edit functionality still works as before, without redundant instance creation, so this LGTM

ddp-workspace/projects/ddp-singular/src/app/components/pages/participant-list/participant-list.component.ts
MocanaAtBroad: @grigol-brd I am being told by Whitney that the creation of new activities that copy old data is not a real requirement. So I am going to go ahead and merge.

Thanks for review!

PR #1340, opened 2022-06-03
Prevent Angular from sending page_view Gtag events. Gtag does that au…

MocanaAtBroad:
Tried to explain problem in https://broadinstitute.atlassian.net/browse/DDP-8139
See comments with files for proposed fix explanation.

I have deployed this branch to dev environment and here are a couple of screenshots showing what happens when we now navigate to the /for-researchers page:
From browser, there is only one event (compare to screenshot in ticket)
Screen Shot 2022-06-03 at 2 27 37 PM
And only one page location event with "for researchers" in Google Analytics console:
Screen Shot 2022-06-03 at 2 27 14 PM

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/services/analyticsEvents.service.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/services/analyticsEvents.service.ts
@@ -1,78 +1,87 @@
-import { Injectable } from '@angular/core';
-import { Subject, Observable } from 'rxjs';
-import { AnalyticsEvent } from '../models/analyticsEvent';
-import { GTagEvent } from '../models/gtagEvent';
-import { filter } from 'rxjs/operators';
-import { GoogleAnalyticsEvent, isAnalyticsEvent, isGtagEvent } from '../models/googleAnalyticsEvent';
+import {Injectable} from '@angular/core';
+import {Subject, Observable} from 'rxjs';
+import {AnalyticsEvent} from '../models/analyticsEvent';
+import {GTagEvent} from '../models/gtagEvent';
+import {filter} from 'rxjs/operators';
+import {GoogleAnalyticsEvent, isAnalyticsEvent, isGtagEvent} from '../models/googleAnalyticsEvent';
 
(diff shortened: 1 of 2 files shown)


ddp-workspace/projects/ddp-singular/src/app/app.module.ts
```diff
@@ -90,7 +90,7 @@ declare const gtag: (...args: any[]) => void;
 export class AppModule {
     constructor(private analytics: AnalyticsEventsService) {
         // https://developers.google.com/tag-platform/gtagjs/reference#event
-        this.analytics.gTagEvents
+        this.analytics.gTagEventsExcludingPageViews
             .subscribe(event => gtag('event', event.event_name, event.parameters));
```
MocanaAtBroad, line 94: And here we use the filtered observable.

ddp-workspace/projects/ddp-sdk/src/lib/services/analyticsEvents.service.ts
```diff
@@ -1,78 +1,87 @@
...
-    return `${location.protocol}//${location.hostname}${location.port ? `:${location.port}` : ``}${location.pathname}`;
-  }
+    /* There does not appear to be a way to prevent gtag.js from automatically sending page view events
+    despite what their documentation says here:
+    https://developers.google.com/analytics/devguides/collection/gtagjs#disable_pageview_measurement
+    Use this option to avoid sending duplicate page_view events given gtag is going to send its own no matter
+    what */
+    public get gTagEventsExcludingPageViews(): Observable<GTagEvent> {
+        return this.gTagEvents.pipe(filter(event => event.event_name !== 'page_view'));
+    }
```
MocanaAtBroad, line 69: Please read comment above. Tried as much as I could and could not stop gtag.js from sending `page_view` on its own. Saw some comments in stackoverflow that folks developing for react faced the same thing.
So we are just going to filter those out. For Singular purposes the generic gtag.js event should be acceptable.

bskinner (approved):

PR #1392, opened 2022-06-15
add modular folders

josemanuel167:
Add modular folders and make use of mat-drawer-container


josemanuel167:
Activities component can't be moved because depends on DdpModule, if I try to load again DdpModule it throws this error:
image

MocanaAtBroad:
Without a ticket or description, I really cannot tell what you are trying to achieve.  Please explain what your goal is with this PR (creating a ticket or pointing to one would be good). And then in PR, try to add comments that walk the reviewer through your approach to solving the problem.

josemanuel167:
Main goal for this is achieve a feature structure following best practices:
-- core
    -> module
-- feature
   feature-module
-- feature2..
   feature2-module
-- shared
   shared-module

In core module now you can find the main layout that is going to frame all feature pages.

Fon component now implements a responsive layout making use of `<mat-drawer-container>, <mat-drawer-content>, and <mat-drawer>.`

Folders structure now looks like this:

image

ddp-workspace/projects/ddp-dsm-ui/src/app/FON/core/fon-routing.module.ts
```diff
@@ -0,0 +1,33 @@
...
+      { path: '', redirectTo: 'home' },
+      { path: 'home', component: HomeComponent },
+      {
+        path: 'patients',
+        loadChildren: () =>
+          import('./../patients/patients.module').then((m) => m.PatientsModule),
+      },
+      // { path: 'patients', component: ParticipantsListComponent },
+      // { path: 'participantList', component: ParticipantsListComponent },
+      // {
```
GiCharkviani, line 20: I suggest not to leave a commented-out code.

GiCharkviani (changes requested):
Jose, please before changing architecture (in this case folder structure and extra lazy-loaded module inside the same type of module), let's discuss because there are lots of things going on.
I also saw that you are having a problem with DdpModule and that was expected because you have duplicated its declaration since you've made the child lazy module and have imported DdpModule in both, parent and child modules (even though they are both lazy-loaded).

P.S. Let's for now just add new features, without changing any structure. Definitely, further improvements will take place later on, but for now, we've got priorities that we should be concentrated on.
P.P.S. Anyway, Thank you for your effort!

@alextanlan @MocanaAtBroad What are your thoughts?

alextanlan:
I agree that because of the fact that we are running out of time we should not change any not related to demo stuff, for instance the FON architecture. 
(see [current goals comment in Slack](https://broadinstitute.slack.com/archives/C02FZQST19S/p1655314783672549)).
The PR changes are out of demo preparation scope, as I can see.
If we really need we can apply it later. And definitely any global changes/refactoring of the code should be discussed together.

alextanlan:
For now, I'd suggest to postpone this PR.

PR #1425, opened 2022-06-21
fixed session key setting problem

GiCharkviani:


ddp-workspace/projects/ddp-dsm-ui/src/app/FON/fon.module.ts
```diff
@@ -82,31 +76,5 @@ const layoutComponents = [NavigationComponent];
...
-
-    // Store
-    this.storeService.setStudy = MainConstants.study;
-    this.storeService.dispatchGetSettings(MainConstants.participantsListParent);
-
-    // Translation
-    this.translateService.setDefaultLang(LOCALE);
-    this.translateService.use(LOCALE);
-  }
+  constructor() {}
```
alextanlan, line 79: We don't  need it more, right?

GiCharkviani: Not here, but in fon.component, as that is initializing always when you log in to FON.

alextanlan (approved):

PR #1462, opened 2022-06-28
DDP-8200: fix validation for Mailing address

gigijanelidze26:
Validations in emailing section were not correct. Basically it was enough just to fill one of the fields and then it would let you to submit, which is incorrect and that's what I fixed in the pr. See related tickets below. 

https://broadinstitute.atlassian.net/jira/software/c/projects/DDP/boards/836?modal=detail&selectedIssue=DDP-8165&quickFilter=2669,

https://broadinstitute.atlassian.net/jira/software/c/projects/DDP/boards/836?modal=detail&selectedIssue=DDP-8200&quickFilter=2669

basically I refactored `CountOfFieldWithData` and changed it with `validateRequiredFields` function where I am checking if all required fields are filled with data. Before it was first checking all fields and then in `enoughDataToSave` it was checking if at least one of them was filled with data (why would we check all of the fields if then we are relying on only one field to validate the form), which was breaking the lms and osteo studies.


GiCharkviani (changes requested):
Please run tests

alextanlan:
Can you please describe the issue that you have been fixing or provide a link to the related ticket?

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -708,7 +708,7 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
         if (this.block.requireVerified && !currentAddress) {
             return false;
         }
-        return !(this.block.requirePhone && currentAddress && !(currentAddress.phone));
+        return this.block.requirePhone ? !currentAddress.phone : this.countOfFieldsWithData(currentAddress);
```
alextanlan: 1. What if currentAddress is null here?

alextanlan: 2. Previously we checked if a phone is required and if the phone exists. 
You added a check if all (most of) address fields are not empty. 
What was the reason for the change?

alextanlan: 3. 
` return this.block.requirePhone ? !currentAddress.phone : ...`
If `this.block.requirePhone===true` and `currentAddress.phone === null`
(means like a block requires a phone number but there is no phone) 
the method `meetsActivityRequirements` returns true. 
Is that behavior correct?

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -727,17 +727,19 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
...
     enoughDataToSave(address: Address | null): boolean {
-        return address && this.countOfFieldsWithData(address) >= 1;
+        return address && this.countOfFieldsWithData(address);
     }
 
-    countOfFieldsWithData(address: Address): number {
-        const isNonBlankString: FuncType<boolean> = (val) => util.isString(val) && val.trim().length > 0;
-        const propsToCheck: (keyof Address)[] = ['name', 'country', 'street1', 'street2', 'state', 'city', 'zip', 'phone'];
-        return propsToCheck.map(prop => address[prop]).filter((value) => isNonBlankString(value)).length;
+    countOfFieldsWithData(address: Address): boolean {
```
alextanlan: Name of the method does not fit the returned type.
`countOf..  => boolean`
It is a bit confusing.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -727,17 +727,19 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
...
+        return Object.entries(address).every(([key, value]) =>
+            !['street2', 'phone', 'guid'].includes(key)
```
alextanlan: Previously we checked a defined list of address fields and whether all of them are not empty.

1. Why do you exclude `street2`  and   `phone` from that ^ check? It is not obvious for me.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -727,17 +727,19 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
...
+                ? value.toString().trim().length > 0
```
alextanlan: What if someone adds a new Address class property and it equals null/undefined or even an empty object?
The check will fail.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
alextanlan: 2. The check is supposed to go through all the address properties (except `'street2', 'phone', 'guid'`).
There is also boolean field 'isDefault' there. 
It passes the check (`value.toString().trim().length > 0`) in any case 
But why should we check it at all?

MocanaAtBroad (changes requested):
Please provide explanation of what you are trying to do. Read the bug report and I do not understand what you are trying to do.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
gigijanelidze26: these problems are already caught and are being fixed, will update pr soon, thanks

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
gigijanelidze26: Street2 and phone is not required fields that's why we are excluding it, but adding logic so that if phone is required it's not excluded, will update pr soon.
about isDefault property - that becomes string 'true' and so passes the test but sure we can exlude it as well.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
gigijanelidze26: that's already fixed, will be included in update of the pr.

alextanlan:
Can you please take a look at my alternative bug-fix [PR#1502](https://github.com/broadinstitute/ddp-angular/pull/1502)?
It does not have any potentially "breaking" changes. And I believe that PR fixes the described issue.

Also I checked the fix on LMS study. Seems working.
Added a screen capture there.

MocanaAtBroad:
We have merged @alextanlan proposed fix https://github.com/broadinstitute/ddp-angular/pull/1502
@gigijanelidze26 I am going to close this PR for now. If you or QA finds that there is still a problem, reopen and we can talk about it next week (@alextanlan is out on vacation now).

PR #1469, opened 2022-06-29
Fixed logic of Day, Month, Year required rules.

leonidrakitin:
Fixed logic of Day, Month, Year required rules.
This fixes is needed because if user doesn't pick whatever day/month/year (answer the date question with any of upper mentioned rules), then user can simply skip this question. Required rule is not approach for situations like user must to answer the date question (it should required), but also he should be able to specify only year.

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/dateValidators/activityDayRequiredDateValidationRule.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/dateValidators/activityDayRequiredDateValidationRule.ts
@@ -11,7 +11,7 @@ export class ActivityDayRequiredDateValidationRule extends ActivityAbstractValid
     public recalculate(): boolean {
         if (this.question.answer != null) {
             const value = this.question.answer;
-            if (!this.isBlank(value) && value.day === null) {
+            if (value.day === null) {
                 this.result = this.message;
                 return false;
             }
--- a/ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/dateValidators/activityMonthRequiredDateValidationRule.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/services/activity/validators/dateValidators/activityMonthRequiredDateValidationRule.ts
@@ -11,7 +11,7 @@ export class ActivityMonthRequiredDateValidationRule extends ActivityAbstractVal
     public recalculate(): boolean {
         if (this.question.answer != null) {
             const value = this.question.answer;
-            if (!this.isBlank(value) && value.month === null) {
+            if (value.month === null) {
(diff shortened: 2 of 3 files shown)


GiCharkviani (changes requested):
Please run tests

alextanlan:
Firstly, what is the reason for changing the logic that has been used for a while in different studies?
Is there a business requirement/related ticket?

leonidrakitin:
@alextanlan 
> Firstly, what is the reason for changing the logic which has been used for a while in different studies?
> Is there a business requirement/related ticket?

So nope, but it looks like an obvious bug.

alextanlan:
> it looks like an obvious bug.

It does not, at least for me.
The logic that you are trying to get rid of was added 2 years ago. 
After that it has been used on production in several studies. 
How it is possible to change the behavior without any requirements?

If it is a bug (that we probably were not able to catch during a couple of years), I suggest documenting it. 
Like create a related ticket with described flow to reproduce, current behavior and desired behavior.
Is anyone from QA aware of the bug? Have you discussed it with someone? 

In case it is still a bug that should be fixed - the changed logic should be covered by unit-tests.

@MocanaAtBroad, any thoughts about it?

leonidrakitin:
> > it looks like an obvious bug.
> 
> It does not, at least for me. The logic that you are trying to get rid of was added 2 years ago. After that it has been used on production in several studies. How it is possible to change the behavior without any requirements?
> 
> If it is a bug (that we probably were not able to catch during a couple of years), I suggest documenting it. Like create a related ticket with described flow to reproduce, current behavior and desired behavior. Is anyone from QA aware of the bug? Have you discussed it with someone?
> 
> In case it is still a bug that should be fixed - the changed logic should be covered by unit-tests.
> 
> @MocanaAtBroad, any thoughts about it?

@alextanlan @MocanaAtBroad 
You might be right, maybe it works as it should be, so in this case we need to implement new rule, when date question is required, but user should be able select only year, or month and year. By the way, it uses only in testboston study. 

About QA  - yep, we have several tasks about date question error logic.

MocanaAtBroad:
If this is a problem that needs fixing, please make sure there is a ticket for it. (As @alextanlan mentioned, this functionality has existed for a long time so unless something has changed, would be surprised there are problems here).

leonidrakitin:
> If this is a problem that needs fixing, please make sure there is a ticket for it. (As @alextanlan mentioned, this functionality has existed for a long time so unless something has changed, would be surprised there are problems here).

@MocanaAtBroad, https://broadinstitute.atlassian.net/browse/DDP-7493?focusedCommentId=1317845 (page 4, point 4)

MocanaAtBroad:
Discussed in more detail with @leonidrakitin and we now have better understanding of problem. Closing this PR and will be creating new tickets.

PR #1482, opened 2022-07-05
DDP/7520 - DDP/7501 tooltip and dropdown year menu fix

llobzhanidze:
There is a changes in SDK. one change is about tooltip, because there was not any icon declaration which displayed this tooltip icon and this is the reason why I made changes in SDK, second one is about dropdown year menu change, I had changed a "startYear" property and "YEARS_BACK" property because in the backend,  year calculation started from 1898 and our QA sad that it should be 1913 and also in front side dropdown menu showing last year only 1922 so this solution fix this problem for all studies.

### DIFF
--- a/ddp-workspace/projects/ddp-fon/src/index.html
+++ b/ddp-workspace/projects/ddp-fon/src/index.html
@@ -17,7 +17,9 @@
     <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet" />
 
     <script async src="https://www.googletagmanager.com/gtag/js?id=G-95RZ7RFCYH"></script>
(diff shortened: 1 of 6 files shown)


ddp-workspace/projects/ddp-fon/src/index.html
```diff
@@ -17,7 +17,9 @@
     <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet" />
 
     <script async src="https://www.googletagmanager.com/gtag/js?id=G-95RZ7RFCYH"></script>
+      <script src="https://kit.fontawesome.com/9096d3094f.js" crossorigin="anonymous"></script>
```
GiCharkviani, line 20: **#Optional:**
I'm just wondering, do you still need this script import?

PR #1488, opened 2022-07-06
handle undefined language

gigijanelidze26:
Handle undefined currentlanguage with '?' mark. problem was caused because - language is set asynchronously and in function it was undefined and was not handled correctly.


alextanlan (approved):

MocanaAtBroad:
Please make sure merge commit message references ticket id. We want to be able to trace back every commit to a ticket.

PR #1496, opened 2022-07-07
DDP/7472 - DDP/8283 tooltip and mail compare menu fix

llobzhanidze:
I added tooltip from configuration Service files only for FON project, second changes is in the SDK and its about mailing comparison, when you type your mail in confirmation field its getting error that 'emails does not match' and I fix this problem with adding a lowercase indicator and this comparing emails with lowercase, it doesn't matter if its typed lowercase or uppercase this bug was in the ticket and I check all other studies and its not damaging anything. you can check all this tickets in down bellow.

https://broadinstitute.atlassian.net/browse/DDP-7472 - FON : Patient profile bugs
https://broadinstitute.atlassian.net/browse/DDP-8283 - [DSS] FON_Tooltips


alextanlan:
> one change is about tooltip icon, there was no any icon declaration in sdk and i was add it, this was a global problem and this 
> solution will fix this problem for all study. 

Actually, it was not a problem.
`ddp-sdk\src\lib\components\tooltip.component.ts` component contains a default value for the icon path:
```
export class TooltipComponent {
    @Input() icon = this.config.tooltipIconUrl;
```
It allows to customize that icon image for every study. 
(e.g. see `ddp-lms\src\app\app.module.ts`, line 75 - `sdkConfig.tooltipIconUrl = 'assets/images/info.png';`)

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activityEmailInput.component.ts
```diff
@@ -160,7 +160,7 @@ export class ActivityEmailInput implements OnInit, OnChanges, OnDestroy {
                 return null;
             }
             // Set error on matchingControl if validation fails
-            if (control.value !== matchingControl.value) {
+            if (control.value.toLowerCase !== matchingControl.value.toLowerCase) {
```
alextanlan: Even though email address is not a case sensitive in general, it seems (to me) that it is more clear for a user to have the same values for `email` and `confirmEmail` fields. 
I have not found any requirements to change the behavior in related tickets DDP-7520 / DDP-7501.
Was there a demand to do it ?

ddp-workspace/projects/ddp-sdk/src/lib/services/configuration.service.ts
```diff
@@ -124,5 +124,5 @@ export class ConfigurationService {
      * Configure additional fields to show for institutions
      */
     institutionsAdditionalFields: Partial<Record<InstitutionType, InstitutionAdditionalFields[]>> = {};
-    detail_MaxLength: number = 500;
+    detail_MaxLength = 500;
```
MocanaAtBroad, line 127: We do camel case, so please change to something like pickListDetailMaxLength and add comment explaining what the variable means (I see it was there before, but by reading name only I had no idea what it was until I browsed code to see how it was being used).

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activityEmailInput.component.ts
```diff
@@ -160,9 +160,12 @@ export class ActivityEmailInput implements OnInit, OnChanges, OnDestroy {
                 return null;
             }
             // Set error on matchingControl if validation fails
-            if (control.value !== matchingControl.value) {
+            const emailequal = control.value;
+            const confemailequal = matchingControl.value;
+            if (emailequal?.toLowerCase() !== confemailequal?.toLowerCase()) {
```
MocanaAtBroad, line 165: I would get rid of this check. Does not buy us anything. Why should we allow a user to type things in upper and lowercase when what we want is for them to be careful entering email address?
(I understand email addresses are case-insensitive but I think this does not help the objective of getting a valid email).

llobzhanidze: when you type your mail in confirmation field its getting error that 'emails does not match' and I fix this problem with adding a lowercase indicator and this comparing emails with lowercase it doesn't matter if its typed lowercase or uppercase this bug was in the ticket and I check all other studies and its not damaging anything.

PR #1497, opened 2022-07-07
Remove husky library and its git pre-push hook

alextanlan:
We don't need `ng lint` git pre-push hook more as we have a pipeline script to run tests and linter check on PR's.


GiCharkviani (approved):
Thanks a lot for removing the actual source of the problem.

PR #1534, opened 2022-07-15
DDP-7351-scroll-up-error

gigijanelidze26:
Related ticket -> https://broadinstitute.atlassian.net/jira/software/c/projects/DDP/boards/836?modal=detail&selectedIssue=DDP-8341&quickFilter=2669

I think issue here is that in some cases validationRequest property was not updated in the if statemant, because of that changeDetection was not activating validationRequested setter in activityQuestion component, where validationRequested$ 
observable emits the new value. scrolling is only triggered when a new value is received. I moved validationRequest property out of the if statement so that it's always updated.


MocanaAtBroad (approved):
Looks like fixes the problem! 👍

PR #1544, opened 2022-07-19
[DSM] DDP-8392 fixed-fields-saving-issue

MishaK1204:
Update value for camel cased column in additionalValuesJson

https://broadinstitute.atlassian.net/browse/DDP-8392


ddp-workspace/projects/ddp-dsm-ui/src/app/medical-record/medical-record.component.ts
```diff
@@ -399,6 +399,8 @@ export class MedicalRecordComponent implements OnInit {
     }
     if (v !== null) {
       if (this.medicalRecord.additionalValuesJson != null) {
+        const camelCaseColumnName = Utils.convertUnderScoresToCamelCase(colName);
+        this.medicalRecord.additionalValuesJson[ camelCaseColumnName ] = v;
         this.medicalRecord.additionalValuesJson[ colName ] = v;
```
SimoneMaiwald, line 442: is that line still needed with the new added one

MishaK1204: Yes, this line is still needed because, if colName is snake cased, backend requires snake cased parameter to update value, otherwise it throws error.

SimoneMaiwald: will that mean we are sending values in 2 different writing cases to the backend? I wonder if that makes new problems if we have a value double in camelCase and SNAKE_CASE

MishaK1204: Yes, we are sending values in 2 different cases, more exactly we were sending 2 different writing cases before I made this change. What I have done is that I updated camelCase (which was already there with snake_case value) value in the object, we are sending to backend. I have tested these changes and there was no problem. Also I have discussed this issue with Nika and we agreed that it is the proper solution for this issue for now.

MishaK1204: Yes, we are sending values in 2 different cases, more exactly we were sending 2 different writing cases even before I made this change, because value of colName we were getting was in snake_case but in this.participant.participant.additionalValuesJson initial value was in camelCase. So what I have done is that I updated camelCase value (which was already there with snake_case value) in the object, we are sending to backend. I have tested these changes and there was no problem. Also I have discussed this issue with Nika and we agreed that it is the proper solution for this issue for now.

SimoneMaiwald (approved):

PR #1572, opened 2022-07-22
DDP-7581 - remove dashboard export

devonbush:
This removes the broken export feature from the bottom of the dashboard -- all exports should be done via the particpant list.  As best as I can tell this feature did not use any backend APIs (part of the reason why it was broken), and so there is no corresponding backend change needed.

BEFORE:
image

AFTER:
image

TO TEST:
1. Load any study's dashboard
2. confirm that the "export" button does not appear


devonbush:
included wrong commits by mistake, will reopen as different branch

PR #1573, opened 2022-07-22
[DSM] DDP-7581: removing export feature from dashboard

devonbush:
This removes the broken export feature from the bottom of the dashboard -- all exports should be done via the particpant list. As best as I can tell this feature did not use any backend APIs (part of the reason why it was broken), and so there is no corresponding backend change needed.

BEFORE:
image

AFTER:
image

TO TEST:

Load any study's dashboard
confirm that the "export" button does not appear


devonbush:
I accidentally merged this direct to the RC branch -- so I will make sure to address any comments here in the RC as well

SimoneMaiwald (approved):

SimoneMaiwald:
can we also remove that from the RC

PR #1589, opened 2022-07-26
[DSM] DDP-8449 - fixed blank email validation issue

MishaK1204:
Initial email value, if an email was not set was undefined. Instead of undefined we should return an empty string so that the updated regex pattern can match the empty string and the email will be valid.

https://broadinstitute.atlassian.net/browse/DDP-8449?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel

### DIFF
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-page/participant-page.component.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-page/participant-page.component.ts
@@ -173,7 +173,7 @@ export class ParticipantPageComponent implements OnInit, OnDestroy, AfterViewChe
       window.scrollTo( 0, 0 );
       this.scrolled = true;
     }
-    this.validateEmailInput(this.participant.data.profile['email'] ? this.participant.data.profile['email'] : '');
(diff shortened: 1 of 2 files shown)


SimoneMaiwald (dismissed):

ddp-workspace/projects/ddp-dsm-ui/src/app/participant-page/participant-page.component.ts
```diff
@@ -173,7 +173,7 @@ export class ParticipantPageComponent implements OnInit, OnDestroy, AfterViewChe
       window.scrollTo( 0, 0 );
       this.scrolled = true;
     }
-    this.validateEmailInput(this.participant.data.profile['email']);
+    this.validateEmailInput(this.participant.data.profile['email'] ? this.participant.data.profile['email'] : '');
```
alextanlan: `this.participant.data.profile['email'] ? this.participant.data.profile['email'] : ''` 
can be shortened like
`this.participant.data.profile['email'] || ''`

alextanlan (approved):

PR #1602, opened 2022-07-28
[SINGULAR]DDP-8482: Input fields for feet and inches for PATIENT_SURVEY and question stable id WHAT_IS_YOUR_HEIGHT_FEET is not preventing negative values

juanseba21:
Preventing negative values for Numeric Questions.

Now it's taking the min and max values from the validations entry:

image

image

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activityNumericAnswer.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activityNumericAnswer.component.ts
@@ -23,8 +23,8 @@ import { LayoutType } from '../../../models/layout/layoutType';
                type="number"
                [appInputRestriction]="isIntegerQuestion ? InputRestriction.Integer : ''"
                [formControl]="numericField"
-               [min]="block.min"
-               [max]="block.max"
+               [min]="minValue"
+               [max]="maxValue"
                autocomplete="off"
                [step]="valueChangeStep"
                [placeholder]="placeholder || block.placeholder"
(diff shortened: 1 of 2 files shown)


alextanlan:
In this particular example on basil dev (that you provided) we miss only min value for the input. 
d12

Seems like the value 0 (for min here) is handled incorrect.

Moreover, we have a separate mapper for transformation a DTO data to a front-end model  
- see `projects\ddp-sdk\src\lib\services\activity\activityQuestionConverter.service.ts`

alextanlan:
The more interesting stuff that for the case described in the [DDP-8482 ticket ](https://broadinstitute.atlassian.net/browse/DDP-8482) (the numeric question is a child of a composite question)   we don't have neither max not min values. 
image

However we have got them in back-end response:
U1
 
 
 Seems like we lost the values in `activityQuestionConverter.service.ts`.
 Please take a look at `getCompositeBlock`, `getNumericBlock` methods.

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/answers/activityNumericAnswer.component.ts
```diff
@@ -46,10 +46,20 @@ export class ActivityNumericAnswer implements OnInit, OnChanges, OnDestroy {
     public numericField: FormControl;
     InputRestriction = InputRestriction;
     private subs: Subscription;
+    public minValue: Number;
+    public maxValue: Number;
 
     public ngOnInit(): void {
         this.initForm();
-
+        const validatorItem=this.block.validators.find((item) => !!item);
```
alextanlan: We have a separate mapper for transformation a DTO data to a front-end model (see https://github.com/broadinstitute/ddp-angular/pull/1602#issuecomment-1199200283).
No need for updating the component.

juanseba21: After taking a look at  projects\ddp-sdk\src\lib\services\activity\activityQuestionConverter.service.ts I adjusted part of the logic related to the values for min and max.
Now we can see the input is taking the values properly:
image

ddp-workspace/projects/ddp-sdk/src/lib/services/activity/activityQuestionConverter.service.ts
```diff
@@ -278,8 +278,8 @@ export class ActivityQuestionConverter {
         numericBlock.placeholder = questionJson.placeholderText;
         const intRangeValidation = questionJson.validations.find(validation => validation.rule === ValidationRuleType.IntRange);
         if (intRangeValidation) {
-            numericBlock.min = intRangeValidation.min || null;
-            numericBlock.max = intRangeValidation.max || null;
+            numericBlock.min = isNaN(intRangeValidation.min) ? null : intRangeValidation.min;
+            numericBlock.max = isNaN(intRangeValidation.max) ? null : intRangeValidation.max;
```
alextanlan: The idea is correct.
There is one note only. 
Our API spec says that `min` and `max` can be null (not sure if it is the valid case)
ed

So, more general would be to check whether it is not null and not undefined, 
e.g. like 
`numericBlock.min = intRangeValidation.min == null ? null : intRangeValidation.min;`

juanseba21: that's a good point. Null check added already

alextanlan (approved):

PR #1608, opened 2022-07-29
DDP-8466 | absent ddpParticipantId

ScalaChamp:
Ticket: https://broadinstitute.atlassian.net/browse/DDP-8466

Problem: The client code tries to read `ddpParticipantId` from `dsm.participant` which may be absent due to the fact that some participants don't have any record in DB thereby don't have anything under `dsm.participant` too, which is common case.

Solution: Change code so that `ddpParticipantId` is read from `dsm.data.profile`, it's guaranteed to be there.


GiCharkviani (approved):

SimoneMaiwald:
I think we should also move that fix into the RC

SimoneMaiwald:
Can you also check if there are parts of the code where we use the missing one instead of GUID

PR #1609, opened 2022-07-29
[DSM] DDP-7896 - send symbolic date if N/A is chosen

MishaK1204:
We were sending the value 'N/A' to the backend when N/A was chosen in the date picker. 
I have created the defaultDate variable and now, if N/A is chosen we are saving the default date instead of 'N/A'. Also if we get the default date from the backend we are showing N/A to the user.

https://broadinstitute.atlassian.net/browse/DDP-7896


SimoneMaiwald (approved):

SimoneMaiwald:
I think that change should also go into the RC for DSM

PR #1614, opened 2022-08-01
RC [DSM] DDP-7896 - send symbolic date if N/A is chosen

MishaK1204:
We were sending the value 'N/A' to the backend when N/A was chosen in the date picker.
I have created the defaultDate variable and now, if N/A is chosen we are saving the default date instead of 'N/A'. Also if we get the default date from the backend we are showing N/A to the user.

https://broadinstitute.atlassian.net/browse/DDP-7896

### DIFF
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.ts
@@ -59,7 +59,7 @@ export class FieldDatepickerComponent implements OnInit, OnChanges {
   }
 
   private handleInput(dateString: string): void {
-    if (!['N/A', 'Not Found', this.defaultNaDate].includes(dateString)) {
+    if (this.isDateNotFound(dateString)) {
       if (dateString != null) {
         let tmpDate: string = dateString;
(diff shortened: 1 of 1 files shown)


ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.ts
```diff
@@ -58,7 +59,7 @@ export class FieldDatepickerComponent implements OnInit, OnChanges {
   }
 
   private handleInput(dateString: string): void {
-    if (dateString !== 'N/A' && dateString !== 'Not Found') {
+    if (dateString !== 'N/A' && dateString !== 'Not Found' && dateString !== this.defaultDate) {
```
alextanlan: The date values `N/A`, `Not Found`, `1000-01-01` are used in several files.
It would be nice to have an enum for values.

alextanlan: `(dateString !== 'N/A' && dateString !== 'Not Found' && dateString !== this.defaultDate)`
can be shortened like
`!['N/A', 'Not Found', this.defaultDate].includes(dateString)`

PR #1634, opened 2022-08-05
Ddp 8498 pdf

gigijanelidze26:
DDP-8498
update pdf with google cloud


ddp-workspace/projects/ddp-brugada/src/app/components/header/header.component.ts
```diff
@@ -14,6 +14,7 @@ import { filter, first, tap } from 'rxjs/operators';
 export class HeaderComponent implements OnInit {
   Route = Route;
   private _isNavigationShown = false;
+  public pdfUrl = 'https://storage.googleapis.com/cmi-study-dev-assets/brugada/pdf/for-physicians.pdf';
```
GiCharkviani, line 17: **#OPTIONAL**
I suggest making this property read-only, or even if it's possible, to move it from here into another const variables file.

GiCharkviani (approved):

PR #1638, opened 2022-08-08
[DSM] DDP-8494 Download from signed URL

pegahtah:


ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/models/data.model.ts
```diff
@@ -42,9 +42,15 @@ export class Data {
         });
       }
     }
+    const parsedFiles = [];
+    if(json.files){
+      for(const f of json.files){
+        parsedFiles.push(File.parse(f));
+      }
+    }
```
alextanlan: **_#optional_** 
just fyi (don't have to fix)

```
 const parsedFiles = [];
  if(json.files){
    for(const f of json.files){
      parsedFiles.push(File.parse(f));
    }
  }
```
can be shortened like

` const parsedFiles = (json.files || []).map(file => File.parse(file)) `

ddp-workspace/projects/ddp-dsm-ui/src/app/services/dsm.service.ts
```diff
@@ -615,14 +615,21 @@ export class DSMService {
...
     const url = this.baseUrl + DSMService.UI + 'downloadFile';
     const map: { name: string; value: any }[] = [];
     map.push( {name: DSMService.REALM, value: realm} );
+    map.push( {name: 'ddpParticipantId', value: ddpParticipantId} );
     map.push( {name: 'fileName', value: fileName} );
     map.push( {name: 'bucket', value: bucketName} );
-    map.push( {name: 'blob', value: blob} );
-    return this.http.get( url, this.buildQueryBlobHeader( map )).pipe( catchError( this.handleError ) );
+    map.push( {name: 'blobName', value: blob} );
+    map.push( {name: 'fileGuid', value: fileGuid} );
```
alextanlan: **_#optional_**

```
const map: { name: string; value: any }[] = [];
map.push( {name: DSMService.REALM, value: realm} );
... 
map.push( {name: 'ddpParticipantId', value: ddpParticipantId} );
```

can be shortened like

```
 const map: { name: string; value: any }[] = [
  {name: DSMService.REALM, value: realm},
  ...
  {name: 'ddpParticipantId', value: ddpParticipantId}
];
```

ddp-workspace/projects/ddp-dsm-ui/src/app/file-download/file-download.component.ts
```diff
@@ -21,47 +22,79 @@ export class FileDownloadComponent implements OnDestroy {
...
 
   }
 
   ngOnDestroy(): void {
     if (this.subscription != null) {
       this.subscription.unsubscribe();
     }
+    if (this.subscription2 != null) {
+      this.subscription2.unsubscribe();
+    }
```
alextanlan: it is possible to use multiple subscriptions in one Subscription instance. 
https://tinytip.co/tips/rxjs-subscription-add/

e.g. 
```
export class FileDownloadComponent {
  subscriptions = new Subscription();

  constructor(...) {...}

  downloadParticipantFile( file: File ): void {
    const subscription1 = this.dsmService.getSignedUrl(...);
    this.subscriptions.add(subscription1);
    ...
  }

  saveParticipantFile( url: any, type: string, fileName: string ): void {
     const subscription2 = this.dsmService.downloadFromSignedUrl( ... );
     this.subscriptions.add(subscription2);
     ...
  }

  ngOnDestroy(): void {
    this.subscriptions.unsubscribe();
  }
}

```

ddp-workspace/projects/ddp-dsm-ui/src/app/file-download/file-download.component.ts
```diff
@@ -21,47 +22,79 @@ export class FileDownloadComponent implements OnDestroy {
...
-  }
+//    const downloadURI = (uri, name) => {
+//      const link = document.createElement("a");
+//      link.download = name;
+//      link.href = uri;
+//      document.body.appendChild(link);
+//      link.click();
+//      document.body.removeChild(link);
+//    }
+//    downloadURI(url, fileName);
```
alextanlan: Is that commented block still needed ?

ddp-workspace/projects/ddp-dsm-ui/src/app/file-download/file-download.component.ts
```diff
@@ -21,47 +22,79 @@ export class FileDownloadComponent implements OnDestroy {
...
-      .subscribe( {
+    this.subscription = this.dsmService.getSignedUrl( this.participant.data.profile[ 'guid' ], file.fileName, file.bucket,
+      file.blobName, file.guid, realm ).subscribe( {
         next: data => {
-          this.saveParticipantFile( data, file.mimeType, file.fileName );
+          this.saveParticipantFile( data.url, file.mimeType, data.fileName );
+        },
+        error: err => {
+          this.downloadMessage = err;
+          this.downloading = false;
```
alextanlan: The block 

```
this.downloadMessage = some message;
this.downloading = false;
```
is used 4 times in the file.

Perhaps, it is worth extracting it as a separate method?

Also, the static/constant messages for `this.downloadMessage` can be kept as an enum.

ddp-workspace/projects/ddp-dsm-ui/src/app/file-download/file-download.component.ts
```diff
@@ -21,47 +22,79 @@ export class FileDownloadComponent implements OnDestroy {
...
-    if (file.scanResult !== this.CLEAN) {
+    if (!this.isFileClean( file )) {
       this.downloadMessage = 'Error - file has not passed scanning';
       this.downloading = false;
       return;
     }
     const realm = localStorage.getItem( ComponentService.MENU_SELECTED_REALM );
-    this.subscription = this.dsmService.downloadParticipantFile( file.fileName, file.bucket, file.blobName, realm, file.mimeType )
-      .subscribe( {
+    this.subscription = this.dsmService.getSignedUrl( this.participant.data.profile[ 'guid' ], file.fileName, file.bucket,
```
alextanlan: it would be nice to limit number of arguments (for instance by using [TS object destructuring](https://www.typescriptlang.org/docs/handbook/variable-declarations.html#object-destructuring))

e.g. 
`this.dsmService.getSignedUrl( this.participant.data.profile[ 'guid' ], file.fileName, file.bucket, file.blobName, file.guid, realm)`

=> `this.dsmService.getSignedUrl( this.participant.data.profile[ 'guid' ], file, realm)`

Then in the service method:
 - either 
 ```
  public getSignedUrl( ddpParticipantId: string, file: File, realm: string) {
     const {fileName, bucket, blobName, guid} = file;
     ...
  }
```
- or short version  
`public getSignedUrl( ddpParticipantId: string, {fileName, bucket, blobName, guid}, realm: string) {...}`

- or short version with property renaming
`public getSignedUrl( ddpParticipantId: string, {fileName: newRenamedVariable, bucket, blobName, guid}, realm: string) {...}`

File: ddp-workspace/projects/ddp-dsm-ui/src/app/file-download/file-download.component.ts (hunk 4 above)
Comment (pegahtah) [reply ddp-workspace/projects/ddp-dsm-ui/src/app/file-download/file-download.component.ts]: fixed

File: ddp-workspace/projects/ddp-dsm-ui/src/app/services/dsm.service.ts (hunk 2 above)
Comment (MocanaAtBroad) [reply ddp-workspace/projects/ddp-dsm-ui/src/app/services/dsm.service.ts]: I think this can be shortened even further!
Looking at the http.get method signature it looks like a simple object with strings as values will work.
```typescript
const headerParamMap: {[header: string]: string} = {realm, ddpParticipantId, fileName, bucketName: bucket, etc, etc};
return this.http.get( url, headerParamMap).pipe( catchError( this.handleError ) );
```
(Not sure if you know this little object literal creation syntax trick, but if the variable inside the braces for the literal object is the name of the property, you don't have to say the property name, so this is why I can just say `{fileName}` instead of `{'fileName': fileName}`. 
I suggest renaming the arguments for the method so as many as possible match the property names for the object you are trying to construct.

ddp-workspace/projects/ddp-dsm-ui/src/app/file-download/file-download.component.ts
```diff
@@ -21,47 +22,79 @@ export class FileDownloadComponent implements OnDestroy {
...
+      file.blobName, file.guid, realm ).subscribe( {
         next: data => {
-          this.saveParticipantFile( data, file.mimeType, file.fileName );
+          this.saveParticipantFile( data.url, file.mimeType, data.fileName );
+        },
```
MocanaAtBroad: When using rxjs it is generally considered bad practice to call another observable from the `.subscribe` handler.
Also, in current code you are creating subscriptions each time the method `downloadParticipantFile` is being called.

You are assigning the subscription to a variable to later close (good!), but every time the method gets called, you create a new subscription and the previous subscription goes unassigned and unclosed so you end up with a memory leak anyways.


A single pipeline is going to be cleaner simpler and you wont even have to worry about closing subscriptions.
So you want to do something like:
```typescript
this.dsmService.getSignedUrl( this.participant.data.profile[ 'guid' ], etc.etc. ).pipe(
   .mergeMap(data -> downloadFromSignedUrl(data.url),
   .take(1))
   .subscribe({
        next: data => {
          const blob = new Blob( [ data ], {type: type} );
          fileSaver.saveAs( blob, fileName );
          this.downloading = false;
          this.downloadMessage = 'File download finished.';
        },
        error: err => {
          console.log( err );
          this.downloadMessage = err;
          this.downloading = false;
        }
      }
```
The take(1) will unsubscribe every time it gets called. You only need to the error handler once etc.
Try it and see if it flies.

ddp-workspace/projects/ddp-dsm-ui/src/app/services/dsm.service.ts
```diff
@@ -615,14 +616,21 @@ export class DSMService {
...
-    map.push( {name: 'blob', value: blob} );
-    return this.http.get( url, this.buildQueryBlobHeader( map )).pipe( catchError( this.handleError ) );
+    const map: { name: string; value: any }[] = [
+      {name: DSMService.REALM, value: realm} ,
+      {name: 'ddpParticipantId', value: ddpParticipantId} ,
+      {name: 'fileName', value: fileName} ,
+      {name: 'bucket', value: bucket},
+      {name: 'blobName', value: blobName},
+      {name: 'fileGuid', value: guid} ];
+    return this.http.get( url, this.buildQueryHeader( map )).pipe( catchError( this.handleError ) );
```
MocanaAtBroad, line 627: If I am reading code correctly handleError just throws an error. Might as well not do the catch here at all/does not do anything for you.
(I believe that the `error:` handler in the subscription will get the error. But you have to try it! Sometimes there are surprises!)

ddp-workspace/projects/ddp-dsm-ui/src/app/file-download/file-download.component.ts
alextanlan: Agree, that would be much better.

pegahtah: Thanks Marco! This was a great tip! It is fixed now

ddp-workspace/projects/ddp-dsm-ui/src/app/services/dsm.service.ts
pegahtah: For the header part, as you said, we can also just pass the file values, however we have to call `this.buildQueryHeader( map )` to be able to get required HTTP headers in there as well. (For some reason GitHub doesn't let me comment on the right thread for that, sorry!)

ddp-workspace/projects/ddp-dsm-ui/src/app/file-download/file-download.component.ts
pegahtah: For the header part, as you said, we can also just pass the file values, however we have to call this.buildQueryHeader( map ) to be able to get required HTTP headers in there as well.

ddp-workspace/projects/ddp-dsm-ui/src/app/services/dsm.service.ts
pegahtah: I think it's good if we keep this just in case we ever decide to actually do something for error handling? Currently it just throws the error so i think as you said, the `error:` part is supposed to catch it...

ddp-workspace/projects/ddp-dsm-ui/src/app/services/dsm.service.ts
```diff
@@ -1017,6 +1025,13 @@ export class DSMService {
       params
     };
   }
+
+  public buildQueryBlobHeaderForGCP(): any {
```
alextanlan: Seems like it should be a private method. It is used in the service only.

alextanlan (approved):

ddp-workspace/projects/ddp-dsm-ui/src/app/file-download/file-download.component.ts
```diff
@@ -1,67 +1,68 @@
...
+    if (!this.isFileClean( file )) {
+      this.setDownloadMessageAndStatus( this.NOT_SCANNED_FILE_MESSAGE, false );
       return;
     }
+
     const realm = localStorage.getItem( ComponentService.MENU_SELECTED_REALM );
-    this.subscription = this.dsmService.downloadParticipantFile( file.fileName, file.bucket, file.blobName, realm, file.mimeType )
-      .subscribe( {
+    this.dsmService.getSignedUrl( this.participant.data.profile[ 'guid' ], file, realm ).pipe(
+      mergeMap(data => this.dsmService.downloadFromSignedUrl(data['url']).pipe(take(1)))).subscribe( {
```
juanseba21, line 42: I think it's important to handle the subscription in order to make sure it is going to be unsubscribed after the component is destroyed. I suggest the following:

this.subscription = this.dsmService.getSignedUrl( this.participant.data.profile[ 'guid' ], file, realm ).pipe(
      mergeMap(data => this.dsmService.downloadFromSignedUrl(data['url']).pipe(take(1)))).subscribe(...)

MocanaAtBroad: Right. I don't think think that second .pipe() method is necessary.
I think the whole thing should look something like:
```typescript
this.dsmService.getSignedUrl(blah,blah).pipe(
        mergeMap(data => this dsmService.downloadFromSignedUrl(blah,blah)),
        take(1)).subscribe(blah,blah,blah);
```

juanseba21: That's ok , but my point is that we need to unsubscribe from the observavble once the component gets destroyed . That's what I think is still missing.

juanseba21: @alextanlan Please correct me if I'm wrong.

pegahtah: @MocanaAtBroad @juanseba21 I am sorry I missed these comments before merging my branch. I will make a branch for the fixes you requested

pegahtah: @MocanaAtBroad when writing it like `mergeMap(data => this.dsmService.downloadFromSignedUrl(data['url']), take(1)))` I get the error and can not build ` Type 'Observable<unknown>' is not assignable to type 'BlobPart'.
  Type 'Observable<unknown>' is missing the following properties from type 'ArrayBufferView': buffer, byteLength, byteOffset`

making it go through a pipe keeps the  Observable<ArrayBuffer> type

pegahtah: @juanseba21 adding the `take(1)` takes care of unsubscribing here, also HTTPClient takes care of unsubscribing itself

alextanlan: I agree that [Marco's comment](https://github.com/broadinstitute/ddp-angular/pull/1638#discussion_r942726640) is valid here. 
Using `take(1)` is enough for our case. 
So it will look like:

```
this.dsmService.getSignedUrl(this.participant.data.profile['guid'], file, realm)
    .pipe(
        mergeMap(data => this.dsmService.downloadFromSignedUrl(data['url'])),
        take(1)
    )
    .subscribe(...)
```
It will unsusbcribe after the first use.

alextanlan: As for the [Juan comment](https://github.com/broadinstitute/ddp-angular/pull/1638#discussion_r942543769), I suppose it is about the (rare) case when the component is destroyed but the subscription is still is not completed (like [an example](https://blog.bitsrc.io/6-ways-to-unsubscribe-from-observables-in-angular-ab912819a78f#:~:text=Beware%2C%20even%20if%20the%20AppComponent%20is%20destroyed%20thesubscription%24%20will%20not%20unsubscribe%20until%20the%20interval%20emits%20a%20value.))

I believe in the particular case it is not important (and may be even desired  - we wait for for the file gets downloaded and then unsubscribe even though the component is (unlikely) destroyed for some reason)

alextanlan: As an option/ a compromise we can use `takeUntil` operator instead of `take(1)` (see lots of its usages in our code).
Even it is overly here, imo.

PR #1640, opened 2022-08-08
[SINGULAR] PS: Cannot load contact information form after AOM

juanseba21:
'Provide Contact Information' form is now displaying properly:

Before the fix:

image

After the fix:

image

### DIFF
--- a/ddp-workspace/projects/ddp-singular/src/app/constants/activity-code.ts
+++ b/ddp-workspace/projects/ddp-singular/src/app/constants/activity-code.ts
@@ -12,4 +12,5 @@ export enum ActivityCode {
   AddParticipantParental = 'ADD_PARTICIPANT_PARENTAL',
   AddParticipantDependent = 'ADD_PARTICIPANT_DEPENDENT',
   MedicalRecordFileUpload = 'MEDICAL_RECORD_FILE_UPLOAD',
+  ChildContact = 'CHILD_CONTACT'
 }
--- a/ddp-workspace/projects/ddp-singular/src/app/utils/helpers.ts
+++ b/ddp-workspace/projects/ddp-singular/src/app/utils/helpers.ts
@@ -24,7 +24,7 @@ const getRenderActivityKeyByActivityCode = (activityCode: ActivityCode): RenderA
     return RenderActivityKey.Consent;
   }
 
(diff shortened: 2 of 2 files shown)


ddp-workspace/projects/ddp-singular/src/app/constants/activity-code.ts
```diff
@@ -12,4 +12,5 @@ export enum ActivityCode {
   AddParticipantParental = 'ADD_PARTICIPANT_PARENTAL',
   AddParticipantDependent = 'ADD_PARTICIPANT_DEPENDENT',
   MedicalRecordFileUpload = 'MEDICAL_RECORD_FILE_UPLOAD',
+  ChildContact = 'CHILD_CONTACT'
```
juanseba21, line 15: Adding CHILD_CONTACT as an activity-code

ddp-workspace/projects/ddp-singular/src/app/utils/helpers.ts
```diff
@@ -35,6 +35,8 @@ const getRenderActivityKeyByActivityCode = (activityCode: ActivityCode): RenderA
       return RenderActivityKey.MedicalRecordUpload;
     case ActivityCode.PatientSurvey:
       return RenderActivityKey.PatientSurvey;
+    case ActivityCode.ChildContact:
+      return RenderActivityKey.About;
```
juanseba21: Returning About as a Render-Activity-Code

alextanlan: I believe you have caught the bug root cause.

**One note**:  suggest moving that check to line 27 where the condition has already been returning `RenderActivityKey.About`

MocanaAtBroad:
This seems like it fixes problem. Only thing I am not 100% sure is if it is correct what the wizard looks like, i.e., the activity asking for the child's email appears in wizard under "About My Child".
I am inclined to say, let's merge to get fix in and then let's verify fix with Whitney and the rest.

MocanaAtBroad (dismissed):

ddp-workspace/projects/ddp-singular/src/app/utils/helpers.ts
juanseba21: sure, Makes sense.

alextanlan (approved):

PR #1641, opened 2022-08-09
[SINGULAR] DDP-8566 Placeholders and text in input fields still being cutoff in mobile

juanseba21:
Adding some styling changes on mobile view.

Before the fix:

image

After the fix:

image

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/dropdownActivityPicklistQuestion.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/dropdownActivityPicklistQuestion.component.ts
@@ -68,13 +68,13 @@ import { NGXTranslateService } from '../../../services/internationalization/ngxT
           width: auto;
         }
         @media only screen and (max-width: 1024px) {
-          ::ng-deep .mat-form-field .mat-select-native-control{
+          ::ng-deep .width .mat-select-native-control{
           line-height: 17px;
             }
-          ::ng-deep .mat-form-field .mat-select{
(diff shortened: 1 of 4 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/picklist/dropdownActivityPicklistQuestion.component.ts
```diff
@@ -66,6 +67,17 @@ import { NGXTranslateService } from '../../../services/internationalization/ngxT
...
+          ::ng-deep .mat-form-field .mat-select-native-control{
+          line-height: 17px;
+            }
+          ::ng-deep .mat-form-field .mat-select{
+            line-height: 17px;
+            }
+          ::ng-deep .mat-form-field .mat-form-field-infix input{
+            line-height: 17px;
+            }
+        }
```
alextanlan, line 80: suggest shortening lines 70-80 like
```
@media only screen and (max-width: 1024px) {
    ::ng-deep .mat-form-field {
        .mat-select-native-control, 
        .mat-select, 
        .mat-form-field-infix input{
            line-height: 17px;
        }
    }        
 } 
```

Same for `addressInput.component.ts`

juanseba21: I tried to put it in that way. But it did not work properly. I guess it's because we need to use an ampersand ('&') and it's not working since this is CSS embedded into the component.

alextanlan (approved):

PR #1650, opened 2022-08-16
DDP-8597/OS-static-page

llobzhanidze:
https://broadinstitute.atlassian.net/browse/DDP-8597 
change static page style and also linter fix commits
current version
space
previous version
space1


gigijanelidze26 (dismissed):

GiCharkviani (changes requested):
It looks like in this PR you have included linter fixes as well, which is described neither in commit nor in the description.
Also, could you please include an image of the visual change (previous and current versions would be desirable).

GiCharkviani (approved):

PR #1658, opened 2022-08-18
DDP-8628: [DSM] adding explicit human-friendly export option

devonbush:
We would like to give the users an explicit choice between human readable and analysis friendly formatting. This adds a single new variable "humanReadable" which consolidates a choice between splitting multiselect questions into multiple columns, and rendering option stable ids vs. the display text. 

image

This is paired with https://github.com/broadinstitute/ddp-study-server/pull/2175

TO TEST:

open the participant list for the AT study
customize view to include the Medical History "telangiectasia" question
export the participant list, selecting "analysis-friendly"
confirm the telangiectasia responses are split into 4 columns - 1 each for eyes and skin, and 1 for each of the supporting details
export the participant list again, selecting "human-readable"
confirm the telangiectasia responses are shown in two columns -- one showing the main responses as dispaly text, and the other showing the consolidated details

### DIFF
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/services/dsm.service.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/services/dsm.service.ts
@@ -214,7 +214,7 @@ export class DSMService {
   }
 
   public downloadParticipantData(realm: string, jsonPatch: string, parent: string, columns: {}, json: ViewFilter,
-                                 filterQuery: string, sortBy?: Sort, fileFormat?: string, splitOptions?: boolean,
+                                 filterQuery: string, sortBy?: Sort, fileFormat?: string, humanReadable?: boolean,
                                  onlyMostRecent?: boolean):
     Observable<any> {
     const viewFilterCopy = this.getFilter(json);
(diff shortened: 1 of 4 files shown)


ddp-workspace/projects/ddp-dsm-ui/src/app/services/dsm.service.ts
```diff
@@ -230,10 +230,10 @@ export class DSMService {
     if (fileFormat) {
       map.push({name: 'fileFormat', value: fileFormat});
     }
-    if (typeof splitOptions === 'boolean') {
-      map.push({name: 'splitOptions', value: splitOptions});
+    if (typeof humanReadable === 'boolean') {
```
GiCharkviani, line 233: Can the "humanReadable" ever be another type than the boolean?

GiCharkviani (approved):

ddp-workspace/projects/ddp-dsm-ui/src/app/services/dsm.service.ts
devonbush: no, but based on my reading, this was the best/safest way to undefined-safe the method

PR #1683, opened 2022-08-26
[DSM] DDP-7606 received by

SimoneMaiwald:
adding received by to the response for kitrequests endpoint to be able to display "research sample" or "clinical sample" for LMS/ OSTEO

1) go to sample > receive or overview page for lms
2) select saliva kit

Screen Shot 2022-08-26 at 4 42 39 PM

### DIFF
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/shipping/shipping.component.html
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/shipping/shipping.component.html
@@ -63,116 +63,169 @@
              [mfSortBy]="sortField" [(mfSortOrder)]="sortDir">
         <thead>
         <tr>
-          <th *ngIf="getRole().allowedToHandleSamples() && (queueToPrint() || shippingPage === UPLOADED) && !kitType.externalShipper" class="Line--Break">Print Kit
+          <th *ngIf="getRole().allowedToHandleSamples() && (queueToPrint() || shippingPage === UPLOADED) && !kitType.externalShipper" class="Line--Break">
+            <!--            checkbox to select for printing label -->
+            Print Kit
 <mat-checkbox color="primary" disableRipple [checked]="allSelected" (change)="allSelected = $event.checked; allChecked()"></mat-checkbox></th>
           <th>
+            <!--            shortID-->
             <mfDefaultSorter by="collaboratorParticipantId">Short ID</mfDefaultSorter>
           </th>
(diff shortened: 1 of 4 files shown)


ddp-workspace/projects/ddp-dsm-ui/src/app/shipping/shipping.component.html
```diff
@@ -182,50 +235,89 @@ <h4>There are no kit requests</h4>
...
+          </td>
+          <td [ngClass]="{'highlight_sample': kitRequest.receivedBy === 'BSP'}">
+            <!--            kit type-->
+            {{kitRequest.kitTypeName}}
+          </td>
+          <td *ngIf="isClinicalStudy && (shippingPage === RECEIVED || shippingPage === OVERVIEW)" [ngClass]="{'highlight_sample': kitRequest.receivedBy === 'BSP'}">
+            <!--            sample type-->
+            <ng-container *ngIf="kitRequest.receivedBy === 'MERCURY'">Clinical Sample</ng-container>
+            <ng-container *ngIf="kitRequest.receivedBy === 'BSP'">Research Sample</ng-container>
+          </td>
```
pegahtah, line 317: Based on the ticket, shouldn't we also display `Research kit`   if `kit.sequencingRestriction` is marked as research only?

ddp-workspace/projects/ddp-dsm-ui/src/app/shipping/shipping.component.ts
```diff
@@ -276,6 +279,9 @@ export class ShippingComponent implements OnInit {
               if (kit.noReturn) {
                 this.kitsWithNoReturn = true;
               }
+              if(kit.receivedBy === 'MERCURY') {
+                this.isClinicalStudy = true;
+              }
```
pegahtah, line 284: If I understood the means behind this correctly, I think this means that before getting any kits received by "Mercury", the `this.isClinicalStudy` will always be false, which means that we won't be seeing `SampleType` column. Can't we use instance settings for this?

SimoneMaiwald: Mercury is the main receiver for this study, so as soon as we receive 1 "normal" kit for this study we will display the info. no need to add an additional call to the backend to get the information from instance settings

ddp-workspace/projects/ddp-dsm-ui/src/app/shipping/shipping.component.html
SimoneMaiwald: that would be on the changes for DDP-8569. This ticket is only for adding the highlighting there

PR #1692, opened 2022-08-30
[DSM] DDP-8624 new changes

pegahtah:
Changing from `Object.assign(this.selectedColumns, filteredColumns)` in participant-list.component.ts to just normal object assignment which was causing the main error
Also I changed the `extractDefaultColumns()` to `getDefaultColumns()` to find the columns from `sourceColumns` instead of `selectedColumns`, otherwise the not selected default column could get lost

To test: select a saved filter from the saved filters list and apply it by clicking on the funnel sign.

Then select `Reload With Default Filter` , desired behavior is that the columns from the default column should appear and all the other columns should be gone

### DIFF
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/participant-list.component.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/participant-list.component.ts
@@ -1125,9 +1125,8 @@ export class ParticipantListComponent implements OnInit {
       const val = value as Filter[];
       const newVal = [];
       val.forEach(el => {
-        let defaultColumn = this.defaultColumns.find(col => {
-          el['participantColumn']['name'] === col['participantColumn']['name']
-        });
+        const defaultColumn = this.defaultColumns.find(col =>
+          el['participantColumn']['name'] === col['participantColumn']['name']);
         if (defaultColumn) {
           newVal.push(el);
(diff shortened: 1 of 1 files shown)


ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/participant-list.component.ts
```diff
@@ -1077,10 +1077,10 @@ export class ParticipantListComponent implements OnInit {
       this.setSavedFilterColumnsIfSelected();
       return;
     }
-    const filteredColumns = this.extractDefaultColumns(this.selectedColumns);
-    Object.assign(this.selectedColumns, filteredColumns);
+    const filteredColumns = this.getDefaultColumns();
+    this.selectedColumns = filteredColumns;
```
pegahtah, line 1081: The main cause of the bug was here. The `Objects.assign` just copied over everything from filteredColumns to selectedColumns, without removing what was in the selectedColumn before. So the old survey columns stayed there

alextanlan:
Can you please fix [linter warnings](https://app.circleci.com/pipelines/github/broadinstitute/ddp-angular/19260/workflows/2b2173a6-4360-4916-909e-4195fcf06191/jobs/28843)?

GiCharkviani (approved):

PR #1708, opened 2022-09-07
DDP-8697 remove underlines from lms links

LKapanad:
DDP-8697 remove underlines from lms links


GiCharkviani (approved):

GiCharkviani (changes requested):
Please check tests

GiCharkviani:
Looks like @gigijanelidze26 already has merged the same [fix](https://github.com/broadinstitute/ddp-angular/pull/1710)

PR #1723, opened 2022-09-11
[SINGULAR] DDP-8723: Required field error in Medical Release Form is not cleared immediately after selecting an method

juanseba21:
the bug was fixed by adding an extra logic when a value changes as part of  'ActivityQuestion' component:

image
image

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityQuestion.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityQuestion.component.ts
@@ -30,7 +30,7 @@ import { LayoutType } from '../../../models/layout/layoutType';
                 [studyGuid]="studyGuid"
                 [activityGuid]="activityGuid"
                 [layoutType]="layoutType"
-                (valueChanged)="enteredValue$.next($event)"
+                (valueChanged)="changeValueAction($event)"
                 (componentBusy)="componentBusy.next($event)">
             </ddp-activity-answer>
             <ng-container *ngIf="block.shown && block.enabled">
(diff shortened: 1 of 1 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity-blocks/activityQuestion.component.ts
```diff
@@ -80,6 +80,10 @@ export class ActivityQuestionComponent implements OnInit, OnDestroy {
         this.setupSavingData();
         this.setupScrollToErrorAction();
     }
+    public changeValueAction(answerValueEvent:AnswerValue){
+        this.enteredValue$.next(answerValueEvent);
+        this.validationRequested$.next(!!answerValueEvent);
+    }
```
MocanaAtBroad, line 86: Seems to fix problem/ error message is definitely cleared. However I think this is a risky fix as it changes the base behavior for every question in every project.
And updating validationRequested$ observable from this method actually does not make a lot of sense/goes against how the whole object is designed to work.
But I think you are probably close to the source of the problem!. I wonder if the root of the problem is when an how that validationRequested$ is called/set from the parent component?
So I think it freakin' awesome that we have a fix! But please look deeper. Why is validationRequested$ not emitting in normal operations? Why is it that the error message does not clear in this activity but seems to clear in other ones?

juanseba21: According to what I have found, the 'validationRequested$' observable is being handled only by the submit button at the bottom of the page, and that's why it's necessary to be handled when any value from the answers changes. That's why it makes sense to me since the activityQuestion component is in charge of taking care of values changing. The main danger with this move, in my opinion, is that we are propagating validations everytime a value changes.
Anyway, we can investigate further to come up with a better idea.

alextanlan: It works except cases when `answerValueEvent` is falsy (e.g. `false` value for boolean questions).

Can you please take a look at an alternative [PR#1724](https://github.com/broadinstitute/ddp-angular/pull/1724)?
@MocanaAtBroad , @juanseba21

alextanlan: The bug is fixed in PR#1724. 
Let's close the PR, shall we?
@juanseba21

PR #1724, opened 2022-09-12
[SDK] [SINGULAR] DDP-8723 Fix a bug when a field error is not cleared immediately

alextanlan:
The issue description is in the [DDP-8723](https://broadinstitute.atlassian.net/browse/DDP-8723) ticket.

### The root cause of the bug

We set an error message for any question with a failed validation like: 
```
  private setupErrorMessage(): void {
        const firstFailedValidator$ =  combineLatest([
            this.enteredValue$.pipe(startWith(this.block.answer as AnswerValue)),
            this.validationRequested$  // <= PAY ATTENTION TO HERE
        ]).pipe(
            // not displaying any local validations until a validation is requested
            filter(([_, validationRequested]) => !!validationRequested),
            ...
            // Then set/reset an error message
```
The `validationRequested` flag is set manually: 

- on **Submit** always to **true**
  (see `ddp-sdk\src\lib\components\activityForm\baseActivity.component.ts` -> `setupSubmitPipeline()` => `this.setFlags(true);`)  
- on **Next** button click firstly to **true** then with a delay to **false**  
  (see `ddp-sdk\src\lib\components\activityForm\activity.component.ts` => `incrementStep(..)` 
      => ` this.validationRequested = true / false;`)

#### Thus, when we set `this.validationRequested = false;` (after the **Next** button click) the error message handler is not called even the invalid field answer was changed to valid one.

Seems like we have never had such cases before. I mean failed vaidation questions on a section that is not the last one (with Submit button).

--------------

@juanseba21 suggested a [fix PR#1723](https://github.com/broadinstitute/ddp-angular/pull/1723).

However, If we trigger `validationRequested$` by every answer value change it will lead to firing the related handlers very often
(e.g. `firstFailedValidator$` in `ddp-sdk\src\lib\components\activityForm\activity-blocks\activityQuestion.component.ts`).
Even though we have a filter there in order not  to pass falsy `validationRequested$`.

I went deeper as to have a more general fix.
Let's force the failed section validation after **Next** button click, shall we?

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
@@ -212,7 +212,8 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit, 
                         // delay needed for validationRequested to be processed
                         delay(0),
                         tap(() => {
-                            this.validationRequested = false;
+                            this.validationRequested = !this.currentSection.valid;
+
                             this.sendSectionAnalytics();
                             if (this.currentSection.valid) {
                                 // reset scrolling signal and disable a local validation


ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -212,7 +212,8 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
                         // delay needed for validationRequested to be processed
                         delay(0),
                         tap(() => {
-                            this.validationRequested = false;
+                            this.validationRequested = !this.currentSection.valid;
```
juanseba21, line 215: The validation is going to be triggered only when the user clicks on the next button, is that right?

juanseba21 (approved):

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
alextanlan: In scope of the ticket - yes. 
The flag is set to **true** on line 210.
Then if the section is: 
 -  **valid**  -> we reset `validationRequested ` flag as not to have additional checks.
 -  **invalid**  -> we have `validationRequested ` flag as true in order to have additional checks (after any answer change).

PR #1725, opened 2022-09-12
[DSM] DDP-8734 display of mercury pdo number

pegahtah:
DDP-8734 showing the pdo order number in frontend. I have added the field to `SequencingOrder` model.Screen Shot 2022-09-12 at 2 15 32 PM

### DIFF
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/sequencing-order/sequencing-order.model.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/sequencing-order/sequencing-order.model.ts
@@ -14,7 +14,7 @@ export class SequencingOrder {
   }
 
   public static getEmptyInstance(): SequencingOrder{
-    return new SequencingOrder( '', '', '', '', '', '', '', '', '', '', '' );
+    return new SequencingOrder( ...Array(11).fill('') );
   }
 
 }
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/sequencing-order/sequencing-order.model.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/sequencing-order/sequencing-order.model.ts
@@ -3,17 +3,17 @@ export class SequencingOrder {
 
   constructor( public sampleType: string, public sample: string, public sampleStatus: string, public collectionDate: string,
                public sequencingOrderDate: string, public tissueId: string, public dsmKitRequestId: string,
-               public sequencingRestriction: string, public lastStatus: string, public lastOrderNumber: string ) {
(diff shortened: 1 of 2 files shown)


ddp-workspace/projects/ddp-dsm-ui/src/app/sequencing-order/sequencing-order.model.ts
```diff
@@ -3,17 +3,18 @@ export class SequencingOrder {
...
 
   public static parse( json ): SequencingOrder {
     return new SequencingOrder( json.sampleType, json.sample, json.sampleStatus, json.collectionDate, json.sequencingOrderDate,
-      json.tissueId, json.dsmKitRequestId, json.sequencingRestriction, json.lastStatus, json.lastOrderNumber );
+      json.tissueId, json.dsmKitRequestId, json.sequencingRestriction, json.lastStatus, json.lastOrderNumber, json.pdoOrderId );
   }
 
   public static getEmptyInstance(): SequencingOrder{
-    return new SequencingOrder( '', '', '', '', '', '', '', '', '', '' );
+    return new SequencingOrder( '', '', '', '', '', '', '', '', '', '', '' );
```
alextanlan: **_just a tip fyi:_**

In this particular case:
the arguments (11 empty spaces) can be simplified as to not miss something
like ` return new SequencingOrder(  ...Array(11).fill('') );`

Another option is to have default values in the class constructor:

```
constructor( 
  public sampleType: string = '', public sample: string = '', public sampleStatus: string ='', 
  public collectionDate: string = '', public sequencingOrderDate: string = '', public tissueId: string ='',
  public dsmKitRequestId: string = '', public sequencingRestriction: string ='', public lastStatus: string ='', 
  public lastOrderNumber: string ='' , public pdoOrderId: string = ''
) { }

public static getEmptyInstance(): SequencingOrder{
    return new SequencingOrder(); // default values will be used
}
```

GiCharkviani: Or you could just use static factory methods like it's mentioned in this [article](https://medium.com/young-coder/supporting-multiple-constructors-in-javascript-48cab12264ed) (As far as I know, we are already using that approach in several classes)

alextanlan (dismissed):

ddp-workspace/projects/ddp-dsm-ui/src/app/sequencing-order/sequencing-order.model.ts
pegahtah: I applied the `return new SequencingOrder( ...Array(11).fill('') );` because I am not sure if I want it as a default... I'd rather they default to undefined...

SimoneMaiwald (approved):

PR #1730, opened 2022-09-14
[SDK][SINGULAR]DDP-8744: Scroll-up to invalid field, MRRF page 1 works only once

juanseba21:
Some adjustments were made in order to fix scroll-up functionality.
It's important to test it for the projects that are using the same components.

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
@@ -215,7 +215,7 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit, 
                         delay(0),
                         tap(() => {
                             this.validationRequested = !this.currentSection.valid;
-                            
+
                             this.sendSectionAnalytics();
                             if (this.currentSection.valid) {
                                 // reset scrolling signal and disable a local validation
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
@@ -193,6 +193,8 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit, 
     public incrementStep(scroll: boolean = true): void {
         const nextIndex = this.nextAvailableSectionIndex();
         if (nextIndex !== -1) {
+            //validationRequested needs to be reset before starting validation.
+            this.validationRequested=false;
             this.submitAnnouncementService.announceSubmit();
             // The announcement could make listener components busy, but not instantly
             // introduce a wait before we check whether we busy or not
@@ -213,7 +215,7 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit, 
                         delay(0),
                         tap(() => {
                             this.validationRequested = !this.currentSection.valid;
-
+                            
                             this.sendSectionAnalytics();
                             if (this.currentSection.valid) {
                                 // reset scrolling signal and disable a local validation
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
@@ -18,7 +18,6 @@ import {
     catchError,
     concatMap,
     debounce,
-    debounceTime,
     distinctUntilChanged,
     filter,
     finalize,
@@ -87,7 +86,7 @@ interface AddressSuggestion {
     template: `
         <p *ngIf="block.titleText" class="ddp-address-embedded__title" [innerHTML]="block.titleText"></p>
         <p *ngIf="block.subtitleText" class="ddp-address-embedded__subtitle" [innerHTML]="block.subtitleText"></p>
-        <div scroll-up [triggerScrollUp]="(scroll_up$|async)" (scrollUpExecuted)="scrollUpExecutedAction($event)">
+        <div scroll-up [triggerScrollUp]="(scroll_up$|async)" (scrollUpExecuted)="scrollUpExecutedAction()">
         <ddp-address-input
             (valueChanged)="inputComponentAddress$.next($event); dirtyStatusChanged.emit(true)"
             (formValidStatusChanged)="formValidStatusChanged$.next($event)"
@@ -154,7 +153,6 @@ interface AddressSuggestion {
 })
 export class AddressEmbeddedComponent implements OnDestroy, OnInit {
     @Input() block: MailAddressBlock;
-    @ViewChild('scrollAnchor', {static: true}) scrollAnchor: ElementRef;
     scroll_up$: Observable<boolean>;
 
     @Input()
@@ -266,13 +264,11 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
 
     private setupScrollToErrorAction(): void {
         this.scroll_up$ = this.validationRequested$.pipe(
-            map(validationRequested => validationRequested && this.block.scrollTo),
-            debounceTime(300)
+            map(validationRequested => validationRequested && this.block.scrollTo)
         );
     }
-    scrollUpExecutedAction(executed:boolean): void {
-        this.block.scrollTo=!executed;
-        this.validationRequested$.next(!executed);
+    scrollUpExecutedAction(): void {
+        this.block.scrollTo=false;
     }
 
     private initializeComponentState(): void {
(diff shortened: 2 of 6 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -249,13 +256,34 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
...
+        @Optional() private submitService: SubmitAnnouncementService,
+        private windowRef: WindowRef
     ) {
         this.suggestionForm = new FormGroup({
             suggestionRadioGroup: new FormControl('entered')
         });
         this.initializeComponentState();
     }
+    
+    private setupScrollToErrorAction(): void {
```
juanseba21: Since it manages scroll-up functionality in the case of the "addressEmbedded" component, it was necessary to include this method.

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/MailAddressBlock.ts
```diff
@@ -31,7 +31,5 @@ export class MailAddressBlock extends ActivityBlock {
         return true;
     }
 
-    public shouldScrollToFirstInvalidQuestion(): boolean {
```
juanseba21, line 34: Because the MailAddressBlock class extends from the ActivityBlock class and this method is already defined elsewhere, this one was eliminated. I do not believe it is necessary to create an override method that only returns false. We can reconsider it if it was designed for another purpose, though, as it is now causing problems.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -249,13 +256,34 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
...
+
+        this.validationRequested$.pipe(
+            tap((validationRequested)=>{console.log(this.block);console.log(validationRequested); debugger;}),
```
alextanlan: I believe we should remove that debug stuff.

juanseba21: done

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
```diff
@@ -193,6 +193,8 @@ export class ActivityComponent extends BaseActivityComponent implements OnInit,
     public incrementStep(scroll: boolean = true): void {
         const nextIndex = this.nextAvailableSectionIndex();
         if (nextIndex !== -1) {
+            //validationRequested needs to be reset before starting validation.
+            this.validationRequested=false;
```
alextanlan, line 197: Good catch! 👍

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
alextanlan: That makes sense for me. 
After the fix, we will have same scroll up to a failed validation field feature for Mailing address as well.

ddp-workspace/projects/ddp-sdk/src/lib/models/activity/MailAddressBlock.ts
alextanlan: Historically, we block the "scroll to first invalid question" feature for ContentBlock, InstitutionBlock and for **MailAddressBlock**. 

@MocanaAtBroad do you remember whether it has been done in purpose ? 
I mean for the Mailing address.

I suppose we would like to have that feature for Mailing Address as well (according to the Juan's fix).

alextanlan:
@juanseba21 can you please fix failed tests ?

ddp-workspace/projects/ddp-sdk/src/lib/directives/scroll-up.directive.ts
```diff
@@ -0,0 +1,27 @@
+import { Directive, ElementRef, EventEmitter, Inject, Input, OnChanges, OnInit, Output, SimpleChanges } from "@angular/core";
+import { ConfigurationService } from "../services/configuration.service";
+import { WindowRef } from "../services/windowRef";
+
+@Directive({
+    selector: '[scroll-up]'
+})
+export class ScrollUpDirective implements OnChanges {
```
juanseba21, line 8: I've created a new directive in order to make it reusable since there are some components that are using the scroll-up functionality.

juanseba21:
> @juanseba21 can you please fix failed tests ?

done

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -257,6 +263,17 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
...
     }
 
+    private setupScrollToErrorAction(): void {
+        this.scroll_up$ = this.validationRequested$.pipe(
+            map(validationRequested => validationRequested && this.block.scrollTo),
+            debounceTime(300)
+        );
+    }
+    scrollUpExecutedAction(executed: boolean): void {
+        this.block.scrollTo=!executed;
```
alextanlan: We always set `this.block.scrollTo = false;` in this case (after a page was scrolled up to a failed question/component).
No need to pass a boolean argument here.

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
```diff
@@ -257,6 +263,17 @@ export class AddressEmbeddedComponent implements OnDestroy, OnInit {
...
+        this.validationRequested$.next(!executed);
```
alextanlan: I believe we should not trigger validationRequested$ flag here.
The main logic with `validationRequested$` is in activity.component related to Next/Submit buttons click behavior.

I guess, the reason you added it is a `debounceTime` on line 269.
The debounce prevents the scroll-up directive working, can be removed.

alextanlan: So, eventually it should be like:

```
  private setupScrollToErrorAction(): void {
        this.scroll_up$ = this.validationRequested$.pipe(
            map(validationRequested => validationRequested && this.block.scrollTo),
        );
  }

  scrollUpExecutedAction(): void {
      this.block.scrollTo = false;
  }


```

ddp-workspace/projects/ddp-sdk/src/lib/components/activityForm/activity.component.ts
alextanlan: Actually, it is the bug fix. 

Other stuff is just an improvement/refactoring.

ddp-workspace/projects/ddp-sdk/src/lib/directives/scroll-up.directive.ts
alextanlan: In general, it is a good idea.
Why would not we use it in `ddp-sdk\src\lib\components\activityForm\activity-blocks\activityQuestion.component.ts` as well?

ddp-workspace/projects/ddp-sdk/src/lib/directives/scroll-up.directive.ts
```diff
@@ -0,0 +1,29 @@
...
+        @Inject('ddp.config') private configuration: ConfigurationService,
+        private windowRef: WindowRef) {}
+
+    ngOnChanges(changes: SimpleChanges): void {
+        if (changes.triggerScrollUp.currentValue) {
+            const headerOffset = this.configuration.scrollToErrorOffset;
+            const top = this.eleRef.nativeElement.getBoundingClientRect().top
+                + this.windowRef.nativeWindow.scrollY - headerOffset;
+            this.windowRef.nativeWindow.scrollTo({
+                top,
```
alextanlan: **_#minor_**

we can extract getting the top data (lines 18-20) to a separate method, like:
```
 this.windowRef.nativeWindow.scrollTo({
    top: this.getTop(),
    behavior: 'smooth'
});

private getTop(): number {
    ...
    return top;
}
```

ddp-workspace/projects/ddp-sdk/src/lib/directives/scroll-up.directive.ts
juanseba21: done

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
juanseba21: done

ddp-workspace/projects/ddp-sdk/src/lib/components/address/addressEmbedded.component.ts
juanseba21: done

ddp-workspace/projects/ddp-sdk/src/lib/directives/scroll-up.directive.ts
juanseba21: done

alextanlan (approved):

PR #1736, opened 2022-09-16
plotly-npm-package

GiCharkviani:


MishaK1204 (approved):

ddp-workspace/projects/ddp-dsm-ui/src/app/services/role.service.ts
```diff
@@ -183,7 +183,7 @@ export class RoleService {
   }
 
   public allowedToViewEELData(): boolean {
-    return this._isViewingEEL;
```
GiCharkviani, line 186: This is temporary, for testing purposes

PR #1761, opened 2022-09-20
*DON'T MERGE!*  [SINGULAR] DDP-8828: Implement a feature flag as to show/hide a captcha

alextanlan:
Implemented a new feature flag for a possibility to show/hide our catcha on **Prequal** form.

By default, the feature flag is true (means the captcha is visible).
We can switch/trigger it manually - press `Ctrl+Alt+7` (Windows) / `Control+Option+7` (MacOS) 
and toggle the feature. 

captcha1
captcha2

### DIFF
--- a/ddp-workspace/projects/ddp-singular/src/app/components/activity/activity.component.html
+++ b/ddp-workspace/projects/ddp-singular/src/app/components/activity/activity.component.html
@@ -84,9 +84,12 @@
         <span class="last-updated">{{ model.lastUpdatedText }} </span>
       </section>
 
-      <section *ngIf="isPrequal" class="section section--captcha">
-        <re-captcha [siteKey]="captchaSiteKey" (resolved)="onCaptchaResolve()"></re-captcha>
-      </section>
+      <!-- we can hide captcha for testing purposes by the feature flag -->
+      <ng-container *ngIf="featureFlag_DDP_8828 | async">
+        <section *ngIf="isPrequal" class="section section--captcha">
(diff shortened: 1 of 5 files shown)


alextanlan:
@aweng98, @MocanaAtBroad 
what do you think about the implementation?

Is it possible to switch off the feature flag (as to hide the captcha) by auto-tests? 
something like 
```
beforeEach() {
   pressKeyCombination('Ctrl', 'Alt', '7');
   clickOnFeatureFlag(getelementById('feature DDP-8828 show captcha'));
   closeDialog() 
}
```
or should we proceed in another way ?

aweng98:
> 

I worry about duplicate code if we have to do it for every enrollment test. Here is an [example](https://github.com/all-of-us/workbench/blob/62edc3bda59a767b3a9e1dbe659bf583528eb75e/ui/src/environments/environment.prod.ts#L21) how All-of-Us project works with UI feature flags. For `enableCaptcha` flag, set it to  `false` once in `environment.*.ts` and it's all taking care of.

alextanlan:
@aweng98 do you mean that the captcha should be disabled/hidden by default for all our environments 
except the production one? 

And if it is needed we can show/switch it on by our feature flags dialog (e.g. for manual testing).
Is it correct?

juanseba21 (approved):

alextanlan:
I have suggested another option as to disable the captcha for auto-tests on test env. 
See [PR#1769](https://github.com/broadinstitute/ddp-angular/pull/1769).

PR #1766, opened 2022-09-21
[NEED TICKET] Update CircleCI nightly job with matrix

aweng98:


aweng98:
@MocanaAtBroad What do you thinik?

MocanaAtBroad:
That would be much nicer! Though I will share that we have talked about these nightly jobs in the team and had agreed that we should get rid of them.
The issue is that 1. They never really seem to have given us any value 2. Sometimes messes things up because, once in while, we will setup a study manually and the night job messes everything up.
So If it works the way you set it up, awesome cuz it is definitely better. But don't feel bad when remove it!

aweng98:
I agree with point # 1. Ideally, deploy should happen automatically when a PR is merged. Deploy nightly doesn't provide any benefit. I'll close PR but not delete the branch (in case we need it in future). Thanks.

PR #1769, opened 2022-09-22
[SINGULAR] DDP-8828: Add secret keys as to hide our captcha for auto-tests on test env

alextanlan:
Suggest another option to hide/disable our captcha for auto-tests (first option was in [PR-1761](https://github.com/broadinstitute/ddp-angular/pull/1761)).

The idea is to use default keys provided by the reCAPTCHA widget for automated tests. 
We can keep the client key in our Vault.

image

### DIFF
--- a/config/pepperConfig.js.ctmpl
+++ b/config/pepperConfig.js.ctmpl
@@ -317,6 +317,7 @@ var DDP_ENV = {
         DDP_ENV['auth0Audience'] = "{{.Data.auth0.loginDomain}}";
         DDP_ENV['basePepperUrl'] = "{{.Data.serverBaseUrl}}";
         DDP_ENV['recaptchaSiteClientKey'] = "{{.Data.recaptchaSiteClientKey}}";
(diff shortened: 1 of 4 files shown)


MocanaAtBroad:
Nevermind: I see what is going on. Let me see if I can stick those values into Vault for you.
~~ ~~Aleksei: I see you are providing a way to make the key value available in the activity component. But what are we supposed to do with that value?~~ ~~

alextanlan:
Marco:
  If I got it correctly from [the docs](https://developers.google.com/recaptcha/docs/faq#id-like-to-run-automated-tests-with-recaptcha.-what-should-i-do) there is a couple of hardcoded keys specially for auto-tests purposes.
That client key tells the captcha not to execute the full check with images. 

However, it is worth checking.

alextanlan:
Update: checked it locally. One key is enough.

juanseba21 (dismissed):

MocanaAtBroad:
@alextanlan Could you consider my comments here? Trying to see if we can make this even simpler.
https://broadinstitute.slack.com/archives/C03R0A8HTD2/p1663841828584159

juanseba21 (approved):

PR #1785, opened 2022-09-30
[SDK][SINGULAR] DDP-8821: Tooltip box on pre-qual page for mobile

juanseba21:
The issue was fixed by changing the tooltip box's position dynamically only for mobile view.
After the fix:

image

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/components/tooltip.component.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/components/tooltip.component.ts
@@ -1,5 +1,10 @@
-import { Component, Input, Inject } from '@angular/core';
+import { Component, Input, Inject, OnInit, OnDestroy } from '@angular/core';
 import { ConfigurationService } from '../services/configuration.service';
+import {
+    BreakpointObserver,
+    Breakpoints
+} from '@angular/cdk/layout';
+import { Subject, takeUntil } from 'rxjs';
 
 @Component({
     selector: 'ddp-tooltip',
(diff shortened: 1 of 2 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/components/tooltip.component.ts
```diff
@@ -13,10 +18,28 @@ import { ConfigurationService } from '../services/configuration.service';
...
+        this.breakpointObserver.observe([
+            Breakpoints.XSmall,
+            Breakpoints.Small,
+            Breakpoints.Medium,
+            Breakpoints.Large,
+            Breakpoints.XLarge
+        ]).pipe(takeUntil(this.destroyed))
+            .subscribe(() => {
+                this.position = this.breakpointObserver.isMatched(Breakpoints.XSmall) ? 'below' : initialPosition;
+        });
```
alextanlan, line 37: Seems like we can simplify it a bit, like:
```
this.breakpointObserver.observe([Breakpoints.XSmall])
  .pipe(takeUntil(this.destroyed))
  .subscribe((result: BreakpointState) => {
     if (result.matches) {
       this.position = 'below'; 
     }    
  });
```

Have you tried that?

juanseba21: I hadn't tried and it actually makes more sense since on a real device it is not possible to move from mobile to a larger screen. It's done.

alextanlan (approved):

PR #1800, opened 2022-10-06
[DSM] DDP-8808 Singular conditional textarea

pegahtah:
We are adding `conditionalFieldSetting` and `condition` to the `Value` class to be able to parse the new types of actions. We are then displaying  the new FieldSettings based on the value of Radio buttons or check boxes and if they meet the `condition` ( in `radio-button-directive.ts` and in  `conditional-form-data.component.ts`). 

We have changed the `createFilter(field)` function to create filters for the new conditionalFieldSettings

For radio buttons, in order to render textarea below the checked one, we are using a custom structural directive, which helps us to find which is the right radio button and also passes data to rendered textarea.


ddp-workspace/projects/ddp-dsm-ui/src/app/conditional-form-data/conditional-form-data.component.html
```diff
@@ -0,0 +1,8 @@
+<ng-container *ngIf="fieldSetting.displayType === 'TEXTAREA'">
+        <textarea maxlength="50000" placeholder="{{fieldSetting.columnDisplay}}"
```
alextanlan: just wondering whether 50000 (50 thousand) is correct value here?

pegahtah: That's going to change in another PR to come from backend as well

pegahtah: But great catch! Thanks!

GiCharkviani (dismissed):

SimoneMaiwald (dismissed):

GiCharkviani (approved):

PR #1832, opened 2022-10-20
[SINGULAR] [PEPPER-190] Playwright e2e test framework improvement

aweng98:
https://broadworkbench.atlassian.net/browse/PEPPER-190

Improvement:

- Files reorganize: Separate page-objects files from test files
- New test fixture to eliminate duplicate code in test (beforeEach)
- Introduce page interface and abstract class to eliminate duplicate code: 
   - SingularPage extends PageBase --> implements PageInterface
   - PancanPage extends PageBase --> implements PageInterface


aweng98:
Hi, I refactored some more. Added more abstractions, new test fixture and separated page-object files from the test files. Let's go over changes together tomorrow. Thank you.

playwright-e2e/pages/page-base.ts
```diff
@@ -0,0 +1,107 @@
...
+import { PageInterface } from './page-interface';
+
+export default abstract class PageBase implements PageInterface {
+  protected readonly page: Page;
+  protected readonly baseUrl: string;
+
+  protected constructor(page: Page, baseURL: string) {
+    this.page = page;
+    this.baseUrl = baseURL;
+  }
```
alextanlan, line 11: **_#minor_**

Can be shortened using [typescript constructor shorthand](https://dev.to/satansdeer/typescript-constructor-shorthand-3ibd).

playwright-e2e/authentication/auth-singular.ts
```diff
@@ -1,25 +1,10 @@
 import { Page } from '@playwright/test';
-import { clickLogin, NavSelectors } from 'tests/singular/lib/nav';
+import { clickLogin, NavSelectors } from 'pages/singular/navbar';
 
 // import * as dotenv from 'dotenv';
 // import path from 'path';
 // dotenv.config({ path: path.resolve(__dirname, '../singular/.env.singular') });
```
alextanlan: Are the commented lines needed?

playwright-e2e/pages/page-base.ts
```diff
@@ -0,0 +1,107 @@
...
+  async gotoURLPath(urlPath = ''): Promise<Response | null> {
+    if (urlPath.startsWith('https')) {
+      throw Error('Parameter urlPath is not valid.');
+    }
+    return this.page.goto(`${this.baseUrl}${urlPath}`, { waitUntil: 'domcontentloaded' });
+  }
+
+  protected async clickAndWaitForNav(locator: Locator, opts: { waitForNav?: boolean } = {}): Promise<void> {
+    const { waitForNav = false } = opts;
+    waitForNav ? await this.waitForNavAfter(() => locator.click()) : await locator.click();
```
alextanlan, line 71: It can be just 
`opts.waitForNav ? await this.waitForNavAfter(() => locator.click()) : await locator.click();`
without line 70.

playwright-e2e/pages/page-interface.ts
```diff
@@ -0,0 +1,17 @@
+import { Locator, Response } from '@playwright/test';
+
+export interface PageInterface {
+  gotoURL(url?: string): Promise<Response | null>;
```
alextanlan: Seems like the `url` is required parameter here, not optional one

playwright-e2e/pages/pancan/home/home-page.ts
```diff
@@ -0,0 +1,35 @@
...
+import { PancanPage } from 'pages/pancan/pancan-page';
+import { HomePageInterface } from 'pages/page-interface';
+
+export default class HomePage extends PancanPage implements HomePageInterface {
+  constructor(page: Page) {
+    super(page);
+  }
+
+  async waitForReady(): Promise<void> {
+    // Add additional waits here
```
alextanlan: Do we need that comment?

playwright-e2e/pages/singular/home/home-page.ts
```diff
@@ -0,0 +1,58 @@
...
+    const { waitForNav = true } = opts;
+    await this.clickAndWaitForNav(this.getSignMeUpButton(), { waitForNav });
+  }
+
+  /**
+   * Log into Singular application
+   * @param opts
+   */
+  async logIn(
+    opts: { email?: string | undefined; password?: string | undefined; waitForNavigation?: boolean } = {}
```
alextanlan: I believe we can skip `undefined` here because an optional parameter already means that it can be undefined.
so just 
`  opts: { email?: string; password?: string; waitForNavigation?: boolean } = {}`
would be enough.

playwright-e2e/utils/test-utils.ts
```diff
@@ -1,11 +1,122 @@
...
+export async function enterMailingAddress(
+  page: Page,
+  opts: {
+    fullName: string;
+    country?: string;
+    state?: string;
+    street?: string;
+    city?: string;
+    zipCode?: string;
+    telephone?: string | number;
```
alextanlan, line 56: Seems like `telephone` is a string only everywhere.

playwright-e2e/utils/test-utils.ts
```diff
@@ -1,11 +1,122 @@
...
+  await getTelephone().fill(telephone.toString());
+}
+
+/**
+ * On non-prod env, user must first enter the Site password
+ * @param page
+ * @param password
+ */
+export async function fillSitePassword(page: Page, password?: string): Promise<void> {
+  const passwd: string = typeof password === 'undefined' ? (process.env.singularSitePassword as string) : password;
```
alextanlan: It would be enough to set a default value for the password parameter, would not it?
Like 
```
export async function fillSitePassword(page: Page, password = process.env.singularSitePassword): Promise<void> {
   ...
}
```

playwright-e2e/authentication/auth-singular.ts
aweng98: No, removed now.

playwright-e2e/pages/page-base.ts
aweng98: Need line 70 to set the default value.

playwright-e2e/pages/page-interface.ts
aweng98: Yes, fixed now.

playwright-e2e/pages/pancan/home/home-page.ts
aweng98: Not absolutely needed. I thought it could helps with reminder to others if it needs any improvement. I'm removing it now to keep it clean.

playwright-e2e/pages/singular/home/home-page.ts
aweng98: Yes, updated.

playwright-e2e/utils/test-utils.ts
aweng98: For now, I just thought it may be good to have `number` type as well in case someone is not using data in `fake-user.json`.

playwright-e2e/utils/test-utils.ts
aweng98: Yes, updated.

alextanlan (approved):

playwright-e2e/pages/page-base.ts
alextanlan: It works. 
But my thought is that set `false` as a default value is excess here.

The both methods below will work in the same way
```
  protected async clickAndWaitForNav(locator: Locator, opts: { waitForNav?: boolean } = {}): Promise<void> {
    const { waitForNav = false } = opts;
    waitForNav ? await this.waitForNavAfter(() => locator.click()) : await locator.click();
  }
```

```
  protected async clickAndWaitForNav(locator: Locator, opts: { waitForNav?: boolean } = {}): Promise<void> {
    opts.waitForNav ? await this.waitForNavAfter(() => locator.click()) : await locator.click();
  }
```

aweng98: Just saw this now. I have merged PR already before seeing this. I don't have a strong opinion. Can you update this in your PR when rebasing? Sorry for not seeing your comment earlier.

PR #1834, opened 2022-10-21
Dsm no data pipe fix and test

GiCharkviani:
noData pipe adjustment to 0 case and also unit test for that pipe.

### DIFF
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/pipes/noData.pipe.spec.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/pipes/noData.pipe.spec.ts
@@ -1,7 +1,7 @@
 import {NoDataPipe} from './noData.pipe';
 import {expect} from '@angular/flex-layout/_private-utils/testing';
 
-fdescribe('noData Pipe', () => {
+describe('noData Pipe', () => {
   const pipe = new NoDataPipe();
 
   /*
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/pipes/noData.pipe.spec.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/pipes/noData.pipe.spec.ts
@@ -1,21 +1,15 @@
 import {NoDataPipe} from './noData.pipe';
 import {expect} from '@angular/flex-layout/_private-utils/testing';
 
-describe('noData Pipe', () => {
+fdescribe('noData Pipe', () => {
   const pipe = new NoDataPipe();
-
-  const nonBreakingSpace = (): string => {
-    const tempDiv = document.createElement('div');
-    tempDiv.innerHTML = String.fromCharCode(160);
-    return tempDiv.innerText;
-  };
 
   /*
     Checking if in case of particular type of values, it returns exactly the same value (In string format)
    */
   it('Should return value', () => {
     const diffData = [5, 'ddp', '5', 0, true];
-    diffData.forEach(value => expect(pipe.transform(value)).toEqual(String(value)));
+    diffData.forEach(value => expect(pipe.transform(value)).toEqual(value === 0 ? value.toString() : value));
   });
 
   /*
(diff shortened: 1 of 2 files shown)


ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/pipes/noData.pipe.spec.ts
```diff
@@ -0,0 +1,30 @@
+import {NoDataPipe} from './noData.pipe';
+import {expect} from '@angular/flex-layout/_private-utils/testing';
+
+describe('noData Pipe', () => {
+  const pipe = new NoDataPipe();
+
+  const nonBrakingSpace = (): string => {
```
alextanlan: a typo, should be `nonBreakingSpace`

GiCharkviani: Thanks 🙏

ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/pipes/noData.pipe.ts
```diff
@@ -7,7 +7,7 @@ import {Pipe, PipeTransform} from '@angular/core';
 export class NoDataPipe implements PipeTransform {
   transform(value: any): string {
     const tempElement = document.createElement('div');
-    tempElement.innerHTML = value || '&nbsp;';
+    tempElement.innerHTML = (typeof value === 'number' ? String(value) : value) || '&nbsp;';
     return tempElement.innerText;
```
alextanlan: Seems like you can use just `String.fromCharCode(160)` for the non-breaking space 
(see [link](https://stackoverflow.com/questions/5237989/how-is-a-non-breaking-space-represented-in-a-javascript-string)).

GiCharkviani: Better, Thank you

ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/pipes/noData.pipe.ts
```diff
@@ -7,7 +7,7 @@ import {Pipe, PipeTransform} from '@angular/core';
 export class NoDataPipe implements PipeTransform {
   transform(value: any): string {
     const tempElement = document.createElement('div');
-    tempElement.innerHTML = value || '&nbsp;';
+    tempElement.innerHTML = (typeof value === 'number' ? String(value) : value) || '&nbsp;';
```
alextanlan: What type the value is supposed to be?
If it is only **number** / **string** / **boolean** maybe using `value.toString()` is just enough.

GiCharkviani: it can be anything

GiCharkviani: including null and undefined

ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/pipes/noData.pipe.ts
alextanlan: If the value is just **number** / **string** / **boolean**, 
may be, it is enough to have smth like:

```
transform(value: number | string | boolean): string {
    const nonBreakingSpace = String.fromCharCode(160);
    return (value || value === 0) ? String(value) : nonBreakingSpace;
}
```

ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/pipes/noData.pipe.ts
```diff
@@ -6,8 +6,6 @@ import {Pipe, PipeTransform} from '@angular/core';
 
 export class NoDataPipe implements PipeTransform {
   transform(value: any): string {
-    const tempElement = document.createElement('div');
-    tempElement.innerHTML = value || '&nbsp;';
-    return tempElement.innerText;
+    return (value === 0 ? value.toString() : value) || String.fromCharCode(160);
```
alextanlan: Would you add ` const nonBreakingSpace = String.fromCharCode(160);`?
It is not obvious from the first sight.

GiCharkviani: Done!

alextanlan (approved):

PR #1835, opened 2022-10-21
Save button pepper198

ngigliotosBroad:
Edit: Added a save button to the field-datepicker component. Currently, it will only show when in use under the Shipping search. When clicked the button will attempt to save the date of the associated kit, on success it will show "Success!", on failure it will show "Failed!". Then, after 5 seconds, the button will change back to "Save Date". The save logic is implemented via a callback to a promise of the subscribe call. 

Also fixed a bug where the dates of every row would turn red when editing one date. Now, only the date of the associated row will turn red. This will when it is out of focus and different than the current saved date.

I tested and confirmed medical-component works as intended.


ddp-workspace/projects/ddp-dsm-ui/src/app/shipping-search/shipping-search.component.ts
```diff
@@ -127,6 +128,9 @@ export class ShippingSearchComponent implements OnInit {
     if (typeof value === 'string') {
       kitRequest[ parameterName ] = value;
       v = value;
+      if(parameterName == "collectionDate"){
```
SimoneMaiwald: the field wasn't saved at this point, so the message could say it was saved but there could still be a server error that prevented saving

ngigliotosBroad: Where does it write to the server? Here? `this.patch( patch );`

pegahtah: @ngigliotosBroad yes, we are sure it got saved at L 154

ngigliotosBroad: Alright, it now only saves on a successful subscribe call, so if the server write files the message should not show. It will now also hide the message if someone tries to change the date again after an initial change.

GiCharkviani (approved):

GiCharkviani (changes requested):
My apologies for too early approval, after testing it further, I found the following issue: 
If isSaved is set to true, the text "Date Saved Successfully!" will be displayed under each datepicker, which I suppose should not be desired behavior.

I have changed only the first field and message shows under each of them:
Screenshot 2022-10-26 at 18 24 16

ddp-workspace/projects/ddp-dsm-ui/src/app/shipping-search/shipping-search.component.html
```diff
@@ -144,10 +144,13 @@ <h3 class="Color--primary Line--Break">{{additionalMessage}}</h3>
                   (click)="receiveATKit(kitRequest)">Mark Received</button>
         </td>
         <td>
-          <app-field-datepicker (dateChanged)="valueChanged($event, 'collectionDate', kitRequest)"
+          <app-field-datepicker (input)="changingDate()"
```
GiCharkviani: I suppose you could just write so: 
<app-field-datepicker (input)="isSaved = false" >

Or to come up with a different name than the "changingDate", as it only sets the false value, for example - "hideSuccessMessage"

ngigliotosBroad: Thank you for pointing out the bug and better format for the (input) trigger. I fixed the message showing up for each row bug by adding a "saved" field to the specific kitRequest object that corresponds with the row that is currently being edited. Then I just check that specific object for the "saved" flag rather than a global one within the shipping-search.component. I assume this is okay since we do not write the entire kit request back to the server, just field updates.

GiCharkviani (commented):
Thank you for your updates.
I am not sure if mutating kitRequest object is the best option though. Anyway, as it works, we can leave it temporarily.

**Just as a reminder:** 
We were asked to write unit tests (and later e2e tests as well) for each code change. For this particular component, I think we could have one unit test since it doesn't have so many dependencies and should not take a very long time to set up TestBed.

What are your thoughts? @SimoneMaiwald @pegahtah

SimoneMaiwald:
@ngigliotosBroad why are we adding a message instead of the button? Button would be the easier and cleaner solution I think

ngigliotosBroad:
I attempted a message a first because I figured it was easier and did not realize that the table could return multiple results. I will make a button for it 👍

ngigliotosBroad:
I completely re-did the feature. It should be much cleaner now. There is a responsive button that displays both failures and successes. The button will only show for the ShippingSearch component. I am currently making sure the new additions do not break the other components.

ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.ts
```diff
@@ -83,43 +90,56 @@ export class FieldDatepickerComponent implements OnInit, OnChanges {
...
+          this.error = this.dateFormat;
         }
       }
-    } else {
-      this.colorDuringPatch = false;
-      this.error = null;
     }
   }
 
+  private async saveDate(e: Event): Promise<void>{
```
GiCharkviani: I think the `private` access modifier is not correctly used here, as far as this function is called from the template (field-datepicker.component.html, line 33). With the JiT compilation, it might not seem a problem, however, it will become problematic when we reach AOT compilation.
IDE complaining (WebStorm):
Screenshot 2022-10-28 at 13 44 16

Moreover, you could read the following discussion as well: [AOT issue](https://github.com/angular/angular/issues/11978)

My suggestion would be to use the `public` access modifier.

ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.html
```diff
@@ -29,6 +29,8 @@
           *ngIf="showTodayButton"
           (click)="setToday()"
           [disabled]="disabled">Today</button>
+  <button *ngIf="showSaveBtn" class="btn btn-xs Button--Background--Color ng-star-inserted"
```
GiCharkviani, line 32: This button is still clickable even during the saving process (while the request is in the pending state), and later as well, whether it fails or not, which, I think, is not quite a user-friendly approach. What do you think?

GiCharkviani, line 32: How do you think, would it not be nicer, to display the button in a different color, which will depend on response status, like so:
If successful - green, if failed - red, if waiting - grey (and disabled)

ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.ts
GiCharkviani: I think the `e` parameter is redundant, as it doesn't have any usage.

ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.ts
```diff
@@ -83,43 +90,56 @@ export class FieldDatepickerComponent implements OnInit, OnChanges {
...
+    this.saveButtonText = "Saving..."
+    try{
+      this.isSaved = await this.dateSaved();
+      this.saveButtonText = this.isSaved ? "Success!": "Failed";
+      this.hasDateChanged = false;
+    }
+    catch(e){
+        this.saveButtonText = "Failed";
+        console.log(e)
```
GiCharkviani: Do we need this `console.log` here?

ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.ts
```diff
@@ -83,43 +90,56 @@ export class FieldDatepickerComponent implements OnInit, OnChanges {
...
+    }
+    setTimeout(() => {
+      this.saveButtonText = "Save Date"
+    },5000);
```
GiCharkviani: What if the request took longer than 5 seconds? Should we still show "Save Date", even if the request is still pending?

ddp-workspace/projects/ddp-dsm-ui/src/app/shipping-search/shipping-search.component.ts
```diff
@@ -121,43 +123,48 @@ export class ShippingSearchComponent implements OnInit {
...
+  valueChanged(value: any, parameterName: string, kitRequest: KitRequest): void {
     if (typeof value === 'string') {
-      kitRequest[ parameterName ] = value;
-      v = value;
+      kitRequest[parameterName] = value;
     }
-    if (v != null) {
+  }
+
+  patch(patch: any, kitRequest: KitRequest): Observable<any> {
```
GiCharkviani: `kitRequest` is unused

ddp-workspace/projects/ddp-dsm-ui/src/app/shipping-search/shipping-search.component.ts
```diff
@@ -1,12 +1,14 @@
-import {Component, OnInit} from '@angular/core';
-import {ScanValue} from '../scan/scan.model';
-import {Auth} from '../services/auth.service';
-import {ComponentService} from '../services/component.service';
```
GiCharkviani, line 4: `ScanValue`, `ComponentService`, `range` and `Subscription` are unused imports

ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.ts
```diff
@@ -1,16 +1,16 @@
 import { Component, EventEmitter, Input, OnChanges, OnInit, Output, SimpleChanges } from '@angular/core';
+import { timeout } from 'rxjs';
```
GiCharkviani: `timeout` is unused import

ddp-workspace/projects/ddp-dsm-ui/src/app/shipping-search/shipping-search.component.ts
```diff
@@ -121,43 +123,48 @@ export class ShippingSearchComponent implements OnInit {
...
+  }
+
+  patch(patch: any, kitRequest: KitRequest): Observable<any> {
+    return this.dsmService.patchParticipantRecord(JSON.stringify(patch));
+  }
+
+  //Fires when user clicks "Save Date" button. Returns true
+  //if the promise succeeds and the server successfukky writes and
+  //returns and error if the server write fails.
+  saveDate(kitRequest: KitRequest): any {
```
GiCharkviani: I suppose the return type here should be `Promise<any>`

ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.ts
```diff
@@ -20,15 +20,22 @@ export class FieldDatepickerComponent implements OnInit, OnChanges {
   @Input() addCheckboxEstimated = false;
   @Input() showCalendarButton = true;
   @Input() fieldName: string;
+  @Input() showSaveBtn: boolean = false;
+  @Input() saveSucceeded: boolean = false;
+  @Input() dateSaved: any;
```
GiCharkviani: The same here -> `Promise<any>`

ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.ts
```diff
@@ -83,43 +90,56 @@ export class FieldDatepickerComponent implements OnInit, OnChanges {
...
-      this.colorDuringPatch = false;
-      this.error = null;
     }
   }
 
+  private async saveDate(e: Event): Promise<void>{
+    this.saveButtonText = "Saving..."
+    try{
+      this.isSaved = await this.dateSaved();
+      this.saveButtonText = this.isSaved ? "Success!": "Failed";
```
GiCharkviani: This line confuses me a bit. If Promise was rejected (response is an error), we should be thrown into the `catch() ` block. In which scenario `this.isSaved` will return a falsy value?

GiCharkviani (commented):
Thank you for your effort! 
Just curious, why Promise-based flow, instead of observables and observers?

GiCharkviani (commented):
Will we have unit tests for that?

ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.html
SimoneMaiwald: I think that would be too much and we are not doing that anywhere else!
But I agree with while it is saving. we should disable the button till it is done

ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.ts
SimoneMaiwald: no console log! error should be in backend gcp log

ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.html
ngigliotosBroad: I think it is fine if it is clickable after failing, because the user may want to try again. However, I can disable it during the pending state.

ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.ts
ngigliotosBroad: I will remove it

ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.ts
ngigliotosBroad: Gotcha, I thought I should log the errors to the console. I will remove it.

ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.ts
ngigliotosBroad: That was left over because I initially had the promise resolving regardless of the error. I can change it.

ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.ts
ngigliotosBroad: I will put it inside a finally so the request completes before the timeout begins

ddp-workspace/projects/ddp-dsm-ui/src/app/shipping-search/shipping-search.component.ts
ngigliotosBroad: Trying to make this Promise<any> gives an error

ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.ts
ngigliotosBroad: Same here, trying to make this promise<any> gives an error

ngigliotosBroad:
I went with promises because it is what I felt most comfortable implementing and what seemed like the cleanest solution for me. I fixed all of the issues you brought up @GiCharkviani but the return for saveDate() and type for dateSaved are still any as trying to switch to Promise any gives an error. Thank you very much for all the improvements though, this feature has come along way from a week ago. Please let me know if you see any other issues.

ddp-workspace/projects/ddp-dsm-ui/src/app/shipping-search/shipping-search.component.ts
GiCharkviani: My apologies, I should have been more specific. Have a look at this code, that should give you a hint:
`saveDate(kitRequest: KitRequest): () => Promise<boolean> {
    return async (): Promise<boolean> => {... }}`

ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.ts
GiCharkviani: Here goes the same: `  @Input() dateSaved: () => Promise<boolean>`

GiCharkviani (commented):
@ngigliotosBroad Thank you for your improvements. I have updated my comments about the return type. The rest of the review I will do on Monday. Have a good weekend!

ngigliotosBroad:
Ah I see. I feel dumb for not understanding what you meant earlier. I added those type changes. Have a good weekend too!

GiCharkviani (changes requested):
Clicking on the calendar icon next to the field that is already filled with some data, causes the infinite loop.
**Hint:**
Up until now, what I have investigated, this issue could be solved by:
- simply just setting `datePicker` field to falsy value;
- Not returning an async function that returns a Promise (not using Promise-based flow) - probably this can be the main reason;

It required deeper investigation though.

_(I have inserted `console.log` in the "ngOnChanges" lifecycle hook):_

https://user-images.githubusercontent.com/77500504/198976700-8ce2f79e-7072-4e3e-b8e7-641dee4b120f.mov

ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.ts
```diff
@@ -84,42 +91,62 @@ export class FieldDatepickerComponent implements OnInit, OnChanges {
...
-            }
-          } else if (tmp === Utils.DATE_PARTIAL) {
-            this.error = null;
-            this.emitDate(Utils.getPartialDateFormatted(this._dateString, this.dateFormat));
+    if (this._dateString === '') {
+      this.emitDate('');
+    } else {
+      if (this._dateString !== 'N/A' && this._dateString !== 'Not Found') {
+        const tmp = Utils.parseDate(this._dateString, this.dateFormat, this.allowUnknownDay);
+        // console.log(tmp);
```
GiCharkviani, line 106: I think we can remove this comment

ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.html
```diff
@@ -1,11 +1,11 @@
 <span class="Display--Inline-Block">
   <b *ngIf="fieldName != null">{{fieldName}}</b>
   <mat-form-field class="DatePicker--Input">
-    <input matInput maxlength="10" [ngClass]="{'Input': true, 'Color--Field--Patched': colorDuringPatch}"
+    <input matInput maxlength="10" [ngClass]="{'Input': true, 'Color--Field--Patched': hasDateChanged}"
            placeholder="{{getDateFormatPlaceholder(dateFormat)}}"
```
GiCharkviani: I would have a pure pipe (lowerecase) here instead of a function call.
Example:
`placeholder="{{dateFormat | lowercase}}"`

ngigliotosBroad:
@GiCharkviani @SimoneMaiwald The bug is fixed now and I switched to using Observables. Please let me know if there are any other issues. Thanks!

ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.ts
```diff
@@ -1,16 +1,16 @@
 import { Component, EventEmitter, Input, OnChanges, OnInit, Output, SimpleChanges } from '@angular/core';
 import { Utils } from '../utils/utils';
 import { EstimatedDate } from './field-datepicker.model';
+import { defer, Observable, throwError} from 'rxjs';
```
GiCharkviani: `throwError` and `defer` are unused

ddp-workspace/projects/ddp-dsm-ui/src/app/shipping-search/shipping-search.component.ts
```diff
@@ -121,43 +120,39 @@ export class ShippingSearchComponent implements OnInit {
...
-  }
+saveCompleted(): void{
+  this.currentPatchField = null;
 }
+
+  //Fires when user clicks "Save Date" button. Returns true
+  //if the promise succeeds and the server successfukky writes and
+  //returns and error if the server write fails.
+  saveDate(kitRequest: KitRequest): Observable<boolean> { 
+      return defer(() => {
```
GiCharkviani: **# Optional**
Why did we need `defer()` here? 
As far as I know, it's used mostly in 2 scenarios:
- If we want to execute code on subscribe that will be eagerly executed, like Promise.
- If we want to execute an Observable call, but with some side effects, that should happen only when subscribed (This probably can be handled in `pipe()` as well)

Please see the [discussion](https://stackoverflow.com/questions/38764578/rxjs-understanding-defer).

Maybe we could simply return the `this.patch()`, which itself returns an observable that won't be executed eagerly. 

P.S. Maybe we've got a scenario, where it's necessary to use `defer()` and I didn't find it yet, please let me know.

ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.ts
```diff
@@ -84,88 +97,109 @@ export class FieldDatepickerComponent implements OnInit, OnChanges {
...
+    this.currentlySaving = true;
+      this.dateSaved.subscribe({
+        next: data => {
+          this.saveButtonText = "Success!";
+          this.hasDateChanged = false;
+        },
+        error: err => {
+          this.saveButtonText = "Failed";
+        },
+          }).add(() => {
```
GiCharkviani: **# Optional**
`.add()` function on subscription is called then when we want to add more subscriptions to it, in order to easily unsubscribe them later on. 

I'm just wondering, could we not wrap this block of code in `finalize()` [operator](https://rxjs.dev/api/operators/finalize)?
Like so:
```
 this.dateSaved
        .pipe(finalize(() => {
          this.saveCompleted.emit();
          setTimeout(() => {
            this.saveButtonText = "Save Date";
            this.currentlySaving = false;
          },5000);
        }))
        .subscribe({ ... })
```

Please read this [discussion](https://github.com/ReactiveX/rxjs/issues/2769).

ddp-workspace/projects/ddp-dsm-ui/src/app/shipping-search/shipping-search.component.ts
```diff
@@ -121,43 +120,39 @@ export class ShippingSearchComponent implements OnInit {
...
-      next: data => {
-        this.currentPatchField = null;
-      },
-      error: err => {
-        if (err._body === Auth.AUTHENTICATION_ERROR) {
-          this.auth.logout();
-        }
-      }
-    } );
+//Triggers when user finishes typing
```
GiCharkviani, line 127: Is this comment still valid?

ddp-workspace/projects/ddp-dsm-ui/src/app/shipping-search/shipping-search.component.ts
```diff
@@ -121,43 +120,39 @@ export class ShippingSearchComponent implements OnInit {
...
 
-  isPatchedCurrently( field: string ): boolean {
-    return this.currentPatchField === field;
-  }
+saveCompleted(): void{
+  this.currentPatchField = null;
 }
+
+  //Fires when user clicks "Save Date" button. Returns true
+  //if the promise succeeds and the server successfukky writes and
```
GiCharkviani: Is this comment still valid?

ddp-workspace/projects/ddp-dsm-ui/src/app/utils/utils.ts
```diff
@@ -491,7 +491,7 @@ export class Utils {
     }
   }
 
-  public static getPartialDateFormatted(dateString: string, format: string): string {
+  public static getPartialDateFormatted(dateString: string | undefined, format: string): string {
```
GiCharkviani: A bit confusing. If `dateString` is undefined and we don't check for that, how will we call `.split()` method on it? (line 498).

ddp-workspace/projects/ddp-dsm-ui/src/app/shipping-search/shipping-search.component.ts
ngigliotosBroad: Yes, I believe that is still valid. The valueChanged function is still called on dateChanged, which is called when the user finishes typing.

ddp-workspace/projects/ddp-dsm-ui/src/app/utils/utils.ts
ngigliotosBroad: Sorry, this was a remnant of a wild goose chase I was on yesterday. It will be removed shortly.

ddp-workspace/projects/ddp-dsm-ui/src/app/shipping-search/shipping-search.component.ts
ngigliotosBroad: This is left over from an attempt to solve the infinite loop issue, I can remove it as it is not needed.

ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.ts
```diff
@@ -84,88 +98,108 @@ export class FieldDatepickerComponent implements OnInit, OnChanges {
...
+        this.dateChanged.emit(dateString);
+      }
+    }
+  
+    estimatedChanged(): void {
+      this._dateString = Utils.getDateFormatted(this.datePicker, this.dateFormat); // show user date in the format from the user settings
+      this.emitDate(Utils.getDateFormatted(this.datePicker, Utils.DATE_STRING));
+    }
+  
+    getDateFormatPlaceholder(placeholder: string): string {
```
GiCharkviani, line 202: This function is not called anymore

ddp-workspace/projects/ddp-dsm-ui/src/app/shipping-search/shipping-search.component.ts
```diff
@@ -1,12 +1,11 @@
...
-import {Statics} from '../utils/statics';
-import {DSMService} from '../services/dsm.service';
-import {KitRequest} from '../shipping/shipping.model';
-import {RoleService} from '../services/role.service';
+import { PatchUtil } from '../utils/patch.model';
+import { Statics } from '../utils/statics';
+import { DSMService } from '../services/dsm.service';
+import { KitRequest } from '../shipping/shipping.model';
+import { RoleService } from '../services/role.service';
+import { Observable, defer } from 'rxjs';
```
GiCharkviani, line 8: defer is unused

GiCharkviani (dismissed):
LGTM

GiCharkviani (approved):

ddp-workspace/projects/ddp-dsm-ui/src/app/field-datepicker/field-datepicker.component.html
amimahd: > @@ -1,11 +1,11 @@ {{fieldName}} - 
{{getDateFormatPlaceholder(dateFormat)}}
 - if (this.datePicker != null) { - this.emitDate(Utils.getDateFormatted(this.datePicker, Utils.DATE_STRING)); - this.colorDuringPatch = false; - this.error = null; - } else { - this.error = this.dateFormat; - } - } else if (tmp === Utils.DATE_PARTIAL) { - this.error = null; - this.emitDate(Utils.getPartialDateFormatted(this._dateString, this.dateFormat)); + if (this._dateString === '') { + this.emitDate(''); + } else { + if (this._dateString !== 'N/A' && this._dateString !== 'Not Found') { + const tmp = Utils.parseDate(this._dateString, this.dateFormat, this.allowUnknownDay); + // console.log

PR #1842, opened 2022-10-25
[DSM] PEPPER-57 file upload time in participant list fixed

pegahtah:
* I moved the method to Utils and both files component and participant list are calling it now from there.
* I also fixed the alignment issues with file names and upload times
Screen Shot 2022-10-25 at 2 43 56 PM

### DIFF
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/file-download/file-download.component.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/file-download/file-download.component.ts
@@ -1,4 +1,3 @@
-import {DatePipe} from '@angular/common';
 import {Component, Input, Output, EventEmitter} from '@angular/core';
 import {ESFile} from '../participant-list/models/file.model';
 import {Participant} from '../participant-list/participant-list.model';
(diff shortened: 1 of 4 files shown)


ddp-workspace/projects/ddp-dsm-ui/src/app/file-download/file-download.component.ts
```diff
@@ -21,9 +21,8 @@ export class FileDownloadComponent {
     this.downloadFileEvent.emit(file);
   }
 
-  getNiceDateFormat( uploadedAt: string ): string {
-    return new DatePipe( 'en-US' ).transform( uploadedAt, Utils.DATE_STRING_IN_CVS_WITH_TIME );
+  getUtilStatic(): typeof Utils {
```
alextanlan: **_#optional_**

I'd suggest having a property for that purpose, like
```
import {Utils} from '../utils/utils';
...
export class FileDownloadComponent {
  Utils = Utils;
}
```
and use it in the template like
`<td>{{Utils.getNiceDateFormat(file.scannedAt)}}</td>`

alextanlan:
**_#minor_**

The first line in `file-download.component.ts` is an unused import of `DatePipe`.

ddp-workspace/projects/ddp-dsm-ui/src/app/file-download/file-download.component.ts
pegahtah: Thanks! I fixed it

alextanlan (approved):

PR #1845, opened 2022-10-27
[DSM] PEPPER-17  Saved Filters Improvements

pegahtah:
I changed the html and css for the text of a filter name to look like a link when hovered and also apply the filter when clicked.

Also when clicking on the reload with default filter I set the selected filter to blank, therefore the previous selected filter will be set to black and if there is a default filter set in the user settings that one will turn green
Screen Shot 2022-10-27 at 10 11 23 AM


ddp-workspace/projects/ddp-dsm-ui/src/app/tissue-list/tissue-list.component.html
```diff
@@ -129,31 +129,35 @@ <h3 class="Color--primary Line--Break">{{additionalMessage}}</h3>
...
-          </span>
-                  <button class="float-right" color="warn"
-                          mat-icon-button (click)="deleteView(savedFilter)">
-                    <mat-icon>
-                      <i class="fas fa-trash-alt fa-sm"></i>
-                    </mat-icon>
+                  <button class="float-right"
+                          mat-icon-button [color]="selectedFilterName === savedFilter.filterName ? 'accent' :'black'"
+                          (click)="selectFilter(savedFilter)">
+                     <span style="font-size: 15px;">
```
alextanlan: I believe you can get rid of that span and set the style for the button in the proper scss file (here and below).

ddp-workspace/projects/ddp-dsm-ui/src/app/tissue-list/tissue-list.component.html
```diff
@@ -167,17 +171,20 @@ <h3 class="Color--primary Line--Break">{{additionalMessage}}</h3>
...
-          <span style="font-size: 15px;">
-            <i class="fas fa-filter fa-sm"></i>
-          </span>
+                      <span style="font-size: 15px;">
+                        <i class="fas fa-filter fa-sm"></i>
+                      </span>
                   </button>
                 </td>
+                <td>
+                  <span (click)="selectFilter(savedFilter)" class="clickable" style="font-size: 15px; padding-right: 30px;">
```
alextanlan: it would be nice to avoid inline styles.

alextanlan (approved):

PR #1855, opened 2022-11-07
[DSM]- PEPPER-211 - Date range for statistics

GiCharkviani:
[PEPPER-211
](https://broadworkbench.atlassian.net/browse/PEPPER-211)

### Date range picker for filtering statistics dashboard's data

**Route**: _.../studyName/statisticsDashboard_

**@Inputs:**
- initial start and end dates
- loading state for disabling fields

**@Output:** emitted value change after passing validation checks

**Validations:** 
- Required and Invalid

The error message is shown accordingly

**:triangular_ruler: Includes Unit tests**

**Note:** _backend part is not prepared, therefore dashboard-statistics.component and service will be modified later._

https://user-images.githubusercontent.com/77500504/200320673-45e58155-5502-4ca2-98c1-440d4c4efa54.mov


ddp-workspace/angular.json
```diff
@@ -2953,6 +2953,11 @@
               "projects/ddp-dsm-ui/src/assets/dsm-theme.scss",
               "projects/ddp-dsm-ui/src/styles.scss"
             ],
+            "stylePreprocessorOptions": {
+                "includePaths": [
+                    "projects/ddp-dsm-ui/src/styles"
```
GiCharkviani, line 2958: I have added this for avoiding very long relative paths in SCSS

ddp-workspace/projects/ddp-dsm-ui/src/app/dashboard-statistics/dashboard-statistics.component.ts
```diff
@@ -19,25 +25,44 @@ export class DashboardStatisticsComponent implements OnInit {
...
+
+  private initData(): void {
     this.hasRequiredRole = this.roleService.allowedToViewEELData();
     this.Charts = this.dashboardStatisticsService.ChartFactory()
       .pipe(catchError(this.catchErrorAndReturnArray.bind(this)), finalize(() => this.loading = false));
     this.Counts = this.dashboardStatisticsService.Counts;
   }
 
+  public dateChanged(date: DateRangeModel): void {
+    console.log(date, 'DATE - dashboard.statistics');
```
GiCharkviani: Just for testing purposes

ddp-workspace/projects/ddp-dsm-ui/src/app/dashboard-statistics/dashboard-statistics.component.ts
```diff
@@ -19,25 +25,44 @@ export class DashboardStatisticsComponent implements OnInit {
...
       responsive: true,
       displaylogo: false
     };
   }
 
-  private catchErrorAndReturnArray(error: HttpErrorResponse): [] {
-    this.errorMessage.next(error);
-    return [];
+  public get activeDates(): DateRangeModel {
+    return {startDate: new Date(), endDate: new Date()};
```
GiCharkviani: Just for testing purposes, as backend is not ready yet

ddp-workspace/projects/ddp-dsm-ui/src/styles/fonts.scss
```diff
@@ -21,3 +21,15 @@ $fonts_directory: '../assets/fonts/Montserrat-';
   font-family: Montserrat-Bold;
   src: url($fonts_directory + 'Bold.ttf');
 }
+
+@mixin MontserratFontFamily($font_name: "regular") {
```
alextanlan: Seems like it is never used.

GiCharkviani: Yes, I used it for a while but then I just left it so. 
For sure, we will use it. 
SCSS won't call this code as it's mixin function, therefore don't think this block of code will ever be included in the bundle. 
I mean, we can live it for later uses. 
Please, correct me if I'm wrong

GiCharkviani: Btw, it's just a solution for getting font-family in an easier way, that's why I think we will use it anywhere we need those fonts.

alextanlan: It is definitely up to you.
I just like to avoid any unused code as other people not to guess why it was added (if they face it in the future).

GiCharkviani: I see. 
There is some chance I can use it even in this PR after integrating the updated API. 
If I didn't use it, I'll remove it, though. NP.

ddp-workspace/projects/ddp-dsm-ui/src/app/dashboard-statistics/components/date-range/date-range.component.ts
```diff
@@ -0,0 +1,137 @@
...
+  private readonly destroyed$: Subject<void> = new Subject<void>();
+
+  constructor(private datePipe: DatePipe) {
+  }
+
+  @Input('activeDates') activeDatesOnInit: DateRangeModel;
+
+  @Input('disabled') set disabledState(isDisabled: boolean) {
+    typeof isDisabled === 'boolean' && this.disableOrEnable(isDisabled);
+  }
```
alextanlan, line 45: Do you really need to check if `typeof isDisabled === 'boolean'` ?

alextanlan: I believe 
```
 @Input('disabled') set disabledState(isDisabled: boolean) {
     const options = {emitEvent: false};
     isDisabled? this.dateRangeForm.disable(options) : this.dateRangeForm.enable(options);   
  }
```

would be enough here

ddp-workspace/projects/ddp-dsm-ui/src/app/dashboard-statistics/components/date-range/date-range.component.ts
```diff
@@ -0,0 +1,137 @@
...
+  selector: 'app-date-range',
+  templateUrl: 'date-range.component.html',
+  styleUrls: ['date-range.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class DateRangeComponent implements OnInit, OnDestroy {
+  /**
+   * @MAIN Main Date Range Object
+   */
+  public readonly dateRangeForm: FormGroup = this.generateFormGroup;
```
alextanlan: **#optional**

I'd rather have the initiation here, like
`public readonly dateRangeForm: FormGroup = new FormGroup(...)`

It seems a bit more readable (for me).

ddp-workspace/projects/ddp-dsm-ui/src/app/dashboard-statistics/components/date-range/date-range.component.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  private readonly destroyed$: Subject<void> = new Subject<void>();
+
+  constructor(private datePipe: DatePipe) {
+  }
+
+  @Input('activeDates') activeDatesOnInit: DateRangeModel;
```
alextanlan: may be 'initDates'?

like 
`@Input() initDates: DateRangeModel;`

ddp-workspace/projects/ddp-dsm-ui/src/app/dashboard-statistics/components/date-range/date-range.component.ts
GiCharkviani: Oh, a better. Thanks. Changed!

ddp-workspace/projects/ddp-dsm-ui/src/app/dashboard-statistics/components/date-range/date-range.component.ts
GiCharkviani: Agree. moved

ddp-workspace/projects/ddp-dsm-ui/src/app/dashboard-statistics/components/date-range/date-range.component.ts
GiCharkviani: Fine. Pushed!

ddp-workspace/projects/ddp-dsm-ui/src/app/dashboard-statistics/components/date-range/date-range.component.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  private get isActiveDateOnInitRangeObject(): boolean {
+    const requiredProperties: string[] = ['startDate', 'endDate'];
+    /**
+     * Here we are using .some() method on purpose
+     */
+    return this.activeDatesOnInit
+      && this.activeDatesOnInit instanceof Object
+      && requiredProperties.some((prop: string) => this.activeDatesOnInit.hasOwnProperty(prop));
```
alextanlan: Are the checks/that `isActiveDateOnInitRangeObject` method needed? 
`DateRangeModel` type should have both fields `startDate` / `endDate`

ddp-workspace/projects/ddp-dsm-ui/src/app/dashboard-statistics/components/date-range/date-range.component.ts
```diff
@@ -0,0 +1,137 @@
...
+  private initDateRangeForm(): void {
+    if(this.isActiveDateOnInitRangeObject) {
+      this.dateRangeForm.patchValue(this.activeDatesOnInit, {emitEvent: false});
+    }
+  }
+
+  private listenToValueChangesAndEmit(): void {
+    /**
+     * Here is used auditTime operator, due to known bug in Angular material:
+     * Discussion: https://github.com/angular/components/issues/19776
```
alextanlan, line 72: Good catch!

ddp-workspace/projects/ddp-dsm-ui/src/app/dashboard-statistics/components/date-range/date-range.component.ts
GiCharkviani: Probably will not, but I still thought, what if this initial data later we want from API, then we might get the wrong value there, or one part missing, as TS won't be able to check at runtime. 
I can remove it, though, as for now, we don't get this info from the backend. Later will think about what approach to take.

P.S. In the case of only one value (either startDate or endDate) is missing, we still display the existing one, that's one I used .some() operator

GiCharkviani: Done

ddp-workspace/projects/ddp-dsm-ui/src/app/dashboard-statistics/components/date-range/date-range.component.ts
GiCharkviani: Thanks

ddp-workspace/projects/ddp-dsm-ui/src/app/dashboard-statistics/components/date-range/date-range.component.ts
```diff
@@ -0,0 +1,137 @@
...
+  @Output() dateChanged = new EventEmitter<DateRangeModel>();
+
+
+  ngOnInit(): void {
+    this.initDateRangeForm();
+    this.listenToValueChangesAndEmit();
+  }
+
+  ngOnDestroy(): void {
+    this.destroyed$.next();
```
alextanlan, line 56: suggest adding
`this.destroyed$.complete();`

ddp-workspace/projects/ddp-dsm-ui/src/app/dashboard-statistics/components/date-range/date-range.component.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  public getEntries(object: object): [string, any][] {
+    return Object.entries(object);
+  }
```
alextanlan: The method does not make sense, imo.
Just use `Object.entries(your_object)`

ddp-workspace/projects/ddp-dsm-ui/src/app/dashboard-statistics/components/date-range/date-range.component.ts
GiCharkviani: Oh, ofc, missed that. Fixed!

ddp-workspace/projects/ddp-dsm-ui/src/app/dashboard-statistics/components/date-range/date-range.component.ts
GiCharkviani: How to call that in the template?

GiCharkviani: of course, one way was to create other pipe, but I didn't go further

ddp-workspace/projects/ddp-dsm-ui/src/app/dashboard-statistics/components/date-range/date-range.component.ts
alextanlan: If some wrong data is supposed to be from back-end side it is worth using some mapper there
 to align the received data with `DateRangeModel` type.

ddp-workspace/projects/ddp-dsm-ui/src/app/dashboard-statistics/components/date-range/date-range.component.ts
alextanlan: e.g. use [KeyValue pipe](https://angular.io/api/common/KeyValuePipe) in a template.
But in this particular case I guess you should simplify the error messages array which is dispayed.

GiCharkviani: yes, I tried this pipe as well, but in our structure didn't work properly, and didn't try to adjust it further. I was thinking about this as well, also how critical this can be but, of course, I will change.
Let's leave it open. I will let you know when I update the error messages approach.

ddp-workspace/projects/ddp-dsm-ui/src/app/dashboard-statistics/components/date-range/date-range.component.ts
GiCharkviani: Yes, something like that would be another workaround after integrating API.

GiCharkviani: Yes, something like that would be another workaround after integrating API.

ddp-workspace/projects/ddp-dsm-ui/src/app/dashboard-statistics/components/date-range/date-range.component.ts
GiCharkviani: adjusted!

SimoneMaiwald (approved):

pegahtah (approved):

PR #1861, opened 2022-11-09
DSM-PEPPER-114-improve-messaging

GiCharkviani:
[PEPPER-114](https://broadworkbench.atlassian.net/browse/PEPPER-114)

Up until now, previous logic was checking only for the country - USA & state - NY.
**Added extra check:** 
If a country is not a USA, but it is a Canada, display that error message.

### DIFF
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-page/participant-page.component.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-page/participant-page.component.ts
@@ -1629,6 +1629,8 @@ export class ParticipantPageComponent implements OnInit, OnDestroy, AfterViewChe
                 canBeSequencedBasedOnLocation = true;
               }
             }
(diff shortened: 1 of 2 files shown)


ddp-workspace/projects/ddp-dsm-ui/src/app/participant-page/participant-page.component.ts
```diff
@@ -1629,6 +1629,8 @@ export class ParticipantPageComponent implements OnInit, OnDestroy, AfterViewChe
                 canBeSequencedBasedOnLocation = true;
               }
             }
+          } else if(countryQuestion.answer.indexOf(this.SELF_COUNTRY_CA) > -1) {
```
alextanlan: **_#optional_**
`Array.prototype.includes()` can be used here to simplify. It returns boolean value.

alextanlan (dismissed):

SimoneMaiwald (approved):

alextanlan (approved):

PR #1868, opened 2022-11-10
DSM-PEPPER-296-RGP-resetting-columns-fixed

GiCharkviani:
[PEPPER-296](https://broadworkbench.atlassian.net/browse/PEPPER-296)

This error probably appears only for RGP study and the reason is the `data` field in `selectedColumns` object, so-called - "Participant Columns", which is empty by default:
Whenever we select something that doesn't include in the above-mentioned column and we don't specify `viewFilter`, code execution flow drops into `else` block (line 1021), where it checks for the length of the `data` field, which is 0, therefore all selected columns are reset back to the default selection.

**Solution:**
I have inserted a check for resetting selected columns to an empty array. It will only happen if the selected study is not **RGP**.
Not sure yet If it can break anything, didn't spot any bugs up until now.

**Another solution would be:**
If the `data` field was not empty by default.

**After fix:**

https://user-images.githubusercontent.com/77500504/201032349-679fae14-3a41-4b0e-b512-135984c07c04.mov

### DIFF
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/participant-list.component.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/participant-list.component.ts
@@ -1020,11 +1020,14 @@ export class ParticipantListComponent implements OnInit {
               }
             } else {
               // if selected columns are not set, set to default columns
+              const selectedStudy = localStorage.getItem(ComponentService.MENU_SELECTED_REALM);
               if ((this.selectedColumns['data'] && this.selectedColumns['data'].length === 0)
                 || (!this.selectedColumns['data'] && this.isSelectedColumnsNotEmpty())) {
-                this.dataSources.forEach((value: string, key: string) => {
-                  this.selectedColumns[key] = [];
-                });
+                if(selectedStudy !== 'RGP') {
+                  this.dataSources.forEach((value: string, key: string) => {
+                    this.selectedColumns[key] = [];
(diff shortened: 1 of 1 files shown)


ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/participant-list.component.ts
```diff
@@ -1020,11 +1020,14 @@ export class ParticipantListComponent implements OnInit {
               }
             } else {
               // if selected columns are not set, set to default columns
+              const selectedStudy = localStorage.getItem(ComponentService.MENU_SELECTED_REALM);
               if ((this.selectedColumns['data'] && this.selectedColumns['data'].length === 0)
                 || (!this.selectedColumns['data'] && this.isSelectedColumnsNotEmpty())) {
-                this.dataSources.forEach((value: string, key: string) => {
-                  this.selectedColumns[key] = [];
-                });
+                if(selectedStudy !== 'RGP') {
```
alextanlan, line 1027: I am not aware much about the functionality.
But that fix seems like a hack.
The question is why **RGP** behaves differently in comparison with other activities?
In my opinion the bug cause is something else.

GiCharkviani: I have written why it behaves differently in the description. I am still working on this PR. I spotted another bug caused by this solution

PR #1875, opened 2022-11-14
[DSS] [ATCP] PEPPER-25: Update profiles(languages) for dependent participants

alextanlan:
The issue is in [PEPPER-25](https://broadworkbench.atlassian.net/browse/PEPPER-25) ticket.
 
 Fix - update preferred languages for all dependents/participants in the same time with the main user.


ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/participantProfileServiceAgent.service.ts
```diff
@@ -0,0 +1,37 @@
...
+    public updateParticipantProfiles(participants: Array<{ guid: string; profile: any}>): Observable<any>[] {
+        return participants.map(participant => {
+            const {guid, profile} = participant;
+            return this.updateParticipantProfile(guid, profile);
+        });
+    }
+
+    public updateParticipantProfile(guid, profile: UserProfile): Observable<any> {
+        const profileChanges: object = {};
+        for (const key of Object.keys(profile)) {
```
GiCharkviani, line 29: **# Optional:**
This "kind of" imperative approach looks fine in terms of space and time complexity.
I was just trying to find other workarounds, which still worth taking into consideration: 

**The long forEach() approach:**
```
 Object
      .keys(profile)
      .filter(key => profile[key])
      .forEach((key: string) => profileChanges[key] = profile[key]);
```

**The short forEach() approach (thanks to operator precedence):**
```
        Object
            .keys(profile)
            .forEach((key: string) => profile[key] && (profileChanges[key] = profile[key]));
```

**The long reduce() approach:**
```
const profileChanges: object = Object
      .keys(profile)
      .filter(key => profile[key])
      .reduce((accum: object, key: string) => ({...accum, [key]: profile[key]}), {});
```

**The short reduce() approach (thanks to operator precedence):**
```
const profileChanges: object = Object
      .keys(profile)
      .reduce((accum: object, key: string) => profile[key] && ({...accum, [key]: profile[key]}), {});
```

**As far as I tested, these approaches worked just fine. If you think there might be a scenario, in which any of these fails, please let me know.**

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/participantProfileServiceAgent.service.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    public updateParticipantProfiles(participants: Array<{ guid: string; profile: any}>): Observable<any>[] {
+        return participants.map(participant => {
+            const {guid, profile} = participant;
+            return this.updateParticipantProfile(guid, profile);
+        });
+    }
+
+    public updateParticipantProfile(guid, profile: UserProfile): Observable<any> {
```
GiCharkviani: `updateParticipantProfile` is only used in this service. 
Probably it's better to make it `private`.

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/participantProfileServiceAgent.service.ts
GiCharkviani: You could be more specific with the `object` type, like using `Partial` type for `UserProfile`

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/participantProfileServiceAgent.service.ts
```diff
@@ -0,0 +1,37 @@
...
+    constructor(
+        session: SessionMementoService,
+        @Inject('ddp.config') configuration: ConfigurationService,
+        http: HttpClient,
+        logger: LoggingService) {
+        super(session, configuration, http, logger, null);
+    }
+
+    public updateParticipantProfiles(participants: Array<{ guid: string; profile: any}>): Observable<any>[] {
+        return participants.map(participant => {
```
GiCharkviani, line 21: I think there should be `UserProfile`, instead of `any` type like so:
```
 public updateParticipantProfiles(participants: Array<{ guid: string; profile: UserProfile}>): Observable<any>[] {...}
```

GiCharkviani (dismissed):
LGTM

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/participantProfileServiceAgent.service.ts
alextanlan: Agree, it is possible. 
However, I can't see any benefit of it for that particular case.
So, let's leave it as is, shall we?

GiCharkviani: Oki, NP

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/participantProfileServiceAgent.service.ts
alextanlan: Fixed

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/participantProfileServiceAgent.service.ts
alextanlan: Fixed.

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/participantProfileServiceAgent.service.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+
+        forkJoin(
+            service.updateParticipantProfiles(participants)
+        ).subscribe((result: any[]) => {
+            expect(result.length).toEqual(2);
+            done();
+        });
+
+        const req1 = httpTestingController.expectOne(`${backendUrl}/pepper/v1/user/${participantGuid1}/profile`);
+        expect(req1.request.method).toBe('PATCH');
```
GiCharkviani, line 69: **# Optional**
I would rather have one assertion per test.

> “One assertion per test” is a wise rule to keep in mind, because it helps you have tests that fail for a specific reason, and drives you to focus on a specific behavior at a time.

[One assertion per test](https://pierodibello.medium.com/one-assertion-per-test-732cc2a7d3d#:~:text=%E2%80%9COne%20assertion%20per%20test%E2%80%9D%20is,specific%20behavior%20at%20a%20time.)
[Angular Unite Testing Best Practices](https://www.youtube.com/watch?v=7dcSpinbgqM&t=1323s)

Afterward, you could wrap some parts of logic in the `before` function block that will be shared across the tests.

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/participantProfileServiceAgent.service.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+        req1.flush({});
+
+        const req2 = httpTestingController.expectOne(`${backendUrl}/pepper/v1/user/${participantGuid2}/profile`);
+        expect(req2.request.method).toBe('PATCH');
+        expect(req2.request.body).toEqual(JSON.stringify({preferredLanguage: 'en'}));
+        req2.flush({});
+    });
+});
+
+
```
GiCharkviani, line 80: Seems like we don't test the error case. 
Maybe it would be more efficient to have at least one test case for error as well.
What do you think?

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/participantProfileServiceAgent.service.spec.ts
GiCharkviani: There is another [opinion](https://stackoverflow.blog/2022/11/03/multiple-assertions-per-test-are-fine/), though. :D . 
As this case is specific due to HttpClient, you can stick to whichever you like

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/participantProfileServiceAgent.service.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+    beforeEach(() => {
+        config = new ConfigurationService();
+        config.backendUrl = backendUrl;
+        config.studyGuid = 'ATCP';
+        const loggingServiceSpy: jasmine.SpyObj<LoggingService> = jasmine.createSpyObj('LoggingService', ['logException']);
+        const sessionSpy = new SessionMementoService({} as TranslateService, config);
+        spyOnProperty(sessionSpy, 'sessionObservable').and.returnValue(of({ participantGuid: 'USERGUID' }));
+
+        TestBed.configureTestingModule({
+            imports: [HttpClientTestingModule]
```
GiCharkviani, line 24: **# Optional**
We could also have TestBed setup this way:
```
        TestBed.configureTestingModule({
            imports: [HttpClientTestingModule],
            providers: [
                ParticipantProfileServiceAgent,
                {
                    provide: SessionMementoService,
                    useValue: sessionSpy
                },
                {
                    provide: 'ddp.config',
                    useValue: config
                },
                {
                    provide: LoggingService,
                    useValue: loggingServiceSpy
                }
            ]
        });
        httpTestingController = TestBed.inject(HttpTestingController);
        service = TestBed.inject(ParticipantProfileServiceAgent);
```

In order to avoid the unnecessary import of `HttpClient`, as it's already provided by `HttpClientTestingModule`.

It's totally up to you, though.

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/participantProfileServiceAgent.service.spec.ts
GiCharkviani: This [article](https://testing-angular.com/testing-services/#alternatives-for-finding-pending-requests) could be helpful for avoiding multiple assertions in this particular case by inserting some configuration into the `expectOne` method

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/participantProfileServiceAgent.service.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+            expect(result.length).toEqual(2);
+            done();
+        });
+
+        const req1 = httpTestingController.expectOne(`${backendUrl}/pepper/v1/user/${participantGuid1}/profile`);
+        expect(req1.request.method).toBe('PATCH');
+        expect(req1.request.body).toEqual(JSON.stringify({preferredLanguage: 'en'}));
+        req1.flush({});
+
+        const req2 = httpTestingController.expectOne(`${backendUrl}/pepper/v1/user/${participantGuid2}/profile`);
```
GiCharkviani, line 73: **#Optional**
Maybe instead of duplicating `expectOne` function, we could just have some helper function, like so:

```
const req1 = expectOne(participantGuid1);
const req2 = expectOne(participantGuid2);
...
const expectOne = (participantGuid: string): TestRequest => {
        return  httpTestingController.expectOne(`${backendUrl}/pepper/v1/user/${participantGuid}/profile`);
    }
```

juanseba21 (approved):

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/participantProfileServiceAgent.service.ts
```diff
@@ -0,0 +1,40 @@
...
+            const {guid, profile} = participant;
+            return this.updateParticipantProfile(guid, profile);
+        });
+    }
+
+    private updateParticipantProfile(guid: string, profile: UserProfile): Observable<any> {
+        const profileChanges: object = {};
+        for (const key of Object.keys(profile)) {
+            if (profile[key]) {
+                profileChanges[key] = profile[key];
```
MocanaAtBroad, line 31: Is it a problem that if profile[key] value is null or '' we don't cannot set it? Guess we cannot clear out a property using this approach?

ddp-workspace/projects/ddp-atcp/src/app/components/participant-list/participant-list.component.ts
```diff
@@ -64,7 +66,22 @@ export class ParticipantListComponent implements OnInit, OnDestroy {
     this.anchor.addNew(
       this.languageService
         .getProfileLanguageUpdateNotifier()
-        .pipe(skipWhile(value => value === null))
+        .pipe(
+          skipWhile(value => value === null),
```
MocanaAtBroad, line 70: This skipWhile needed? This notifier thing I would expect would emit whenever a language change has occurred (type is Observable<void>), so checking for values (even null seems odd). You found you needed this while actually trying the code?

ddp-workspace/projects/ddp-atcp/src/app/components/participant-list/participant-list.component.ts
```diff
@@ -64,7 +66,22 @@ export class ParticipantListComponent implements OnInit, OnDestroy {
...
+          // update the preferred language of all dependents/participants if it is required by configuration
+          filter(() => this.config.updatePreferredLanguageForGovernedParticipants && !!this.participants?.length),
```
MocanaAtBroad, line 72: Is it a problem that we are looking for `this.participants` ? Looks like this variable is set asynchronously at startup.

MocanaAtBroad: (I am looking now at that `getParticipants()` method. Geez! I want to run to the mountains go cry! I know you didn't do that one @alextanlan )

ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/participantProfileServiceAgent.service.ts
MocanaAtBroad: Maybe this whole thing can be simplified by having a method that just updates the language in the profile?
I am understanding the API and the backend implementation, you can send in the patch body something like:
{"preferredLanguage": "es"} 
This weird thing we are doing here updating all the governed participants is only applicable for language and nothing else, so I think it is acceptable to have a method dedicated for it.

MocanaAtBroad (approved):

PR #1876, opened 2022-11-14
[PEPPER-317] [Playwright] Synchronization improvement: wait for request to finish for input and checkbox

aweng98:
Occasionally, after click a button to finish one page and go to the next new page, the expected page navigation does not happen because one or more PATCH /answers requests have not complete successfully.

This PR attempts to address issue by checking and waiting for any triggered requests to finish successfully (**status = 200**) when entering text in Input and checking a checkbox.

Other types of UI fields (Select, Button, etc.) are not considered at this time.

Ran all tests on localhost a dozen time without any failures.

Screen Shot 2022-11-14 at 3 29 53 PM

### DIFF
--- a/playwright-e2e/lib/widget/Input.ts
+++ b/playwright-e2e/lib/widget/Input.ts
@@ -17,9 +17,33 @@ export default class Input {
     return this.locator;
   }
 
-  async fill(value: string): Promise<void> {
+  /**
+   * Fill in text value. Waits for triggered request to finish successfully.
+   * If no request from filling in input, set parameter "waitRequestAfter" to false. For example, fill('How old are you?', { waitRequestAfter: false });
+   * @param value
+   * @param waitOpts
+   */
+  async fill(
+    value: string,
+    waitOpts: { waitRequestAfter?: boolean; requestStatus?: number; requestURL?: string } = {}
+  ): Promise<void> {
+    // For majority of inputs, a "PATCH /answers" request is triggered when filled in text
+    const { waitRequestAfter = true, requestStatus = 200, requestURL = '/answers' } = waitOpts;
+
+    const responsePromise = waitRequestAfter
+      ? this.page.waitForResponse(
+          (response) => {
+            const requestData: boolean | undefined = response.request().postData()?.includes(value);
+            const includeValue = requestData ? requestData : false;
+            return response.url().includes(requestURL) && response.status() === requestStatus && includeValue;
+          },
+          { timeout: 30 * 1000 } // More time for retries. UI will resend failed requests
+        )
+      : Promise.resolve();
+
     await this.toLocator().fill(value);
-    await this.toLocator().press('Tab');
+    await this.toLocator().press('Tab'); // A request is triggered when input loses focus
+    await responsePromise;
   }
 
   errorMessage(): Locator {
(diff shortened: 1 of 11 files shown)


playwright-e2e/tests/singular/enrollment/adult-dependent-enrollment-visual.spec.ts
```diff
@@ -49,7 +49,7 @@ test.describe('Adult Dependent visual tests', () => {
     // Clear the error message by select `The dependant being enrolled`
     await whoHasVentricleHeartDefect.uncheck(WHO.SomeoneElse);
     await whoHasVentricleHeartDefect.check(WHO.TheDependantBeingEnrolled);
-    await page.waitForResponse((resp) => resp.url().includes('/answers') && resp.status() === 200);
```
aweng98, line 52: Explicit waitForResponse are no longer needed.

aweng98:
@bskinner @alextanlan Hi, I'm waiting for reviews. Any question about this PR? Thanks.

playwright-e2e/pages/singular/enrollment/pre-screening-page.ts
```diff
@@ -82,12 +82,14 @@ export default class PreScreeningPage {
       hasHeartDefect = true
     } = opts;
 
-    await this.age().fill(age);
+    await this.age().fill(age, { waitRequestAfter: false });
     await this.country().select().selectOption(country);
```
MocanaAtBroad, line 86: I am surprised you need it here as the UI is designed to disable button while PATCH is in flight. But You do need it, right?

aweng98: Are you asking why we need to wait for triggered pending request to finish after type/select/click?

An example. Say, we have a button called 'Next', when clicked, it navigates to a new page. This button will be disabled during PATCH (sent a request & wait for response) request. However, the problem is, if targeted field is not a **Required** field, button was already enabled before sending PATCH request. 🤷‍♂️
Playwright playback speed is very fast. Playwright checks the state of button after type/select/click on a field and clicks immdiately because button was enabled (It doesn't check again and doesn't know any pending PATCH requests). Triggered request sometimes is a second or more late after field loses focus. We have to handle this special synchronization ourself.

Let me know if your question was not about this.

playwright-e2e/lib/widget/Input.ts
```diff
@@ -17,9 +17,33 @@ export default class Input {
...
+    const responsePromise = waitRequestAfter
+      ? this.page.waitForResponse(
+          (response) => {
+            const requestData: boolean | undefined = response.request().postData()?.includes(value);
+            const includeValue = requestData ? requestData : false;
+            return response.url().includes(requestURL) && response.status() === requestStatus && includeValue;
+          },
+          { timeout: 30 * 1000 } // More time for retries. UI will resend failed requests
+        )
+      : Promise.resolve();
```
bskinner: Am I understanding the intent of code correctly as -> "If `waitRequestAfter` is `true`, wait for up to 30s for the PATCH requests to complete. Otherwise, resolve the promise immediately"?

aweng98: Yes, correct.

aweng98: Second thought, waiting for 30 sec is actually too long. Lowering to 10 seconds.

PR #1891, opened 2022-11-21
Pepper 185 Now does not update multiple fields with the same date.

ngigliotosBroad:
This change makes it so that if multiple requests happen on the same day, the next FaxSent, will not be saved. As in, if a document is saved 3 times on 11/21/2022, then ONLY FaxSent1 will have the value 11/21/2022 rather than FaxSent1, FaxSent2, and FaxSent3.


ddp-workspace/projects/ddp-dsm-ui/src/app/participant-page/participant-page.component.ts
```diff
@@ -828,13 +828,19 @@ export class ParticipantPageComponent implements OnInit, OnDestroy, AfterViewChe
                 oncHis.faxSentBy = this.role.userID();
                 this.oncHistoryValueChanged(oncHis.faxSent, 'faxSent', oncHis);
               } else if (oncHis.faxSent2 == null) {
-                oncHis.faxSent2 = Utils.getFormattedDate(date);
-                oncHis.faxSent2By = this.role.userID();
-                this.oncHistoryValueChanged(oncHis.faxSent2, 'faxSent2', oncHis);
+                //If current date is not already on FaxSent1
+                if(oncHis.faxSent != Utils.getFormattedDate(date)) {
+                  oncHis.faxSent2 = Utils.getFormattedDate(date);
```
SimoneMaiwald: it would be nicer if you pull `Utils.getFormattedDate(date)` into it's own parameter and then compare the faxSents to that parameter instead of always calling the same method over and over

ngigliotosBroad: @SimoneMaiwald I refactored the code, let me know if that is what you had in mind

SimoneMaiwald (dismissed):

pegahtah (dismissed):

pegahtah:
You _may_ need to run the lint check on this one again

SimoneMaiwald (approved):

pegahtah (approved):

PR #1929, opened 2022-12-02
Pepper-117 ShortID must now be 6 characters only

ngigliotosBroad:
Added a function to check if the shortID is 6 characters on the initial scan page. If it is not exactly 6 characters then the button will be disabled. This is for ticket [Pepper-117](https://broadworkbench.atlassian.net/browse/PEPPER-117)


SimoneMaiwald (changes requested):
This scan component is used for 
* Initial Scan
* Tracking Scan
* Final Scan

you are now checking the right value in all of these pages if it is length 6.
But only the Initial Scan should do that!

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
```diff
@@ -450,4 +450,16 @@ export class ScanComponent implements OnInit {
     }
     return null;
   }
+
+  isSixCharacters(): boolean {
```
pegahtah: shouldn't the name be `isNotSixCharacters`?

pegahtah (approved):

pegahtah (changes requested):
Oh yes Simone's comment is very important

pegahtah (dismissed):

GiCharkviani (changes requested):
I would suggest showing an appropriate error message (Hint).
Currently, this check does nothing but just disabling the "Save Scan Pairs" button.

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
```diff
@@ -450,4 +450,22 @@ export class ScanComponent implements OnInit {
...
+  isNotSixCharacters(): boolean {
+    if(!this.initialScan)
+    {
+      return this.scanPairs.length < 2
+    }
+
+    if(this.scanPairsValue.length === 0) {
+      return true;
+    }
+    for (var pair in this.scanPairsValue) {
```
GiCharkviani: I would suggest using the let keyword (or const) instead of **var**, as **var** stores variables in a global object, which later might be problematic.
[Article](https://dev.to/mindninjax/stop-using-var-for-declaring-variables-2p3a) for reference

GiCharkviani: [More about the VAR keyword](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var)

GiCharkviani (commented):
Running CircleCi Tests...

**Update:**
Tests failed. Please see [the job](https://app.circleci.com/pipelines/github/broadinstitute/ddp-angular/21746/workflows/501337be-68a2-4179-96ee-76556eb18321/jobs/32329)

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
ngigliotosBroad: Done Diddly Did.  I changed it to const

GiCharkviani (commented):
I think it would be more convenient if we were more specific about which input field has not met the requirement.

Currently, it just displays the general message:

Screenshot 2022-12-06 at 13 52 51

GiCharkviani (commented):
@SimoneMaiwald 
Just wondering, why don't we have an extra check on short ID's length in the backend?
I think It would also help the front-end in displaying an accurate message.

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
```diff
@@ -450,4 +450,22 @@ export class ScanComponent implements OnInit {
     }
     return null;
   }
+
+  isNotSixCharacters(): boolean {
```
GiCharkviani: the name doesn't comply with its functionality, as this function also performs other checks

GiCharkviani (changes requested):
Tests are failing.
Please see [the job](https://app.circleci.com/pipelines/github/broadinstitute/ddp-angular/21761/workflows/a93408bf-f298-4f16-9ac8-94731ae6385a/jobs/32349).

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
ngigliotosBroad: Gotcha, I changed the name.

GiCharkviani (commented):
I was still able to send the request.
Please see the video :

https://user-images.githubusercontent.com/77500504/206182719-17a4528e-cf30-431f-b244-ac63a303230f.mov

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
```diff
@@ -450,4 +453,40 @@ export class ScanComponent implements OnInit {
     }
     return null;
   }
+
+  public areScanPairsValid(): boolean {
```
GiCharkviani: According to the function name: if this function returns true, will this mean that the scanParis is valid?

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
```diff
@@ -450,4 +453,40 @@ export class ScanComponent implements OnInit {
...
+    if(!this.initialScan)
+    {
+      return this.scanPairs.length < 2;
+    }
+
+    if(this.scanPairsValue.length === 0) {
+      return true;
+    }
+    for (const pair in this.scanPairsValue) {
```
GiCharkviani: You could simplify this loop like so:
```
    for (const {leftValue, rightValue} of this.scanPairsValue) {
      if(rightValue.length !== 6 || leftValue.length === 0) {
        return true;
      }
    }
```

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
```diff
@@ -450,4 +453,40 @@ export class ScanComponent implements OnInit {
...
+    }
+    return false;
+  }
+
+  public makeScanErrorMsg(): void {
+
+    this.scanErrorMsg = '';
+
+    for (const pair in this.scanPairsValue) {
+      if(this.scanPairsValue[pair]['rightValue'].length !== 6 &&
```
GiCharkviani: You could simplify this loop like so:

```
    for (const {leftValue, rightValue} of this.scanPairsValue) {
      if(rightValue.length !== 6 && leftValue.length === 0) {
        this.scanErrorMsg = 'Kit Label cannot be blank and ShortID must be 6 characters long';
      }
      else if(rightValue.length !== 6) {
        this.scanErrorMsg = 'ShortID must be 6 characters long';
      }
      else if(leftValue.length === 0) {
        this.scanErrorMsg = 'Kit Label cannot be blank';
      }
    }
```

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
```diff
@@ -450,4 +453,40 @@ export class ScanComponent implements OnInit {
...
+    for (const pair in this.scanPairsValue) {
+      if(this.scanPairsValue[pair]['rightValue'].length !== 6 ||
+      this.scanPairsValue[pair]['leftValue'].length === 0) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  public makeScanErrorMsg(): void {
```
GiCharkviani, line 502: This function is called too many times. Only on init, it was called 55 times.
Maybe to do something with it?
Please see the video:

https://user-images.githubusercontent.com/77500504/206187857-0aa612f3-14f0-4a46-931b-c31c710e3471.mov

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
ngigliotosBroad: If it returns false if the pairs are valid. I can change the name to areScanPairsInvalid()

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
ngigliotosBroad: Thanks! That looks much nicer

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
ngigliotosBroad: Thanks!

ngigliotosBroad:
I fixed the above bug Giorgi showed, but I had to change how some of the internal variables were being used to do so. I verified the kit received and kit tracking still work as intended after these changes. I also changed the name of the function and improved how the code looked as Giorgi suggested. Lastly, the error message function only gets called when either the left or right value changes. This should decrease the calls of the function as requested.

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
```diff
@@ -193,10 +195,11 @@ export class ScanComponent implements OnInit {
         if (this.scanTracking) {
           const scanPayloads = [];
           this.scanPairsValue.forEach(element => {
+            if(element.rightValue !== '' && element.leftValue !== '') {
             scanPayloads.push({
-              kitLabel: element.rightValue,
-              trackingReturnId: element.leftValue
-            });
+              kitLabel: element.leftValue,
+              trackingReturnId: element.rightValue
```
SimoneMaiwald: did we change the order of things here?

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
SimoneMaiwald: @ngigliotosBroad was this resolved?

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
ngigliotosBroad: Thank you for pointing this out! I thought we did not, but we did. I reverted this specific change.

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
```diff
@@ -110,20 +108,22 @@ export class ScanComponent implements OnInit {
...
-    for (let i = 0; i < this.scanPairsValue.length; i++) {
-      if (i !== position) {
-        if (this.scanPairsValue[position].leftValue != null && labelValue === this.scanPairsValue[i].leftValue
-          || this.scanPairsValue[position].rightValue != null && labelValue === this.scanPairsValue[i].rightValue) {
-          isDuplicate = true;
-        }
-      } else {
-        if (isLeft) {
-          if (this.scanPairsValue[position].rightValue != null && labelValue === this.scanPairsValue[i].rightValue) {
+    if(labelValue !== '') {
```
GiCharkviani, line 108: Just found:
This function is called too many times, which, I think, is not necessary:

https://user-images.githubusercontent.com/77500504/207808119-b363b5ce-0ac0-4ca2-b212-ec98edff080a.mov

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
```diff
@@ -139,6 +139,7 @@ export class ScanComponent implements OnInit {
   private addNewScanPair(): void {
     const newRow = new ScanPairComponent();
```
GiCharkviani, line 155: Why are we instantiating a new Angular component here?

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
```diff
@@ -450,4 +455,49 @@ export class ScanComponent implements OnInit {
...
+      return this.scanPairs.length < 2;
+    }
+
+    //No valid pair has been made yet, so we want a disabled button
+    if(this.scanPairs.length === 1) {
+      return true;
+    }
+    for (const {leftValue, rightValue} of this.scanPairsValue) {
+      //If both are null I want the button to be enabled, so return false
+      if(rightValue === '' && leftValue === '') {
```
GiCharkviani, line 487: Here you are checking whether the right and left values equal an empty string or not. I think the comment is misleading - you don't check for `null`.

P.S. The button is not enabled if both are empty strings

GiCharkviani (commented):
As I see the condition for adding new scan pair has been changed.
e.g. before it used to add new scan pair only after filling the 'Short ID' input field, and now 'Kit Label' also does the same.

@SimoneMaiwald Is it fine?

Please see the video:

https://user-images.githubusercontent.com/77500504/207812024-2cb35c15-f3f0-47d3-9317-8597b7bee905.mov

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
```diff
@@ -110,20 +108,22 @@ export class ScanComponent implements OnInit {
 
   private validateValue(labelValue: string, position: number, isLeft: boolean): boolean {
     let isDuplicate = false;
```
GiCharkviani, line 107: As far as I understand, this function checks whether values are duplicated or not, hence the function name (validateValue) is misleading. I would suggest renaming it to 'isDuplicate' or something similar.

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
```diff
@@ -110,20 +108,22 @@ export class ScanComponent implements OnInit {
...
-          isDuplicate = true;
-        }
-      } else {
-        if (isLeft) {
-          if (this.scanPairsValue[position].rightValue != null && labelValue === this.scanPairsValue[i].rightValue) {
+    if(labelValue !== '') {
+      for (let i = 0; i < this.scanPairsValue.length; i++) {
+        if (i !== position) {
+          if (this.scanPairsValue[position].leftValue != null && labelValue === this.scanPairsValue[i].leftValue
+            || this.scanPairsValue[position].rightValue != null && labelValue === this.scanPairsValue[i].rightValue) {
```
GiCharkviani: This conditional check is too long and hard to read.
I would suggest extracting this logic into a separate function and calling that function in `if()`
Like so:
```
private validateValue(...) {
...
  if( iAmNotNullAndBlaBla() ) {
  ...
  }
...
}

private iAmNotNullAndBlaBla() {
...
}
```

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
```diff
@@ -156,6 +157,7 @@ export class ScanComponent implements OnInit {
       if (this.scanPairs.length < 1) {
         const newScanPair = new ScanPairComponent();
         this.scanPairs.push(newScanPair);
+        this.scanPairsValue.push(new ScanPair('', ''));
```
GiCharkviani, line 175: I think it would be better if empty strings were default. like so:

```
export class ScanPair {
  public error: string;

  constructor(public leftValue: string = '',  public rightValue: string = '') {}
}

```

Then you would just instantiate this class like so: `new ScanPair()`

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
ngigliotosBroad: That was there since before this PR. I believe line 140 is what adds another input field to the UI. So after the first label or short id is filled out the second one will appear because of this function. This will also create the first pair that appears in the UI. Does this answer your question @GiCharkviani ?

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
ngigliotosBroad: I changed it so the function is no longer called when the page is loaded, but the other calls are a result of changing the  condition for adding a new scan pair. Specifically it is the change made to scan-pair-component.ts. I am currently waiting for Simone to reply to your question

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
ngigliotosBroad: I fixed the comment to be correct I think. The button should disabled when both are empty and there is only one pair. The button is enabled if the most recent pair is empty but all the previous ones are not.

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
ngigliotosBroad: Done diddly did 👍

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
ngigliotosBroad: I added a helper function as you asked to make the if statement a bit easier to follow.

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
ngigliotosBroad: Done diddly did 👍

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
```diff
@@ -84,53 +85,71 @@ export class ScanComponent implements OnInit {
...
-    for (let i = 0; i < this.scanPairsValue.length; i++) {
-      if (i !== position) {
-        if (this.scanPairsValue[position].leftValue != null && labelValue === this.scanPairsValue[i].leftValue
-          || this.scanPairsValue[position].rightValue != null && labelValue === this.scanPairsValue[i].rightValue) {
-          isDuplicate = true;
-        }
-      } else {
-        if (isLeft) {
-          if (this.scanPairsValue[position].rightValue != null && labelValue === this.scanPairsValue[i].rightValue) {
+    console.log("here");
```
pegahtah: I think you forgot to remove console log from here

ngigliotosBroad: My bad! Removed

pegahtah (dismissed):

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
GiCharkviani: I see. It just doesn't make sense to instantiate a new Angular component (which is a way bigger thing than just a class) only for filling the empty array.
+ As far as I investigated, we don't even use the values stored in the 'scanPairs', which means that even if you pushed strings or numbers it would be just fine :D 

I would just suggest slowly, whenever you can, to start improving the code a little bit.

GiCharkviani: It's just my opinion, do as you wish.

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
```diff
@@ -84,53 +85,70 @@ export class ScanComponent implements OnInit {
     }
   }
 
-  private checkIfKitLabelChanged(left: string, right: string, position: number): boolean {
-    for (let i = 0; i < this.scanPairsValue.length; i++) {
-      if (this.scanPairsValue[i].leftValue === left && i === position) {
-        this.scanPairsValue[i].rightValue = right;
-        return true;
-      }
+  private shouldAddNewPair(position: number): boolean {
```
GiCharkviani, line 88: You could simplify this function like so:
```
 private shouldAddNewPair(position: number): boolean {
    return this.scanPairs.length === position + 1;
  }
```

GiCharkviani (commented):
2 questions:
1. Is it fine that we are able to upload even last scan pair are empty?
2. Is not the title redundant? -> "No need to select a realm!", since the realm is always selected.

@SimoneMaiwald 

https://user-images.githubusercontent.com/77500504/208045883-5a5eb416-c0d5-4836-a7ea-7cd31de3fb17.mov

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
GiCharkviani: @ngigliotosBroad

ddp-workspace/projects/ddp-dsm-ui/src/app/scan/scan.component.ts
```diff
@@ -84,53 +85,70 @@ export class ScanComponent implements OnInit {
...
+              isDuplicate = true;
+            }
           }
         }
       }
     }
     return isDuplicate;
   }
 
+  isNullOrDuplicate(right: boolean, left: boolean, position: number, index: number, labelValue: string): boolean {
```
GiCharkviani: Since this function is used only in `.ts` file, we could make it `private`

SimoneMaiwald:
https://github.com/broadinstitute/ddp-angular/pull/1929#pullrequestreview-1218789947
Good catch @GiCharkviani That would add a bug! after entering the kit label the focus of the input field needs to move to the shortId and not to the next kit label! GP is entering these fields with a scanner and doesn't want to move the focus manually!
@ngigliotosBroad please change that

https://github.com/broadinstitute/ddp-angular/pull/1929#pullrequestreview-1220391223
1) we always had an empty scan pair. because DSM page is automatically adding the next scan pair after the second value is entered. but we didn't sent that empty scan pair to the backend that was removed from the response before sending
2) yes study staff needs to select a realm when they log into DSM but they do not need to select a study when they are on the scan pages, to tell DSM which study the kits are for.
but "realm" should not be a study staff message anymore. can we please change that "No need to select a realm!" to "No need to select a study!"

PR #1943, opened 2022-12-08
[Playwright] [PEPPER-505] Set up test support classes for E2E RGP tests

aweng98:
I'm starting PEPPER-337 (currently in the backlog but I think it will be added to next Sprint) in preparation for new RGP automation tests.
Created necessary RGP page-objects files which are required for writing of new RGP enrollment tests. This PR does not contains any new RGP test.

`build` and `lint` checks passed.

Screen Shot 2022-12-08 at 2 06 04 PM


playwright-e2e/lib/component/Question.ts
```diff
@@ -24,19 +25,23 @@ export default class Question {
...
 
   /**
-   * A dropdown.
-   * <br> Tag name: mat-select or select
-   * @param value
+   * Tag Select.
+   * <br> Note: use toSelect() for tag "mat-select".
+   * @param label
    */
-  select(value?: string): Locator {
```
aweng98, line 31: As I found out, `Select()` function does not work for tag `mat-select`. Change this function to support only tag `Select`, and new `toSelect()` function to support tag `mat-select`.

playwright-e2e/pages/page-base.ts
```diff
@@ -105,8 +105,9 @@ export default abstract class PageBase implements PageInterface {
   }
 
   /** Click "Submit" button */
-  async submit(): Promise<void> {
-    await this.clickAndWaitForNav(this.getSubmitButton(), { waitForNav: true });
+  async submit(opts: { waitForNav?: boolean } = {}): Promise<void> {
+    const { waitForNav = true } = opts;
+    await this.clickAndWaitForNav(this.getSubmitButton(), { waitForNav });
```
aweng98, line 110: Click `Submit` button does not trigger a page navigation in RGP.

.circleci/config.yml
```diff
@@ -630,6 +630,16 @@ jobs:
...
+            # Read Auth0 RGP client credentials
+            export rgpDomain=$(vault read --format=json secret/pepper/test/v1/e2e | jq -r ".data.auth0 | .[] | select(.app==\"rgp\" and .env==\"<< parameters.env >>\") | .domain")
+            export rgpAudience=$(vault read --format=json secret/pepper/test/v1/e2e | jq -r ".data.auth0 | .[] | select(.app==\"rgp\" and .env==\"<< parameters.env >>\") | .audience")
+            export rgpClientId=$(vault read --format=json secret/pepper/test/v1/e2e | jq -r ".data.auth0 | .[] | select(.app==\"rgp\" and .env==\"<< parameters.env >>\") | .clientId")
+            export rgpClientSecret=$(vault read --format=json secret/pepper/test/v1/e2e | jq -r ".data.auth0 | .[] | select(.app==\"rgp\" and .env==\"<< parameters.env >>\") | .clientSecret")
+            echo "export RGP_AUTH0_DOMAIN=rgpDomain" >> playwright-env/envvars
+            echo "export RGP_AUTH0_AUDIENCE=rgpAudience" >> playwright-env/envvars
+            echo "export RGP_AUTH0_CLIENT_ID=rgpClientId" >> playwright-env/envvars
+            echo "export RGP_AUTH0_CLIENT_SECRET=rgpClientSecret" >> playwright-env/envvars
+
```
MocanaAtBroad, line 642: I would recommend instead of extracting this info from `vault` directly to get it from pepperConfig.js file.
Will be less lines of code and we are trying to move to GCP Secret Manager .
Here is one place where we call build-study.sh which generates the pepperConfig.js
https://github.com/broadinstitute/ddp-angular/blob/b2146e488df0f570ffff0eac99b78c227532fcfa/.circleci/config.yml#L176

aweng98: I think that's a different matter which could be addressed in a future PR. I found this [ticket](https://broadworkbench.atlassian.net/browse/PEPPER-352) which could include Playwright test secrets management.

playwright-e2e/utils/test-utils.ts
```diff
@@ -16,7 +16,7 @@ export async function waitForNoSpinner(page: Page): Promise<void> {
 }
 
 export async function waitUntilRemoved(locator: Locator): Promise<void> {
-  expect(await locator.count()).toHaveLength(0);
+  await expect(locator).toHaveCount(0);
```
aweng98, line 19: See https://playwright.dev/docs/test-assertions#locator-assertions-to-have-count
`expect(locator)` waits until the expected condition is met or until time out is reached.

.circleci/config.yml
MocanaAtBroad: I am just pointing out that the info you seek to extract from Vault is in that pepperConfig.js generated by that script. This is how the Angular applications get these exact same values from pepperConfig.js . When we move to Secret Manager we will continue to use pepperConfig.js in Angular apps so it would be stable way to get them

aweng98: Got it. I hope I can update all Playwright secrets to use GCP secret manager in one shot in a small and scoped PR.

aweng98: To confirm my understanding, this change is not required for approval, right?

MocanaAtBroad (approved):

aweng98:
@ngigliotosBroad PTAL

GiCharkviani (approved):

aweng98:
@GiCharkviani Thank you so much.

PR #1965, opened 2023-01-10
Dsm pepper 492 dynamic form data

GiCharkviani:
[PEPPER-492](https://broadworkbench.atlassian.net/browse/PEPPER-492?focusedCommentId=72500&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-72500)

On the participant list page, in the dynamic form data fields, it was displaying values incorrectly.

**SOLUTION:**
The one way to fix that was to modify `additionalValuesJson` in the participant.model.ts, but as far as I didn't have the settings object, which was necessary for mapping values, and also didn't want to pass an extra parameter to the constructor, or the `.parse()` method, as it's called at many locations in the app, I've decided to solve it locally, right from the template, which is not the best solution, but it works.

Up until now, I didn't find any bug introduced by this change. If you find it, please let me know and I will fix it ASAP.

After fix:

https://user-images.githubusercontent.com/77500504/211499683-8abea0c5-4c81-4155-bedf-eacae0c4bab4.mov

[PEPPER-492]: https://broadworkbench.atlassian.net/browse/PEPPER-492?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/participant-list.component.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/participant-list.component.ts
@@ -36,6 +36,17 @@ import { LoadingModalComponent } from '../modals/loading-modal.component';
 import { BulkCohortTagModalComponent } from '../tags/cohort-tag/bulk-cohort-tag-modal/bulk-cohort-tag-modal.component';
 import { CohortTagComponent } from '../tags/cohort-tag/cohort-tag.component';
 import { CohortTag } from '../tags/cohort-tag/cohort-tag.model';
+import {FieldSettingsModel, ValueModel} from "../STORE/models";
+
+interface IAdditionalValues {
+  singularConsentAudit: string;
+  singularDateMrUploaded: string;
+  singularEnrollmentStatus: string;
+  singularMrAudit: string;
+  singularMrStatus: string;
+  singularMrUploaded: string;
+  singularSecondaryResult: string;
+}
 
 @Component({
   selector: 'app-participant-list',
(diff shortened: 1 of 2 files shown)


ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/participant-list.component.ts
```diff
@@ -36,6 +36,17 @@ import { LoadingModalComponent } from '../modals/loading-modal.component';
 import { BulkCohortTagModalComponent } from '../tags/cohort-tag/bulk-cohort-tag-modal/bulk-cohort-tag-modal.component';
 import { CohortTagComponent } from '../tags/cohort-tag/cohort-tag.component';
 import { CohortTag } from '../tags/cohort-tag/cohort-tag.model';
+import {FieldSettingsModel, ValueModel} from '../STORE/models';
+
+interface IAdditionalValues {
```
SimoneMaiwald: did you check for other studies with additional fields like Circadia and RGP? do we need to add the tab names in the frontend now?

GiCharkviani: Do we need to check for studies? 
My code simply just maps new values to `additionalValuesJson` field.
If there is no `additionalValuesJson`, returns just null and the template won't be rendered, or if there is no value to map, simply just won't map. 
You can find my code from 809th line (`parseAdditionalValues` method)

Another part I didn't understand

GiCharkviani: Don't know how they should look for the other studies :( If you find any issues, let me know and I will adjust.

GiCharkviani: **Update:**
Removed `IAdditionalValues` interface, just to avoid confusion as in `additionalValuesJson` field might be stored a different objects

SimoneMaiwald (approved):

pegahtah (approved):

PR #1972, opened 2023-01-19
PEPPER-314: composite questions

SimoneMaiwald:
changing the type only if the composite has options or groups if not type should stay as before

### DIFF
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/participant-list.component.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/participant-list.component.ts
@@ -456,10 +456,10 @@ export class ParticipantListComponent implements OnInit {
                     }
                   } else if (question.questionType === 'COMPOSITE') {
                     options = new Array<NameValue>();
-                    type = Filter.OPTION_TYPE;
                     if (question.childQuestions != null) {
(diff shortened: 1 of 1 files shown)


GiCharkviani (approved):
LGTM

ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/participant-list.component.ts
```diff
@@ -470,6 +470,7 @@ export class ParticipantListComponent implements OnInit {
                           });
                         }
                         if (childQuestion.groups != null) {
```
zyme, line 472: Should this be an `else if` instead of `else`?  Wondering if there's anything special about the case where there are `childQuestions.options` as well as `childQuestion.groups`.

GiCharkviani: @zyme 
I think it is so.
study: **OS PE-CGS:**
Screenshot 2023-01-20 at 17 26 21

SimoneMaiwald: there are groups **and** options

pegahtah (approved):

PR #1980, opened 2023-01-30
DSM Playwright removed date from tests

GiCharkviani:
Removed dates from test strings, which were possibly failing tests


aweng98 (commented):
Simpler fix suggestion: Move `currentDate = new Date().toLocaleString();` outside of `test` block to make its' value constant for the duration of test?  For example, move it to `test.beforeEach` so it is initated once and same for each test run.

aweng98 (approved):
Ran test locally few times, passed every time.

PR #1996, opened 2023-02-07
DSM Playwright improvements v2

GiCharkviani:
Added participant table component for DSM


playwright-e2e/lib/component/dsm/table/table.ts
```diff
@@ -0,0 +1,51 @@
+import { Locator, Page } from '@playwright/test';
+import ParticipantPage from 'pages/dsm/participant-page';
+
+export class Table {
```
aweng98: Is this new class created for Participant table only? If yes, we should rename it. If no, other types of tables cannot use this.
Also, could you add jsdoc for all functions?

playwright-e2e/lib/component/dsm/table/table.ts
```diff
@@ -0,0 +1,51 @@
...
+  private readonly _participantPage: ParticipantPage = new ParticipantPage(this.page);
+
+  constructor(private readonly page: Page) {}
+
+  public async openParticipantPageAt(position: number): Promise<ParticipantPage> {
```
aweng98: `position` is row index, right? And, looks like your position starts from 1, not zero. Which means you have to minus 1 when working with Playwright `.nth()` function (example: `.nth(position - 1)`). I'd prefer not doing it.

playwright-e2e/lib/component/dsm/table/table.ts
```diff
@@ -0,0 +1,51 @@
...
+  public async selectParticipantAt(position: number): Promise<void> {
+    return await this.getParticipantAt(position).nth(0).locator('mat-checkbox').click();
+  }
+
+  private getParticipantAt(position: number): Locator {
+    return this.page.locator(`//table/tbody/tr`).nth(position - 1);
+  }
+
+  /* XPaths */
+  private getParticipantXPathBy(columnName: string, columnValue: string): string {
```
aweng98: Xpath is long. I'd break into smaller pieces to help with reading. For example, I can make this new function:
```
private theadCount(columnName: string): string {
  return count(//table/thead/` +
      `th[text()[normalize-space()='${columnName}']]/preceding-sibling::th)
}
```
Then use this function inside other functions.

playwright-e2e/lib/component/dsm/table/table.ts
```diff
@@ -0,0 +1,51 @@
...
+    return (
+      `//table/thead/th[text()[normalize-space()='${columnName}']]` +
+      `/ancestor::table/tbody//td[position()=` +
+      `count(//table/thead/th[text()[normalize-space()='${columnName}']]` +
+      `/preceding-sibling::th)+1][.//*[text()[normalize-space()='${columnValue}']]]/ancestor::tr`
```
aweng98: BTW, this is a slow and confusing xpath selector because: find a element, then find element's ancestor that contains a new element (which is same as first element) then find element's ancestor tr again.

playwright-e2e/lib/component/dsm/table/table.ts
GiCharkviani: This table can be used for the Tissue List page as well, there is a similar table. 
I also thought of renaming it and actually, I did it locally. Let me push.

playwright-e2e/lib/component/dsm/table/table.ts
GiCharkviani: It is already taken into account, you can find it in this function -> `getParticipantAt`;
I did it so because I thought it would be easier for the other developer to select a participant

aweng98: Sounds good. jsdoc would be helpful to explain it's 1-indexed. I prefer zero-indexed because almost all programming languages have been designed this way for arrays, maps, etc. indexing from 0.

playwright-e2e/lib/component/dsm/table/table.ts
GiCharkviani: Oki, Done!

playwright-e2e/lib/component/dsm/table/table.ts
GiCharkviani: I know, but this is the functionality I wanted to be there, If you know better way, please suggest

playwright-e2e/lib/component/dsm/table/table.ts
GiCharkviani: Yes, I know, but it's Playwright, where we are selecting items from view, where it is the 1, not 0, that's why for a better understanding of which item we are selecting, I decided so. I can change, NP :)

GiCharkviani: Changed

SimoneMaiwald (approved):

playwright-e2e/pages/dsm/participantList-page.ts
```diff
@@ -1,14 +1,14 @@
 import { expect, Locator, Page } from '@playwright/test';
 
 import { waitForNoSpinner } from 'utils/test-utils';
-import Checkbox from 'lib/widget/checkbox';
-import ParticipantPage from './participant-page';
-import { Filters } from '../../lib/component/dsm/filters/filters';
+import { Filters } from 'lib/component/dsm/filters/filters';
+import { ParticipantListTable } from 'lib/component/dsm/participantListTable/participantListTable';
```
aweng98: Looks like the folder name were accidentally changed from `table` to `participantListTable`, right?

GiCharkviani: Oh, yes. Fixed - > tables

PR #2015, opened 2023-02-16
[Playwright] PEPPER-247 CircleCI scheduler launches playwright-test and build-and-deploy-all-apps workflows at the same time

aweng98:
I noticed this issue today in CircleCI with playwright-test schedule.

Issue description: 
Whenever CircleCI launches scheduled `scheduled-playwright-test` workflow, it also launches `build-and-deploy-all-apps-workflow` workflow at the same time. We don't want that to happen. It happens because the condition to launch `build-and-deploy-all-apps-workflow`  workflow is only `unless: << pipeline.parameters.do-builds >>`.  The default value for `do_builds` is false so the check always pass when scheduler launches any workflow.

Screenshot 2023-02-15 at 7 36 37 PM

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,6 +30,8 @@ references:
   # Job runs for develop branch only
   filter-develop-branch: &filter-develop-branch
     filters:
+      tags:
+        ignore: /.*/
       branches:
         only: develop
 
@@ -38,7 +40,11 @@ references:
       tags:
         ignore: /.*/
       branches:
-        ignore: develop
+        ignore:
+          - develop
+          - master
+          - /^rc.*/
+          - /^hotfix.*/
 
 # Container for all the builds
 executors:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -38,7 +40,11 @@ references:
       tags:
         ignore: /.*/
       branches:
-        ignore: develop
+        ignore:
+          - develop
+          - master
+          - /^rc.*/
+          - /^hotfix.*/
```
aweng98, line 47: Enhance `filter-pr-branch` reference to exclude rc, hotfix and main branches after learning more about branches today.

.circleci/config.yml
```diff
@@ -400,6 +406,27 @@ commands:
       - store_artifacts:
           path: /tmp/junit
 
+  halt-playwright-check:
```
aweng98, line 409: Using this check in Playwright jobs to ensure Playwright jobs do not run for Staging, Prod and UNKNOWN env.

.circleci/config.yml
```diff
@@ -710,6 +730,9 @@ parameters:
   e2e-test-parallelism:
     type: integer
     default: 1
+  scheduled-playwright:
+    type: boolean
+    default: false
```
aweng98, line 735: New pipeline parameter to stop launching `build-and-deploy-all-apps-workflow` by CI scheduler.

zyme (approved):

PR #2023, opened 2023-02-21
DSM-commented-out-unreliable-assertion

GiCharkviani:
I have found two things that are worth attention:

1. the **test-results.xml** file is overwritten by the most recent job if two or more of them are running, which causes a wrong report from the **ARTIFACTS** section (CircleCi)
2. something happens on CircleCi VM (Linux x86_64) which causes from time-to-time JS to generate the wrong date (`1382486400000 -> Wed Oct 23 2013 04:00:00 GMT+0400 (Georgia Standard Time)`) from the `new Date().getTime()` function (The reason why the Submission manager unit test fails)

**Here is the solution for the failing test (2):**
It would be better to remove the test based on the `new Date()` class (that fails occasionally).

It seems the JS Date class is not always predictable:
[Dates in JavaScript are Broken. Who Shall Fix them](https://hackernoon.com/dates-in-javascript-are-broken-who-shall-fix-them)
[Inconsistent Date.prototype.getTime() value across different Node.js versions](https://github.com/nodejs/node/issues/29900)
[Mocking Date.now() doesn't work on CircleCI](https://github.com/facebook/jest/issues/3341)

Failed [test](https://app.circleci.com/pipelines/github/broadinstitute/ddp-angular/23432/workflows/7245b96e-f27a-4cf8-8ee2-23596741aa21/jobs/37279/artifacts) (CircleCi) - Please check the `tmp/junit/test-results.xml` file, where you can search for "START_TIME" and "END_TIME" and check that they both equal to - > 1382486400000

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/submissionManager.service.spec.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/services/serviceAgents/submissionManager.service.spec.ts
@@ -144,7 +144,7 @@ describe('SubmissionManagerTest', () => {
             expect(returnedInProgressValues[1].delay).toBeLessThan(httpCallDelay);
             expect(returnedInProgressValues[2].value).toBe(false);
             // the last status change should occur after the http call completed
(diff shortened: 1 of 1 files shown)


aweng98 (approved):
LGTM

PR #2039, opened 2023-03-01
[DSM] PEPPER-531 Lost To Follow Up participant tab and MR cover pdf download

pegahtah:
I changed the condition to show the onc history tab for participants after their status is changed to Lost To Follow Up, also changed the MR cover PDF download to have the AOM date as default as the end date.


ddp-workspace/projects/ddp-dsm-ui/src/app/medical-record/medical-record.component.ts
```diff
@@ -60,6 +60,7 @@ export class MedicalRecordComponent implements OnInit {
 
   startDate: string;
   endDate: string;
+  hasEndDateChanged = false;
```
pegahtah, line 63: This has been added so if they manually changed the end date they can over write the default AOM date

GiCharkviani (approved):
LGTM

PR #2072, opened 2023-03-21
[DSM] GP Final Scan hot fix

pegahtah:
I have made changes to the display so the returned short id is treated differently than an error
Screen Shot 2023-03-21 at 2 15 19 PM
Screen Shot 2023-03-21 at 2 17 42 PM

### DIFF
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/scanner/scanner.component.html
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/scanner/scanner.component.html
@@ -47,6 +47,8 @@
       <button mat-mini-fab type="button" [disabled]="scannerFields.length - 1 === scannerFieldIndex" (click)="removeFields(scannerFieldIndex)" color="primary" >-</button>
       <p class="Width--100 Color--warn" *ngIf="scannerField.hasError('notFound')">
         Error occurred sending this scan pair! <br/> {{scannerField.getError('notFound')}}</p>
+      <p class="Width--100" *ngIf="shortIds[scannerFieldIndex]">
+        Scanned successfully for <br/> {{shortIds[scannerFieldIndex]}}</p>
     </div>
   </ng-container>
   <br />
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/scanner/scanner.component.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/scanner/scanner.component.ts
@@ -77,7 +77,7 @@ export class ScannerComponent implements DoCheck, OnDestroy {
   public save(): void {
     this.additionalMessage = '';
     const filteredActiveScannerFieldsGroupsArray = this.filteredNonNullFieldsGroups;
-    this.shortIds = Array.apply(null, Array(this.scannerFields.length)).map(function () {});
+    this.shortIds = Array(...Array(this.scannerFields.length)).map(function() {});
     this.activeScanner.saveFn(filteredActiveScannerFieldsGroupsArray)
       .pipe(
         takeUntil(this.subscriptionSubject$)
@@ -181,7 +181,7 @@ export class ScannerComponent implements DoCheck, OnDestroy {
 
     responseData.forEach((data: any, index: number) =>
       {
-      data?.hasOwnProperty('shortId') && this.setShortId(index, data)
+      data?.hasOwnProperty('shortId') && this.setShortId(index, data);
       }
     );
 
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/scanner/scanner.component.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/scanner/scanner.component.ts
@@ -35,6 +35,8 @@ export class ScannerComponent implements DoCheck, OnDestroy {
   });
   public additionalMessage: string | undefined;
 
+  private shortIds = [];
+
   private updatePreviousFieldValidations = false;
   private readonly subscriptionSubject$: Subject<void> = new Subject<void>();
 
@@ -67,6 +69,7 @@ export class ScannerComponent implements DoCheck, OnDestroy {
 
   public removeFields(index: number): void {
     this.scannerFields.length - 1 && this.scannerFields.removeAt(index);
+    this.shortIds.length - 1 && this.shortIds.splice(index, 1);
     this.activeScanner.inputFields.forEach((inputField: InputField) =>
       this.checkForDuplicates(inputField.controllerName));
   }
@@ -74,7 +77,7 @@ export class ScannerComponent implements DoCheck, OnDestroy {
   public save(): void {
     this.additionalMessage = '';
     const filteredActiveScannerFieldsGroupsArray = this.filteredNonNullFieldsGroups;
-
+    this.shortIds = Array.apply(null, Array(this.scannerFields.length)).map(function () {});
     this.activeScanner.saveFn(filteredActiveScannerFieldsGroupsArray)
       .pipe(
         takeUntil(this.subscriptionSubject$)
@@ -82,9 +85,8 @@ export class ScannerComponent implements DoCheck, OnDestroy {
       .subscribe({
         next: (data: any[]) => {
           this.cdr.markForCheck();
-          if (data.some(d => d)) {
+          if (data.some(d => d?.hasOwnProperty('error') || d?.hasOwnProperty('shortId'))) {
             this.removeSuccessfulScans(data);
-            this.additionalMessage = 'Error - Failed to save all changes';
           } else {
             this.resetForm();
             this.additionalMessage = 'Data saved';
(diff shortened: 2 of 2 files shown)


ddp-workspace/projects/ddp-dsm-ui/src/app/scanner/scanner.component.html
```diff
@@ -47,6 +47,8 @@ <h3 class="Color--warn">{{additionalMessage}}</h3>
       <button mat-mini-fab type="button" [disabled]="scannerFields.length - 1 === scannerFieldIndex" (click)="removeFields(scannerFieldIndex)" color="primary" >-</button>
       <p class="Width--100 Color--warn" *ngIf="scannerField.hasError('notFound')">
         Error occurred sending this scan pair! <br/> {{scannerField.getError('notFound')}}</p>
+      <p class="Width--100" *ngIf="shortIds[scannerFieldIndex]">
+        Scanned successfully for <br/> {{shortIds[scannerFieldIndex]}}</p>
```
pegahtah, line 51: Adding a separate display for returned shortId

ddp-workspace/projects/ddp-dsm-ui/src/app/scanner/scanner.component.ts
```diff
@@ -67,24 +69,24 @@ export class ScannerComponent implements DoCheck, OnDestroy {
...
+    this.shortIds.length - 1 && this.shortIds.splice(index, 1);
     this.activeScanner.inputFields.forEach((inputField: InputField) =>
       this.checkForDuplicates(inputField.controllerName));
   }
 
   public save(): void {
     this.additionalMessage = '';
     const filteredActiveScannerFieldsGroupsArray = this.filteredNonNullFieldsGroups;
-
+    this.shortIds = Array(...Array(this.scannerFields.length)).map(function() {});
```
pegahtah, line 80: Creating a short id with the size of scanned kits

ddp-workspace/projects/ddp-dsm-ui/src/app/scanner/scanner.component.ts
```diff
@@ -169,14 +171,29 @@ export class ScannerComponent implements DoCheck, OnDestroy {
   }
 
   private removeSuccessfulScans(responseData: any[]): void {
-    responseData.forEach((data: any, index: number) => data &&
-      this.scannerFields.at(index).setErrors({notFound: data?.error})
+    responseData.forEach((data: any, index: number) => {
+    if (data?.hasOwnProperty('error')) {
+        this.scannerFields.at(index).setErrors({notFound: data?.error});
+        this.additionalMessage = 'Error - Failed to save all changes';
```
pegahtah, line 177: Set this message only if an error was returned

ddp-workspace/projects/ddp-dsm-ui/src/app/scanner/scanner.component.ts
```diff
@@ -35,6 +35,8 @@ export class ScannerComponent implements DoCheck, OnDestroy {
   });
   public additionalMessage: string | undefined;
 
+  private shortIds = [];
```
GiCharkviani: This field is accessed from the template, therefore it has to be public. While JIT is okay with that, AOT might have some problems.

ddp-workspace/projects/ddp-dsm-ui/src/app/scanner/scanner.component.ts
```diff
@@ -67,24 +69,24 @@ export class ScannerComponent implements DoCheck, OnDestroy {
 
   public removeFields(index: number): void {
     this.scannerFields.length - 1 && this.scannerFields.removeAt(index);
+    this.shortIds.length - 1 && this.shortIds.splice(index, 1);
```
GiCharkviani, line 72: I think this part -> `this.shortIds.length - 1` is not necessary, as according to this function:
```
 responseData.forEach((data: any, index: number) =>
      {
      data?.hasOwnProperty('shortId') && this.setShortId(index, data);
      }
    );
```
It will store the values to the `shortIds` array that was returned from the backend, which in theory, must be in the same order as the front end send those, therefore there is no need to check for the empty pair (Last ones) as they just won't be there.

Especially, if you removed this part:
`this.shortIds = Array(...Array(this.scannerFields.length)).map(function() {});`
Which I think is also not needed.

What do you think?

ddp-workspace/projects/ddp-dsm-ui/src/app/scanner/scanner.component.ts
GiCharkviani, line 80: I don't think we need to do that, as we can still add items to the array, even without explicitly defining its length in forward (I mean at any index).

**Example:**
```
const exampleArray = [];
exampleArray[2] = 'data 1'
exampleArray[5] = 'data 2'

console.log(exampleArray)
console.log(exampleArray[2])
console.log(exampleArray[5])

// Output:
// [empty × 2, 'data 1', empty × 2, 'data 2']
// data 1
// data 2
```

Or if you still want to use the Array constructor, you could simplify your code like so:
`this.shortIds = Array(this.scannerFields.length)`

ddp-workspace/projects/ddp-dsm-ui/src/app/scanner/scanner.component.ts
```diff
@@ -67,24 +69,24 @@ export class ScannerComponent implements DoCheck, OnDestroy {
...
+    this.shortIds = Array(...Array(this.scannerFields.length)).map(function() {});
     this.activeScanner.saveFn(filteredActiveScannerFieldsGroupsArray)
       .pipe(
         takeUntil(this.subscriptionSubject$)
       )
       .subscribe({
         next: (data: any[]) => {
           this.cdr.markForCheck();
-          if (data.some(d => d)) {
+          if (data.some(d => d?.hasOwnProperty('error') || d?.hasOwnProperty('shortId'))) {
```
GiCharkviani, line 88: When the response object includes the `shortId` field and we don't remove those scan pairs, maybe better to create a separate function that would only be responsible for setting a success message under those scan pairs, as I think it would be more convenient because we don't remove scan pairs, hence we don't need to perform that action under the function called removeSuccessfulScans.

**For example:**
To remove that function and do something like this:
```
if (data.some(d => d?.hasOwnProperty('error') || d?.hasOwnProperty('shortId'))) {
            this.checkResponse(data); // could be better name
          } else {
            this.resetForm();
            this.additionalMessage = 'Data saved';
          }

...

public checkResponse(responseData: any[]): void {
responseData.forEach((data: any, index: number) => {
    if(data?.hasOwnProperty('error')) {
        this.setError(index) // would set error to the formGroup at the passed index
    }  else if(data?.hasOwnProperty('shortId')) {
         this.shortIdSuccess(index) // would set or store successful messages, which later would be rendered in the UI
       }
      }
    );

  this.removeSuccessfullySavedFields() // would remove other successful pairs, as per legacy logic
}
```

What do you think?

ddp-workspace/projects/ddp-dsm-ui/src/app/scanner/scanner.component.ts
```diff
@@ -169,14 +171,29 @@ export class ScannerComponent implements DoCheck, OnDestroy {
...
+
+    responseData.forEach((data: any, index: number) =>
+      {
+      data?.hasOwnProperty('shortId') && this.setShortId(index, data);
+      }
     );
+
     for (let i = this.scannerFields.length - 2; i >= 0; i--) {
-      !this.scannerFields.at(i).hasError('notFound') && this.scannerFields.removeAt(i);
+      !this.scannerFields.at(i).hasError('notFound') && !this.shortIds[i] && this.scannerFields.removeAt(i);
```
GiCharkviani, line 189: What do you think about setting the different type of error, instead of the `notFound`, in which case you would not need to create a separate array and render it separately in the UI? 
You would only need to check for that error and act accordingly

GiCharkviani (approved):

ddp-workspace/projects/ddp-dsm-ui/src/app/scanner/scanner.component.ts
pegahtah: Since this is not an error, I don't think we should add it to `hasError`

ddp-workspace/projects/ddp-dsm-ui/src/app/scanner/scanner.component.ts
pegahtah: Since if they remove the entry at an index, the short id is also removed at that index, then we need to initialize the short id.  Otherwise you are correct

pegahtah:
This is the hotfix for [PEPPER-676](https://broadworkbench.atlassian.net/browse/PEPPER-676) [Ticket here](https://broadworkbench.atlassian.net/browse/PEPPER-676). This PR will be closed and directly deployed to prod, another PR is made for the develop

[PEPPER-676]: https://broadworkbench.atlassian.net/browse/PEPPER-676?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2076, opened 2023-03-27
[PEPPER-693][DSM][BUG] - Support multiple browser tabs.

otchet-broad:
[PEPPER-693](https://broadworkbench.atlassian.net/browse/PEPPER-693) was discovered while testing to identify the root cause of a data corruption issue identified in [PEPPER-688](https://broadworkbench.atlassian.net/browse/PEPPER-688).

This is PR represents a minimally viable way to support multiple browser tabs.  Existing DSM-UI localStorage lookups for the **realm** have been updated to use [sessionStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage).  With this change users will be able to:

- Reliably use multiple browser tabs without one tab overwriting another
- Log in to DSM with each new tab to the application
- ~~Use a different JWT or user account in each tab~~
- ~~Log out of each tab without affecting work in another~~
- Use Chrome's duplicate tab feature to quickly get started in a new DSM tab

Current items impacted:
First tab in Chrome to use DSM will always prompt for user login and study selection.  Our DSM playwright tests expect this behavior and were not impacted.
Use Chrome's "Duplicate tab" feature will bring a logged in user to a new tab with the same study loaded.  The user can now use these tabs independently.  
~~Logging out in one tab will not break using another tab.~~
Logging out in one tab logs all tabs out
The "oncHistory" issue cannot be repeated in the UI.
Participant/Tissue Lists will always display the lists for the selected study.

No DSM tests appear to have been impacted by these changes.  **The changes should be reviewed with product owners and leads to discuss potential login/workflow issues in DSM.**

Follow-on work may include:

- Refactoring code so a service is responsible for in-memory storage of attributes outside of well known locations in session storage (Same XSS Risk as with localStorage).
- Adding support for extracting "realm" from URL.  This is a heavier lift than anticipated because of the current unauthenticated to authenticated flow.
- Improved UX for User settings (indicate/use realm for settings like filters)

[PEPPER-693]: https://broadworkbench.atlassian.net/browse/PEPPER-693?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PEPPER-688]: https://broadworkbench.atlassian.net/browse/PEPPER-688?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddp-workspace/projects/ddp-dsm-ui/src/app/services/auth.service.ts
```diff
@@ -118,9 +118,6 @@ export class Auth {
   }
 
   public authenticated(): boolean {
-    // Check if there's an unexpired JWT
```
otchet-broad, line 121: Comments in this block were removed since they don't reflect what is actually happening in isAuthenticated().

GiCharkviani (commented):
Since you don't call anymore the `localStorage` from the LocalStorageService, maybe it would be appropriate to rename the LocalStorageService to SessionStorageService as well.

GiCharkviani (commented):
Didn't find any issue yet, but as the flow has been changed, let's better wait on Zim's approval.

ddp-workspace/projects/ddp-dsm-ui/src/app/services/auth.service.ts
```diff
@@ -89,11 +89,11 @@ export class Auth {
 
   constructor(private router: Router, private activatedRoute: ActivatedRoute, private http: HttpClient,
               private sessionService: SessionService, private role: RoleService,
-              private compService: ComponentService, private dsmService: DSMService, private localStorageService: LocalStorageService
+              private compService: ComponentService, private dsmService: DSMService, private localStorageService: SessionStorageService
                ) {
     // Add callback for lock `authenticated` event
     this.lock.on('authenticated', (authResult: any) => {
-      localStorage.setItem(Auth.AUTH0_TOKEN_NAME, authResult.idToken);
+      sessionStorage.setItem(Auth.AUTH0_TOKEN_NAME, authResult.idToken);
```
pegahtah: Why not keep the auth0 token in the `localStorage` and the study in the `sessionStorage`, and change the authentication guards and log in page?
This way the user can log in once, use separate tabs, and once they log out in one tab, they are out of DSM. This way the log in flow remains unchanged to the user

ddp-workspace/projects/ddp-dsm-ui/src/app/tissue-page/tissue-page.component.ts
```diff
@@ -139,7 +139,7 @@ export class TissuePageComponent implements OnInit {
           value: v,
         }, null, 'participantId',
         this.participant.participant.ddpParticipantId, Statics.ONCDETAIL_ALIAS,  null,
-        localStorage.getItem(ComponentService.MENU_SELECTED_REALM), this.participant.data.profile['guid']
+        sessionStorage.getItem(ComponentService.MENU_SELECTED_REALM), this.participant.data.profile['guid']
```
pegahtah, line 142: Thank you for changing all of the places this was called.

General comment - not specific to this line - can we refactor this line `sessionStorage.getItem(ComponentService.MENU_SELECTED_REALM)`?  Maybe we can move it to Utils? 
I am also all for creating new components in favor of refactoring

ddp-workspace/projects/ddp-dsm-ui/src/app/services/auth.service.ts
otchet-broad: Great question and one I thought about while doing this.  It's a similar issue to storing the realm in local storage.  If the user logs in with one set of credentials in the first tab and a different set of credentials in a future tab, the code will incorrectly attribute any follow-on work in the first tab to the user of the second tab.

ddp-workspace/projects/ddp-dsm-ui/src/app/tissue-page/tissue-page.component.ts
otchet-broad: We absolutely can refactor this (component, service, etc), however it will take some additional time to do.  ~~If we decide to go forward with this solution, I'll add a ticket for it.~~ I've added notes about additional enhancements we should consider with both the code (including this) and UX.

ddp-workspace/projects/ddp-dsm-ui/src/app/services/auth.service.ts
otchet-broad: We may also want to significantly refactor our Auth0 work based on the Single Page Application (SPA) advice here; https://auth0.com/docs/secure/security-guidance/data-security/token-storage and implementation examples here: https://auth0.com/docs/libraries/auth0-single-page-app-sdk

otchet-broad:
The outcome of our review meeting was to only address the realm issue, not the login/logout.  I'll be pushing minimal changes to this PR for review.  Follow-on tickets for refactoring will be made.

ddp-workspace/projects/ddp-dsm-ui/src/app/WELCOME/auth/auth.component.ts
```diff
@@ -43,12 +44,24 @@ export class AuthComponent implements OnInit {
...
   pickList$: Observable<any>;
+  private source$: Observable<StorageEvent>;
 
   constructor(private auth: Auth, private title: Title, private router: Router) {}
 
   ngOnInit(): void {
+
+    this.source$ = fromEvent<StorageEvent>(window, 'storage');
+    this.source$.subscribe(
+      event => {
```
otchet-broad: Listen for storage events to see if the DSM_TOKEN_NAME is set to null.  This is how we are able to detect a logout event from another tab.

ddp-workspace/projects/ddp-dsm-ui/src/app/guards/checkAuth.guard.ts
```diff
@@ -15,16 +15,16 @@ export class CheckAuthGuard implements CanLoad {
...
     if (!this.auth.authenticated() && !selectedRealm) {
       return true;
     }
     else if (this.auth.authenticated() && !selectedRealm) {
-      this.auth.logout();
+      this.auth.sessionLogout();
       return this.router.createUrlTree(['']);
     } else if (!this.auth.authenticated() && selectedRealm) {
-      this.auth.logout();
+      this.auth.doLogout();
```
otchet-broad, line 27: This path deletes all of the tokens because the local tab is no longer considered authenticated, even though a realm is selected.

ddp-workspace/projects/ddp-dsm-ui/src/app/guards/checkAuth.guard.ts
```diff
@@ -15,16 +15,16 @@ export class CheckAuthGuard implements CanLoad {
...
   canLoad(route: Route, segments: UrlSegment[]): boolean | UrlTree {
-    const selectedRealm = localStorage.getItem(ComponentService.MENU_SELECTED_REALM);
+    const selectedRealm = sessionStorage.getItem(ComponentService.MENU_SELECTED_REALM);
 
     if (!this.auth.authenticated() && !selectedRealm) {
       return true;
     }
     else if (this.auth.authenticated() && !selectedRealm) {
-      this.auth.logout();
+      this.auth.sessionLogout();
```
otchet-broad, line 24: This path leaves the login token because removing it will log other users out.  What we need here is to get the user back to the study pickList.

GiCharkviani (commented):
@otchet-broad Thanks you for your work. Just a couple of things we might have to pay attention to. Please see my comments below.

ddp-workspace/projects/ddp-dsm-ui/src/app/WELCOME/auth/auth.component.ts
```diff
@@ -43,12 +44,24 @@ export class AuthComponent implements OnInit {
   popUpIsShown = true;
   authError: Observable<string | null>;
   pickList$: Observable<any>;
+  private source$: Observable<StorageEvent>;
 
   constructor(private auth: Auth, private title: Title, private router: Router) {}
 
   ngOnInit(): void {
+
```
GiCharkviani: As this component is in a lazily loaded module and is under the `canLoad` guard, there will be cases (e.g. refreshing tab while logged in) when this component won't be downloaded, therefore this logic about listening to the storage event won't be executed, which means that the user still will need to refresh the other tab to detect the change, even though thanks to guards, navigation won't be allowed anymore, still the user will no longer log out automatically from the application. 

https://user-images.githubusercontent.com/77500504/229788734-812367dc-4512-4c68-8b05-f086771a778e.mov

GiCharkviani (commented):
There is still required to authenticate each time the user opens the application in the new tab. Is it desired behavior?

otchet-broad:
Known issue:

> There is still required to authenticate each time the user opens the application in the new tab. Is it desired behavior?

They've said they consider that an acceptable tradeoff.

ddp-workspace/projects/ddp-dsm-ui/src/app/WELCOME/auth/auth.component.ts
```diff
@@ -43,12 +44,24 @@ export class AuthComponent implements OnInit {
...
+    this.source$ = fromEvent<StorageEvent>(window, 'storage');
+    this.source$.subscribe(
```
GiCharkviani: I would suggest (It's the best and recommended practice) unsubscribing any subscribed observable (e.g. `this.source$`) in order to avoid memory leaks and unexpected behavior in the future.

GiCharkviani (commented):
I see: 

> They've said they consider that an acceptable tradeoff.

We might find some workaround in the future if they want.

ddp-workspace/projects/ddp-dsm-ui/src/app/WELCOME/auth/auth.component.ts
otchet-broad: I've moved this to the auth service.  Good catch.

ddp-workspace/projects/ddp-dsm-ui/src/app/WELCOME/auth/auth.component.ts
```diff
@@ -43,12 +44,13 @@ export class AuthComponent implements OnInit {
   popUpIsShown = true;
   authError: Observable<string | null>;
   pickList$: Observable<any>;
+  private source$: Observable<StorageEvent>;
```
GiCharkviani: Tiny change: let's remove this unused field.

GiCharkviani: And imports as well (fromEvent and SessionService)

otchet-broad: Surprised the linter didn't catch that!  They're cleaned up now.

ddp-workspace/projects/ddp-dsm-ui/src/app/services/auth.service.ts
```diff
@@ -46,6 +46,14 @@ export class Auth {
 
   selectedStudy = new BehaviorSubject<string>('');
 
+  storageEventSource$ = fromEvent<StorageEvent>(window, 'storage').subscribe(
+    event => {
+      if (event.key === SessionService.DSM_TOKEN_NAME && event.newValue === null) {
```
GiCharkviani: **[OPTIONAL]**
What do you think about moving the subscription part to the constructor function?
Something like this:
```
class authService {

  storageEventSourceSubscription$: Subscription;


 constructor() {
     this.storageEventSourceSubscription$ = fromEvent<StorageEvent>(window, 'storage').subscribe(
      event => {
        if (event.key === SessionService.DSM_TOKEN_NAME && event.newValue === null) {
          this.doLogout();
          this.router.navigateByUrl('/');
        }
      }
    );
  }
}
```

GiCharkviani: Or to the separate function, which will be called in the constructor

GiCharkviani (approved):
LGTM.

ddp-workspace/projects/ddp-dsm-ui/src/app/services/auth.service.ts
otchet-broad: I'm fine moving it into the constructor.  Done.

PR #2084, opened 2023-04-06
[PEPPER-717] Expired token action may have different message text in some cases; Consolidates error handling due to expired credentials in Auth.doLogout

otchet-broad:
[PEPPER-717](https://broadworkbench.atlassian.net/browse/PEPPER-717)
While testing the first fix for this ticket, a developer was not able to successfully log into the application.  Upon inspection, the error text was different in this case.  This PR removes the inspection of the error.message value since it can change and we have enough information outside of this text to determine we need to send the user to the login page.

Further testing revealed that logout handling was insufficient for cases where the token had expired.  Logic is consolidated in Auth.doLogout where possible (except in a case where cyclical dependencies prevented it).

[PEPPER-717]: https://broadworkbench.atlassian.net/browse/PEPPER-717?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/abstraction-field/abstraction-field.component.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/abstraction-field/abstraction-field.component.ts
@@ -44,7 +44,8 @@ export class AbstractionFieldComponent implements OnInit {
   modalReadOnly: boolean;
   _other: string;
 
-  constructor(private dsmService: DSMService, private router: Router, private compService: ComponentService, private role: RoleService) {
+  constructor(private dsmService: DSMService, private router: Router, private compService: ComponentService,
+              private role: RoleService, private auth: Auth) {
   }
 
   ngOnInit(): void {
@@ -308,7 +309,7 @@ export class AbstractionFieldComponent implements OnInit {
           },
           error: err => {
             if (err._body === Auth.AUTHENTICATION_ERROR) {
-              this.router.navigate([ Statics.HOME_URL ]);
+              this.auth.doLogout();
             }
           }
         });
@@ -343,7 +344,7 @@ export class AbstractionFieldComponent implements OnInit {
       },
       error: err => {
         if (err._body === Auth.AUTHENTICATION_ERROR) {
-          this.router.navigate([ Statics.HOME_URL ]);
+          this.auth.doLogout();
         }
       }
     });
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/sequencing-order/sequencing-order.component.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/sequencing-order/sequencing-order.component.ts
@@ -62,7 +62,7 @@ export class SequencingOrderComponent {
       },
       error: err => {
         if (err._body === Auth.AUTHENTICATION_ERROR) {
-          this.auth.sessionLogout();
+          this.auth.doLogout();
         }
       }
     } );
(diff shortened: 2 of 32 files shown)


ddp-workspace/projects/ddp-dsm-ui/src/app/services/dsm.service.ts
```diff
@@ -1165,8 +1165,9 @@ export class DSMService {
 
   private checkCookieBeforeCall(): boolean {
     if (this.sessionService.getDSMToken() == null) {
+      this.localStorageService.clear();
```
otchet-broad, line 1168: Can't import the auth service because of a cyclical dependency, so logic handles the updates needed.

GiCharkviani (approved):
I didn't find any issue locally yet. Let's merge it and see if anyone can find anything

GiCharkviani (commented):
Concerns:
1. The authentication flow depends only on the selected study -> whether we should request authentication in a new tab or not.
2. The user is not able to log in from a different account, which means that, every time a new tab is opened, the user should reauthenticate with the same credentials. Just thinking, if it makes sense asking the reauthentication every time only for the purpose of electing study - maybe we should find the other solution.

ddp-workspace/projects/ddp-dsm-ui/src/app/sequencing-order/sequencing-order.component.ts
```diff
@@ -62,7 +62,7 @@ export class SequencingOrderComponent {
       },
       error: err => {
         if (err._body === Auth.AUTHENTICATION_ERROR) {
-          this.auth.sessionLogout();
+          this.auth.doLogout();
```
zyme, line 65: Do we have documentation and/or tests on the backend that will help everyone understand how critical `Auth.AUTHENTICATION_ERROR` is?  Just in case we forget.

Can we also add a few lines in the DSM regression test doc that describe how to dial down the session timeout time and test a handful of the logout scenarios?  I don't think we need an automated test right now, but having a manual test documented would be a great start.

ddp-workspace/projects/ddp-dsm-ui/src/app/abstraction-field/abstraction-field.component.ts
```diff
@@ -308,7 +309,7 @@ export class AbstractionFieldComponent implements OnInit {
           },
           error: err => {
             if (err._body === Auth.AUTHENTICATION_ERROR) {
-              this.router.navigate([ Statics.HOME_URL ]);
```
zyme, line 311: Was this code assuming (rightly or wrongly?) that `HOME_URL` would handle logout properly?

zyme (approved):

ddp-workspace/projects/ddp-dsm-ui/src/app/abstraction-field/abstraction-field.component.ts
otchet-broad: Exactly.

ddp-workspace/projects/ddp-dsm-ui/src/app/sequencing-order/sequencing-order.component.ts
otchet-broad: Agree with your thinking Zim.
I'll work to do a few things here:

-  I was thinking about the same thing regarding the importance of the Auth.AUTHENTICATION_ERROR and briefly scanned our backend code to see if this was under test.  I didn't find it there.  Would it be okay to add a ticket so we add a backend unit or integration test that verifies this case?
- Get the token expiration time adjustments documented for dev, test and staging (I think we observed they're different at least between dev and test)
- Add steps to the multi-tab section on the regression spreadsheet for logout and token expiration.

GiCharkviani (approved):
LGTM

pegahtah (approved):
LGTM, let's get it to testing.
Thank you so much Elliot

PR #2085, opened 2023-04-11
[PEPPER-731] Auth0 invalid signature returns a 400 response.

otchet-broad:
[PEPPER-731](https://broadworkbench.atlassian.net/browse/PEPPER-731) was discovered while testing in dev. Zim recieved responses from Auth0 with invalid signature errors.  As a result, the web client received errors with 400 status codes.

This PR updates the logic that checks Auth0 Errors to include the 400 error code as a candidate to examine.

[PEPPER-731]: https://broadworkbench.atlassian.net/browse/PEPPER-731?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/services/auth.service.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/services/auth.service.ts
@@ -179,8 +179,7 @@ export class Auth implements OnDestroy {
 
   private checkForAuth0Error(error: any): boolean {
     return (error instanceof HttpErrorResponse &&  error && error.hasOwnProperty('status')
-      && error.hasOwnProperty('ok') && error.url &&
-      (error.status === 401 || error.status === 400 || error.status === 0)
+      && error.hasOwnProperty('ok') && error.url
       && !error.ok && error.url.includes('ui/auth0'));
   }
   private handleError(error: any): Observable<any> {
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/services/auth.service.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/services/auth.service.ts
@@ -180,7 +180,7 @@ export class Auth implements OnDestroy {
   private checkForAuth0Error(error: any): boolean {
     return (error instanceof HttpErrorResponse &&  error && error.hasOwnProperty('status')
       && error.hasOwnProperty('ok') && error.url &&
-      (error.status === 401 || error.status === 0)
+      (error.status === 401 || error.status === 400 || error.status === 0)
       && !error.ok && error.url.includes('ui/auth0'));
   }
   private handleError(error: any): Observable<any> {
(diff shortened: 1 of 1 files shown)


pegahtah (commented):
I am wondering if we are still missing something, maybe we should check that error is not 500?

zyme (commented):
LGTM, matches what Elliot and I saw on my browser.

otchet-broad:
@pegahtah - true that this might leave other status codes affected that we haven't seen yet.  @zyme and I discussed that in our zoom - so the question is to both of you since I don't have a thumb yet:
Would you prefer I collapse the first two status code checks into something like the condition below?
(error.status => 400 && error.status < 500) || (error.status === 0)

This opens the aperture a bit wider on 4XX errors and other factors limit the response to just auth0:
 && !error.ok && error.url.includes('ui/auth0'));

This would make the evaluation:
    return (error instanceof HttpErrorResponse &&  error && error.hasOwnProperty('status')
      && error.hasOwnProperty('ok') && error.url &&
      ((error.status >= 400 && error.status < 500) || error.status === 0)
      && !error.ok && error.url.includes('ui/auth0'));

pegahtah:
@otchet-broad (I tried replying to your comment but seems like github doesn't have that option)
I think this is a better solution, I am worried about the case of `404`  in this, but we don't have a special way of handling that at the moment as well.... I think as long as we don't force the user to logout on a 404, we are good.
Would love to know what @zyme thinks as well

> @pegahtah - true that this might leave other status codes affected that we haven't seen yet. @zyme and I discussed that in our zoom - so the question is to both of you since I don't have a thumb yet: Would you prefer I collapse the first two status code checks into something like the condition below? (error.status => 400 && error.status < 500) || (error.status === 0)
> 
> This opens the aperture a bit wider on 4XX errors and other factors limit the response to just auth0: && !error.ok && error.url.includes('ui/auth0'));
> 
> This would make the evaluation: return (error instanceof HttpErrorResponse && error && error.hasOwnProperty('status') && error.hasOwnProperty('ok') && error.url && ((error.status >= 400 && error.status < 500) || error.status === 0) && !error.ok && error.url.includes('ui/auth0'));

otchet-broad:
A bit more searching and it looks like [this code](https://github.com/broadinstitute/ddp-study-server/blob/develop/pepper-apis/dsm-core/src/main/java/org/broadinstitute/dsm/route/AuthenticationRoute.java) returns either a 400 or a 401.  It doesn't look like it will return others.

denniscunningham:
@otchet-broad Does any code other than 200 need handling? IOW, even if we don't handle a code ideally right now, would it be wrong to ignore any code other than 200?

otchet-broad:
Since we know we're in an error state from the code below
error instanceof HttpErrorResponse &&

And we're making sure that the path includes auth0:
 && error.url.includes('ui/auth0')

And error.ok will always be false as documented in [this](https://angular.io/api/common/http/HttpErrorResponse) Angular doc (even when it's a 2XX response @denniscunningham):
&& !error.ok 

Maybe we remove the status code check altogether and rely on the path.
Thoughts?  I'm testing locally now.

otchet-broad:
Based on my earlier comment, I've removed the status code check and auth0 errors still seem to have the desired result.
@zyme, @pegahtah, @denniscunningham.

zyme (approved):
I'm good with these changes.  Sounds like we have validated that they catch the various edge conditions we have seen so far.

denniscunningham (approved):

pegahtah (approved):

PR #2086, opened 2023-04-12
[PEPPER-730] Reset roles when setting token.

otchet-broad:
[PEPPER-730](https://broadworkbench.atlassian.net/browse/PEPPER-730) was discovered during release testing.

Before this change, when switching users, roles were not reset.  This could lead the UI to showing features from the previously logged in user.

This PR:
1) Adds logic to reset these values whenever the RoleService.setRoles method is called.
2) Includes handling clearing UserSetting if one isn't provided.
3) Adds code and coverage to ensure valid UserSetting objects are built even if one isn't returned from the backend.

[PEPPER-730]: https://broadworkbench.atlassian.net/browse/PEPPER-730?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


denniscunningham (approved):

pegahtah (approved):
Thank you!

PR #2090, opened 2023-04-19
[DSM] PEPPER-139 Adding new access role for viewing clinical orders

pegahtah:
I have added the new role as the ones that allows the user to view the sequencing tab and the cliniical orders page. They still can not submit orders


ddp-workspace/projects/ddp-dsm-ui/src/app/sequencing-order/sequencing-order.component.html
```diff
@@ -79,7 +79,7 @@
     </div>
     <div class="app-modal-footer">
       <button (click)="closeModal()" class="btn btn-default" type="button">Close</button>
-      <button (click)="submitOrders()" *ngIf="!this.checkError()" class="btn btn-default" type="button">Submit</button>
+      <button (click)="submitOrders()" *ngIf="!this.checkError()" class="btn btn-default" type="button" [disabled]="canNotSubmitOrder()">Submit</button>
```
pegahtah, line 82: There is a guard in the backend that doesn't let the orders go through without proper "submit orders" access role, but I am also removing the buttons from the view if they  only have the view role

denniscunningham: Just so I understand, this is a 'drive-by' fix... not required for this task, but something you noticed that was incorrect?

ddp-workspace/projects/ddp-dsm-ui/src/app/navigation/navigation.component.html
```diff
@@ -15,7 +15,8 @@
         </ul>
       </li>
       <li class="dropdown" *ngIf="getAuth().authenticated()
-        && (hasRole().allowToViewSampleLists() || hasRole().allowedToHandleSamples() || hasRole().allowedToViewReceivingPage() || hasRole().allowedToDiscardSamples() || hasRole().allowedToExitParticipant())">
+        && (hasRole().allowToViewSampleLists() || hasRole().allowedToHandleSamples() || hasRole().allowedToViewReceivingPage() || hasRole().allowedToDiscardSamples() || hasRole().allowedToExitParticipant()
+        || hasRole().allowedToDoOrderSequencing() || hasRole().viewSeqOrderStatus)">
         <a class="dropdown-toggle" data-toggle="dropdown" href="#"><b>Samples</b>
```
denniscunningham, line 20: How does 'allowedToDoOrderSequencing()'  fit into this feature?

otchet-broad: Is viewSeqOrderStatus a property or a function?

pegahtah: I created a test user with only this new permission and participant list permission , and realized that they won't see the `Samples` menu since it only checked for samples and kit shipping access, while people with either of `allowedToDoOrderSequencing` or `viewSeqOrderStatus` should be able to see the `Clinical Orders` menu point in this drop down. That's why I corrected this.

ddp-workspace/projects/ddp-dsm-ui/src/app/sequencing-order/sequencing-order.component.html
pegahtah: It is required in this task that the user with only this new permission, should not be able to submit an order. There is a backend check for this, but I also removed the submit button

ddp-workspace/projects/ddp-dsm-ui/src/app/navigation/navigation.component.html
pegahtah: @otchet-broad `viewSeqOrderStatus` is a get function

denniscunningham: I assumed the issue here was the lack of parens after 'viewSeqOrderStatus'?

pegahtah: Thank you for catching this! It was a naming issue where both the variable and the function had the same name. I fixed it now

ddp-workspace/projects/ddp-dsm-ui/src/app/sequencing-order/sequencing-order.component.html
```diff
@@ -2,7 +2,7 @@
   <table class="table table-striped Width--100">
     <thead>
     <tr>
-      <th>
+      <th *ngIf="!canNotSubmitOrder()">
         Select
```
pegahtah, line 6: I removed the selection check boxes because they are not useful anymore in a view only permission

ddp-workspace/projects/ddp-dsm-ui/src/app/sequencing-order/sequencing-order.component.html
```diff
@@ -31,13 +31,13 @@
...
       <td>
-        <ng-container *ngIf="sample.sampleType === 'Normal'">
+        <ng-container *ngIf="sample.sampleType === 'Normal' && canChangeCollectionDate()">
           <app-field-datepicker (dateChanged)="valueChanged($event, 'collectionDate', sample)" [allowUnknownDay]="false"
                                 [colorDuringPatch]="isPatchedCurrently('collectionDate')"
                                 [dateString]="sample.collectionDate"
                                 [showTodayButton]="true"></app-field-datepicker>
         </ng-container>
-        <ng-container *ngIf="sample.sampleType === 'Tumor'">
+        <ng-container *ngIf="sample.sampleType === 'Tumor' || !canChangeCollectionDate()">
```
pegahtah: I also removed the date submission, but we are waiting on the final decision on that

denniscunningham (approved):
LGTM, but you may want approval from others.

otchet-broad (approved):
Thanks for diving in and answering that question @pegahtah!

ddp-workspace/projects/ddp-dsm-ui/src/app/sequencing-order/sequencing-order.component.ts
```diff
@@ -159,4 +161,23 @@ export class SequencingOrderComponent {
   closeModal(): void {
     this.modal.hide();
   }
+
+  public canNotSubmitOrder(): boolean{
+    return !this.role.allowedToDoOrderSequencing();
+  }
+
+  canEditCollectionDate(sample : SequencingOrder ) {
+  if (sample.sampleType === 'Normal') {
```
denniscunningham: Do you want a return type?
Is the indenting a bit off, or is this just a GIthub view thing?

ddp-workspace/projects/ddp-dsm-ui/src/app/sequencing-order/sequencing-order.component.ts
```diff
@@ -164,7 +166,18 @@ export class SequencingOrderComponent {
...
   }
 
-  public canChangeCollectionDate(): boolean{
-      return this.role.allowedToDoOrderSequencing() || this.role.allowToViewSampleLists();
+  canEditCollectionDate(sample : SequencingOrder ) {
+  if (sample.sampleType === 'Normal') {
+  // only a normal sample should have collectionDate editable, for tumor samples it is the date of px
+    if (this.role.canViewSeqOrderStatus()) {
+    // people with this specific view access can only enter a date if there is no previous date entered
+      return (_.isNull(sample.collectionDate) || _.isUndefined(sample.collectionDate) || _.isEmpty(sample.collectionDate));
```
otchet-broad: I can't remember with the version of underscore we're using, but I thought _.isEmpty returns true for undefined and null as well.  Is this similar to something else in our code?  If so, have a pointer?

otchet-broad: What's the type of sample.collectionDate?  From the [underscore documentation](https://underscorejs.org/#isEmpty):
isEmpty_.isEmpty(collection) [source](https://underscorejs.org/docs/modules/isEmpty.html)
Returns true if collection has no elements. For **strings** and array-like objects _.isEmpty checks if the length property is 0. For other objects, it returns true if the object has no enumerable own-properties. Note that primitive numbers, booleans and symbols are always empty by this definition.

ddp-workspace/projects/ddp-dsm-ui/src/app/sequencing-order/sequencing-order.component.ts
otchet-broad: +1

ddp-workspace/projects/ddp-dsm-ui/src/app/sequencing-order/sequencing-order.component.ts
otchet-broad: It looks like this is a string.  If that's the case, this is fine as is, but likely still could be reduced to just _.isEmpty(sample.collectionDate) because these test pass:

      expect(_.isEmpty(null)).toBe(true);
      expect(_.isEmpty(undefined)).toBe(true);

ddp-workspace/projects/ddp-dsm-ui/src/app/sequencing-order/sequencing-order.component.ts
pegahtah: fixed, thank you :)

ddp-workspace/projects/ddp-dsm-ui/src/app/sequencing-order/sequencing-order.component.ts
pegahtah: it is a string, that's why I kept it as is

denniscunningham (approved):

otchet-broad (approved):

PR #2092, opened 2023-04-24
[Playwright] PEPPER-637 Fix Osteo E2E child-enrollment.spec test

aweng98:
Another text change due to [PEPPER-637](https://broadworkbench.atlassian.net/browse/PEPPER-637)

[PEPPER-637]: https://broadworkbench.atlassian.net/browse/PEPPER-637?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


playwright-e2e/tests/osteosarcoma/child-enrollment.spec.ts
```diff
@@ -190,7 +190,6 @@ test('Osteo enroll kid @osteo', async ({ page }) => {
   await page.waitForTimeout(2000);
   await page.getByRole('button', { name: 'Submit' }).click();
   await page.getByText('A Message from the Osteosarcoma Project').click();
-  await page.getByText('Thank you for providing information regarding your chid’s experiences with osteo').click();
```
Rinchan20, line 193: would it be better to add the fixed typo instead of removing it so that the test can check that the fix is applied?

Rinchan20 (approved):

playwright-e2e/tests/osteosarcoma/child-enrollment.spec.ts
aweng98: Checking is done in other tests like `adult-and-child-enrollment.spec` so I thought no need to duplicate it.

PR #2106, opened 2023-05-04
[Playwright] Fix flaky Osteo playwright test

aweng98:
CircleCI failed [run](https://app.circleci.com/pipelines/github/broadinstitute/ddp-angular/24356/workflows/09bb57e8-994a-4408-8a6b-41189831ff60/jobs/53024/tests).

```
adult-enrollment.spec.ts:25:5 Osteo adult self enroll @osteo
[chromium] › tests/osteosarcoma/adult-enrollment.spec.ts:25:5 › Osteo adult self enroll @osteo ───

    Error: Timed out 30000ms waiting for expect(received).toHaveValue(expected)

    Expected string: "Giant Cell Tumor of the Bone (GCT)"
    Received string: "bone"
    Call log:
      - expect.toHaveValue with timeout 30000ms
      - waiting for locator('.activity-text-input-OTHER_CANCER_NAME').locator('mat-form-field').locator('input').first()
 ```

Test failed because it run faster than application UI can complete rendering. Before clicking `Finish` button, we should wait until page has finished rendering. Fix test by adding few code for synch.

Screenshot 2023-05-04 at 1 08 51 PM

### DIFF
--- a/playwright-e2e/pages/brain/brain-base-page.ts
+++ b/playwright-e2e/pages/brain/brain-base-page.ts
@@ -10,9 +10,4 @@ export abstract class BrainBasePage extends PageBase {
     }
     super(page, BRAIN_BASE_URL);
   }
-
-  async finish(): Promise<void> {
-    await this.page.waitForTimeout(1000);
-    await this.page.getByRole('button', { name: 'Finish' }).click();
-  }
 }
--- a/playwright-e2e/pages/page-base.ts
+++ b/playwright-e2e/pages/page-base.ts
@@ -152,10 +152,9 @@ export default abstract class PageBase implements PageInterface {
   }
 
   async finish(): Promise<void> {
-    await Promise.all([
-      waitForNoSpinner(this.page),
-      this.getFinishButton().click()
-    ]);
+    await waitForNoSpinner(this.page);
+    await this.getFinishButton().click();
+    await waitForNoSpinner(this.page);
   }
 
   /** Click "Next" button */
--- a/playwright-e2e/pages/page-base.ts
+++ b/playwright-e2e/pages/page-base.ts
@@ -70,6 +70,10 @@ export default abstract class PageBase implements PageInterface {
     return this.page.locator('.header button[data-ddp-test="signOutButton"]:has-text("Log Out")');
   }
 
+  getFinishButton(): Locator {
+    return this.page.getByRole('button', { name: 'Finish' });
+  }
+
   /**
    * Return "Next" button locator
    */
@@ -145,6 +149,13 @@ export default abstract class PageBase implements PageInterface {
    */
   async agreeToAllowUsToContactPhysicians(): Promise<void> {
     await new Checkbox(this.page, { label: 'I have already read and signed the informed consent document' }).check();
+  }
+
+  async finish(): Promise<void> {
+    await Promise.all([
+      waitForNoSpinner(this.page),
+      this.getFinishButton().click()
+    ]);
   }
 
   /** Click "Next" button */
--- a/playwright-e2e/tests/osteosarcoma/adult-enrollment.spec.ts
+++ b/playwright-e2e/tests/osteosarcoma/adult-enrollment.spec.ts
@@ -1,5 +1,6 @@
 import { expect, Page } from '@playwright/test';
 import { test } from 'fixtures/osteo-fixture';
+import TextArea from 'lib/widget/textarea';
 import ConsentAddendumPage from 'pages/osteo/consent-addendump-page';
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: playwright-e2e/pages/brain/brain-base-page.ts [hunk 1]
```diff
@@ -10,9 +10,4 @@ export abstract class BrainBasePage extends PageBase {
     }
     super(page, BRAIN_BASE_URL);
   }
-
-  async finish(): Promise<void> {
-    await this.page.waitForTimeout(1000);
-    await this.page.getByRole('button', { name: 'Finish' }).click();
-  }
```
aweng98, line 17: This func can be reuseable so I moved it to `page-base.ts`.

playwright-e2e/tests/osteosarcoma/adult-enrollment.spec.ts
```diff
@@ -347,14 +348,20 @@ test('Osteo adult self enroll @osteo', async ({ page }) => {
   });
 
   await familyHistoryPage.next();
+
+  await new TextArea(page, { ddpTestID: 'answer:FH_OTHER_FACTORS_CANCER_RISK' }).toLocator().waitFor({
+    state: 'visible',
+    timeout: 60 * 1000
+  });
```
aweng98, line 355: Looking for a way to determine whether page has completed rendering, this textarea is visible after rendering.

playwright-e2e/pages/brain/brain-base-page.ts
otchet-broad: Nice!

playwright-e2e/pages/page-base.ts
```diff
@@ -147,6 +151,13 @@ export default abstract class PageBase implements PageInterface {
     await new Checkbox(this.page, { label: 'I have already read and signed the informed consent document' }).check();
   }
 
+  async finish(): Promise<void> {
+    await Promise.all([
+      waitForNoSpinner(this.page),
+      this.getFinishButton().click()
```
otchet-broad: Do you want to wait for the spinner to disappear and then click the finish button?  As is, won't these run concurrently?  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all

aweng98: Great question! I'll do my best to explain.

The order of the promises in `Promise.all` is maintained. The first promise in the array will get resolved first.

More on my thinking about synchronization...

`waitForNoSpinner()` works by checking page one time. It doesn't wait and polls for few seconds to recheck for spinner. That would increase tests' run time.

(1) To check for spinner and then click, I can do this,
```
await waitForNoSpinner(this.page);
await this.getFinishButton().click();
```

(2) If I want to check spinner after click, I can do this,
```
await this.getFinishButton().click();
await waitForNoSpinner(this.page);
```

(3) If I want to check spinner before and after click, I'd have to do this,
```
await waitForNoSpinner(this.page);
await this.getFinishButton().click();
await waitForNoSpinner(this.page);
```
(4) Lastly, `Promise.all`. It works like the first approach.

None actually solves the sync issue 100% of time due to the fact Playwright runs really really fast, `waitForNoSpinner()` finishes in milliseconds if no spinner found. Hoever, spinner can be rendered after the check has finished.
Any one of four approaches will help to reduce the probability of it happens.

aweng98: `concurrent` is a misleading concept from `Promise.all` because JS code in a single-threaded language.

otchet-broad: Right.  I'm using concurrent in the context of the MDN documentation here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#promise_concurrency:~:text=Note%20that%20JavaScript%20is%20single%2Dthreaded%20by%20nature%2C%20so%20at%20a%20given%20instant%2C%20only%20one%20task%20will%20be%20executing%2C%20although%20control%20can%20shift%20between%20different%20promises%2C%20making%20execution%20of%20the%20promises%20appear%20concurrent.

Yes - without workers, the language is single threaded.  I'm not meaning to imply otherwise.

Turning back to the logic flow, do you need to wait for the spinner is gone before clicking finish?  If so, I'd implement the logic in 2 or 3 above.

aweng98: Got it. IMO, # 3 provides little more assurance. I made the change. Thank you.

otchet-broad (approved):
Great discussion and I appreciate all of the work in tracking this down!

PR #2121, opened 2023-05-25
PEPPER-91 - Update language for proxy as well as governed users

rfricke-asymmetrik:
Uses the config setting for whether or not to update governed users preferred language to update the behavior of the languageSelector component to update for the governed users as well as the proxy user. 

The code in the participant-list.component.ts from atcp is removed because this is now being down in the languageSelect.component.ts added in this pull request.


ddp-workspace/projects/ddp-sdk/src/lib/components/internationalization/languageSelector.component.ts
```diff
@@ -267,13 +271,38 @@ export class LanguageSelectorComponent implements OnInit, OnDestroy {
...
     // Update the language in the profile to the current language
     private updateProfileLanguage(): Observable<any> {
-        const profileModifications: UserProfile = new UserProfile();
-        profileModifications.preferredLanguage = this.currentLanguage.languageCode;
-        return this.profileServiceAgent.updateProfile(profileModifications)
-            .pipe(
-                tap(() => this.afterProfileLanguageChange.emit()),
-                tap(() => this.language.notifyOfProfileLanguageUpdate())
-            );
+        if (this.config.updatePreferredLanguageForGovernedParticipants) {
```
otchet-broad, line 274: Based on the feedback in the ticket, is there an **easy** way to incorporate logic that is closer aligned to the response from Whitney?
If (this study allows governed participants) && (this study is translated to more than one language) then
    do this updated code
else
     do the original thing.

If we can do that instead, we should remove the per-study configuration so this becomes the default behavior without having to remember to turn it on.

GiCharkviani (approved):
LGTM

PR #2122, opened 2023-05-31
[Playwright] DSM clinical and research samples check

GiCharkviani:
In case the study is OS2, we are checking if the sample type is either clinical or research, based on the living country/state of the participant


aweng98 (approved):
LGTM

PR #2138, opened 2023-06-22
DSM-PEPPER-810 shared learnings tab

GiCharkviani:
[PEPPER-810
](https://broadworkbench.atlassian.net/jira/software/projects/pepper/boards/164?assignee=621677042d057500723f90e4&selectedIssue=PEPPER-810)

**Shared learnings upload tab:**

Screenshot 2023-06-22 at 22 41 32


ddp-workspace/projects/ddp-dsm-ui/src/app/participant-page/participant-page.component.html
```diff
@@ -185,7 +185,8 @@ <h3 class="Color--warn Line--Break">Lost to Follow-Up</h3>
             <td>Gender</td>
             <td>{{gender | titlecase}}</td>
           </tr>
-          <tr *ngIf="participant.data.profile['preferredLanguage'] != null && participant.data.profile['preferredLanguage'] !== ''">
+          <tr
```
otchet-broad, line 188: Unnecessary CR, but I see the pattern is repeated elsewhere.  Consider if that's necessary and update the code style configuration as needed.

ddp-workspace/projects/ddp-dsm-ui/src/app/sharedLearningUpload/sharedLearningsUpload.component.ts
```diff
@@ -0,0 +1,299 @@
...
+    if (!isVirusFree && isFileDeleted) { // file has been scanned and is infected
+      return SomaticResultsFileVirusStatusEnum.INFECTED;
+    } else if (!isVirusFree && !isFileDeleted) {
+      // file has not been scanned, so it till scan file as well
+      this.scanForVirus(somaticDocumentId);
+      return SomaticResultsFileVirusStatusEnum.SCANNING;
+    } else if (isVirusFree && !isFileDeleted) { // file has already been scanned for viruses and is clean
+      return SomaticResultsFileVirusStatusEnum.CLEAN;
+    } else {
+      // It should not happen, but anyway:
```
otchet-broad, line 172: TODO: isVirusFree && isFileDeleted return SomaticResultsFileVirusStatusEnum.CLEAN;
Not currently needed because deleted items are removed from display.

otchet-broad (approved):
Great job @GiCharkviani.  There are a few things to address, but I think this can merge as is.
1) Code style with CRs in <tr> elements
2) The sharedLearningsUpload.component.ts handleAndReturnVirusStatusFor calculates clean, but deleted files as INFECTED.  They're not visible in the table, but this should be corrected in a small cleanup PR.
3) See comment below about using realms instead of the value obtained from DSS because of the OS1/OS2 issue.

ddp-workspace/projects/ddp-dsm-ui/src/app/participant-page/participant-page.component.ts
```diff
@@ -1761,6 +1761,43 @@ export class ParticipantPageComponent implements OnInit, OnDestroy, AfterViewChe
...
+  }
+
+  public get isParticipantAllowed(): Observable<boolean> {
+    return this.sequencingOrdersArray$
+      .pipe(
+        map((sequencingOrders: SequencingOrder[]) => {
+          const mercuryAllow = sequencingOrders
+            .some(({pdoOrderId, sampleType}: SequencingOrder) => pdoOrderId && sampleType === 'Tumor');
+
+          const allowedStudies = ['cmi-osteo', 'cmi-lms'];
```
otchet-broad, line 1778: @GiCharkviani - I think this has to be updated to look at realms.  This would leak between Osteo and Osteo2 since they are the same studies in DSS and different in DSM.
Since we won't get mercury data back for Osteo, this should be done in the follow-up PR as we won't encounter a case where the tab would be visible.

ddp-workspace/projects/ddp-dsm-ui/src/app/participant-page/participant-page.component.html
GiCharkviani: Elliot, it was just formatted automatically, after using the keyboard shortcut - **options + command + L**.
 all is good here

ddp-workspace/projects/ddp-dsm-ui/src/app/sharedLearningUpload/sharedLearningsUpload.component.ts
GiCharkviani: I guess you mean the **else** case. It is just an **else** case, which, as the comment says, is not expected to happen, but it is there assuming that if it happened, the file is infected, but the backend did something wrong with response data :D .

If you mean **else if case** -  **isVirusFree && !isFileDeleted**, it works correctly, as per discussion: 

> If isVirusFree is true and deletedAt equals 0 (falsy for JS) - the file is clean

Or if you literally mean this case -> **isVirusFree && isFileDeleted**, I don't see such a one in the code

**FYI: Deleted items are already removed from the array once the files array riches the virus scanning process. Files are filtered by the **deletedByUserId** field _(Line 152nd)_**

ddp-workspace/projects/ddp-dsm-ui/src/app/participant-page/participant-page.component.ts
GiCharkviani: @otchet-broad Elliot, the **allowedStudies** are not related to mercury data in this code anyhow. if you look at the code, on line 1779, we are reading the study name from `this.participant.data.ddp?.toLowerCase()`. I checked and both, **osteo** and **osteo2** participants have written the 'cmi-osteo' in the **ddp** field.  If Osteo study doesn't have the mercury data, then it won't be problem, but if yes - then we should read it from the selected study, instead of the **ddp** field.

**Please, check this information with @pegahtah and let me know what to do.**

ddp-workspace/projects/ddp-dsm-ui/src/app/sharedLearningUpload/sharedLearningsUpload.component.ts
otchet-broad: Please fix the  isVirusFree && isFileDeleted case.  If we want to show these deleted items, this latent defect will appear.

PR #2141, opened 2023-06-23
PEPPER-535 removed GA from studies that are actively recruiting

zyme:
For more details about the approach here for PEPPER-535, [please see this doc](https://docs.google.com/document/d/101GrTw3dFgbukNO24pPSBEbLty92_1smGDqAkxXA7ZA).

Each study's index.html loads GA, so these files have had their GA calls removed.  This includes the DSM frontend and each DSS study.  Rather than update many files that make calls to our GA wrapper, I've gone with the approach of commenting out the guts of key methods in the toolkit so that client studies can still call the GA methods, but the methods themselves are no-ops.

Because prion has cookie management that other studies lack, the approach with prion required additional changes so that even if participants opt-in to analytic cookies, the site will still not send data to GA.  Prion's cookie model is now default opt-out instead of opt-in for analytical cookies, and opting in is a no-op that does not result in any metrics going to GA

### DIFF
--- a/ddp-workspace/projects/ddp-angio/src/app/app.module.ts
+++ b/ddp-workspace/projects/ddp-angio/src/app/app.module.ts
@@ -159,9 +159,6 @@ export function translateFactory(translate: TranslateService, injector: Injector
 })
 export class AppModule {
     constructor(private analytics: AnalyticsEventsService) {
-        this.analytics.analyticEvents.subscribe((event: AnalyticsEvent) => {
-            ga('send', event);
-            ga('platform.send', event);
-        });
+
     }
 }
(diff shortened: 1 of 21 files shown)


pegahtah (approved):

GiCharkviani (approved):
LGTM

otchet-broad:
@zyme - Looks good!  

Just one thing I noticed:
- ddp-workspace/projects/toolkit/src/lib/services/analyticsManagement.service.ts looks like it's declaring a constant for GA and has private methods that look like they're still attempting to use it.  Those private methods seem to be unused now, but for completeness should probably also be commented out or removed - since they'll likely break if they were exercised.

otchet-broad (approved):

PR #2146, opened 2023-06-28
DSS-shared learnings virus scan bulk response

GiCharkviani:
After this PR gets merged, instead of having individual states for each row, we will have a global one


otchet-broad:
Nice work @GiCharkviani !  This gets at the heart of the requests and significantly reduces the load the front end generates on the backend for a large set of pending documents for a participant.  Thank you!
Test scenarios I've confirmed locally when running:
1) Large document list polling - One concurrent request for status updates that retrieves all 179 documents.  Before this change, it was 1 request per document.
2) Document upload - works as expected.  Triggers single polling if polling was needed (if a polling request was already in flight, a new one was not created)
3) Document delete - works as expected - did not trigger polling
4) Send to participant - works as expected - did not trigger polling

Memory - observed to be recovered when exiting back to the participant list.  Coincided with polling ending.  Future enhancement: end polling when switching off of the shared learning tab.

otchet-broad (approved):

PR #2152, opened 2023-07-07
Pepper-953-fixing-current-batch-of-playwright-tests

Rinchan20:
fixing dsm-fixtures.ts plus adding a functional tag to some tests for easy running via terminal. Currently around 30 tests passing, 3 skipped, 4 failing, looking into it.


Rinchan20:
just a heads up that a large majority of those files changed are just from screenshots. Actual changes are fixing dsm-fixtures.ts and adding an @functional tag to some tests (to run them in terminal)

aweng98:
Why are screenshots have changed?

Rinchan20:
I usually run the tests with the -u flag (test fails for me otherwise for screenshot comparison) and in order to make sure that the fix was good, I was running as many tests as possible to see if anything else needed to get fixed - so I guess that almost every screenshot got updated

aweng98:
Are you running tests on an external monitor? If yes, tests will fail on localhost for other people in future. If not, it's strange because screenshots don't fail for me.

Rinchan20:
nope, not running tests in external monitor, I'm just running the tests on laptop. Since it's just 1 commit, I'll revert and double check what's going on re: screenshots

Rinchan20:
they've been fixed in different PR

PR #2048, opened 2023-03-02
PEPPER-587: RGP DSM playwright tests_assert creation of new participant by checking DSM

Rinchan20:
Extends the current RGP playwright test to check that the newly created participant can also be seen in DSM - might need a slight tweak - the most recent changes were done using DSM Test and there's a slight text difference when selecting the RGP study there (DSM Test has `Pepper RGP`, DSM Dev has `RGP`). But other than that should be running fine. 

Finishes up the last part of of this ticket: https://broadworkbench.atlassian.net/browse/PEPPER-587

PR #2160, opened 2023-07-14
add automation of the RGP proband tab

Rinchan20:
is slightly flaky when checking that textarea notes are preserved after revisiting the page

is part 3 of breaking down this large PR: https://github.com/broadinstitute/ddp-angular/pull/2048


Rinchan20:
dev probably needs a slight tweak in order for the test to pass - doesn't seem to be housekeeping related - the tweak needed is that the RGP dynamic form -> Contact Info -> State is a textfield in dev env but is a dropdown in test env. It also appears to be a textfield in staging.  So it just needs consistency across the 3 non-prod envs. I can just change it so that it looks for a textfield instead of a dropdown though. Passes in test env though.

aweng98:
I ran all tests against Dev on CircleCI. It had one test failure which I think it's related to your change. Can you take a look see to it can be fixed? 
https://app.circleci.com/pipelines/github/broadinstitute/ddp-angular/25247/workflows/2d615392-ce1c-4a0d-a23b-77f5d79655a8

Rinchan20:
if it's the failure for finding the proband tab - that one would be because of housekeeping. Once a participant enrolls in RGP, there should be a proband tab created in RGP but sometimes in dev (and recently test - at least on Friday, I think?) it will not appear without running DSM RGP export - I think it'll be fixed once this ticket is done: https://broadworkbench.atlassian.net/browse/PEPPER-692

playwright-e2e/utils/date-utils.ts
```diff
@@ -48,3 +48,18 @@ export function offsetDaysFromToday(number = 1, opts: { isAdd?: boolean } = {}):
   }
   return today;
 }
+
+export const calculateBirthDate = (month: string, day: string, year: string): number => {
```
aweng98: This func is nice! 
How about rename to `calculateAge`? I think date is not accurate description for what the func do.

playwright-e2e/tests/rgp/dsm-family-enrollment.spec.ts
```diff
@@ -78,4 +80,432 @@ test.describe.serial('DSM Family Enrollment Handling', () => {
...
+    //select RGP study
+    await new Select(page, { label: 'Select study' }).selectOption('RGP');
+
+    //Verify the Participant List is displayed
+    const participantListPage = await navigation.selectFromStudy<ParticipantListPage>(StudyNavEnum.PARTICIPANT_LIST);
+    await participantListPage.assertPageTitle();
+    await participantListPage.waitForReady();
+    const participantListTable = new ParticipantListTable(page);
+    await participantListPage.filterListByParticipantGUID(user.patient.participantGuid);
+    await participantListTable.openParticipantPageAt(0);
```
aweng98, line 103: Need a check to assert one row exists after filter search before open participant page.

playwright-e2e/dsm/pages/participant-page/rgp/family-member-tab.ts
```diff
@@ -0,0 +1,638 @@
...
+    public getMixedRaceNotes(): Locator {
+        return this.page.locator("//textarea[contains(@data-placeholder, 'Mixed Race Notes')]");
+    }
+
+    public getContactInfoSection(): Locator {
+        return this.page.locator("//button[contains(text(), 'Contact Info')]");
+    }
+
+    public getPrimaryPhoneNumber(): Locator {
+        return this.page.locator("//td[contains(text(), 'Phone')]/following-sibling::td//div//input[@data-placeholder='Phone (Primary)']");
```
aweng98, line 286: Not a fan of long xpath. It is brittle. We can improve it at a later time.

playwright-e2e/dsm/pages/participant-page/rgp/family-member-tab.ts
```diff
@@ -0,0 +1,638 @@
...
+    constructor(page: Page, relationToProband: FamilyMember) {
+        this.page = page;
+        this._firstName; //proband starts with this as unknown, first name is used as an identifier in DSM (family member tab) & DSS (dashboard messaging)
+        this._lastName; //proband starts with this as unknown
+        this._familyID;
+        this._relationshipID;
+        this._relationToProband = relationToProband;
+    }
+
+    public set relationshipID(relationshipID : string) {
```
aweng98, line 26: Why do we need these `set` functions?

aweng98 (commented):
There is a problem with test which explains why it doesn't run on CircleCI and localhost. Failure root cause is not related to Housekeeping.
The `user` object inside the `saveParticipantGuid(participantGuid)`  function is not "known" to the test. `user` objects in the test and that func are two different objects. Let me know if you have any questions about this.

playwright-e2e/dsm/pages/participant-page/rgp/family-member-tab.ts
Rinchan20: plan is to use them at around the next PR when adding family members - but since they're not relevant for this PR I think I'll just take them out and add them back in only if they're actually being used - thinking about it more, that'll help with not causing confusion down the line when seeing a method available that's not clearly being used

aweng98 (approved):

PR #2185, opened 2023-08-14
[DSM] Fix for Onc History patch for legacy participant

pegahtah:
Making sure that for legacy particiapnts the legacy id is included in the patch instead of guid since this is how they are added in the ddp_particiapnt table.


GiCharkviani (approved):
LGTM

PR #2186, opened 2023-08-14
[DSM] PEPPER-647 Update RGP "Specialty Project R21" Search Parameters to exclude "NO" 

aweng98:
[PEPPER-647](https://broadworkbench.atlassian.net/browse/PEPPER-647)

Only change for RGP "Specialty Project R21" search filter. All other columns (checkbox) filters will not be affected.

Screenshot 2023-08-14 at 4 32 52 PM

[PEPPER-647]: https://broadworkbench.atlassian.net/browse/PEPPER-647?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/filter-column/filter-column.component.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/filter-column/filter-column.component.ts
@@ -87,4 +87,8 @@ export class FilterColumnComponent implements OnInit {
     this.dataFilter.notEmpty = false;
     this.dataFilter.empty = false;
   }
+
+  hideCheckbox(dataFilter: Filter): boolean {
+    return dataFilter?.participantColumn?.name === 'r21';
+  }
 }
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/filter-column/filter-column.component.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/filter-column/filter-column.component.ts
@@ -14,6 +14,11 @@ export class FilterColumnComponent implements OnInit {
   showOptions = false;
   selected: number;
 
+  // Hide a search parameter for the following RGP columns
+  rgpColumnFilter = [
+    { displayName: "Specialty Project: R21", name: "r21", tableAlias: "r" }
+  ];
+
   ngOnInit(): void {
     if (this.dataFilter.singleOption) {
       for (const [ key, value ] of Object.entries(this.dataFilter.selectedOptions)) {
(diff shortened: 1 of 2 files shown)


ddp-workspace/projects/ddp-dsm-ui/src/app/filter-column/filter-column.component.ts
```diff
@@ -87,4 +87,8 @@ export class FilterColumnComponent implements OnInit {
     this.dataFilter.notEmpty = false;
     this.dataFilter.empty = false;
   }
+
+  hideCheckbox(dataFilter: Filter): boolean {
```
zyme: Are there any other contextual clues in the data that we can use to minimize the chance that some other thing called `r21` will be impacted by this?  Perhaps a check somewhere on the study so that this is only applied for RGP?

zyme: Let's rename this method so it's really clear that this is a bit of a hack.  How about `hideCheckboxIfRGPR21` or some such?  Or write a method comment to this effect.

zyme: Is there anything we can do on the backend to give the frontend more information about not showing the "No" option more cleanly?

aweng98: Screenshot 2023-08-14 at 3 27 45 PM

^ screenshot of dataFilter. 

@pegahtah and I actually talkd about this. column name `r21` is special and unique to RGP study group. This change will not affect any other studies or columns.

aweng98: Suggested name is self-explanatory. I like it.

Could this be a reusable function which could be applied to other similar requirements in future. If yes, how about this change?

```
hideCheckboxIfThisColumn(dataFilter: Filter, name: string): boolean {
    return dataFilter?.participantColumn?.name === name;
  }
 ```

aweng98: @pegahtah Can you please help answer this question? I'm still learning, I don't know how to do that on the backend.

zyme (approved):

ddp-workspace/projects/ddp-dsm-ui/src/app/filter-column/filter-column.component.ts
pegahtah: @aweng98 I will check the structure and let you know

pegahtah: @aweng98 @zyme I checked backend and db, and we do not specify what options each checkbox gets. The checkbox field by default gets `YES` and `NO` and it is all set on frontend. So I think this is the best approach for removing `No`

pegahtah (approved):

PR #2189, opened 2023-08-18
[DSS] PEPPER-1014 Add more information to GCP logs

aweng98:
[PEPPER-1014](https://broadworkbench.atlassian.net/browse/PEPPER-1014)

Improvement to GCP logging.

[PEPPER-1014]: https://broadworkbench.atlassian.net/browse/PEPPER-1014?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
+++ b/ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
@@ -89,7 +89,7 @@ export class Auth0AdapterService implements OnDestroy {
                 redirectUri
             });
         } catch (e) {
-            this.log.logError(`${this.LOG_SOURCE}.createWebAuth`, e);
+            this.log.logToCloud(`${this.LOG_SOURCE}.createWebAuth`, e);
             throw new Error(e);
         }
     }
@@ -112,7 +112,7 @@ export class Auth0AdapterService implements OnDestroy {
                 redirectUri: this.configuration.auth0CodeRedirect
             });
         } catch (e) {
-            this.log.logError(`${this.LOG_SOURCE}.createLocalWebAuth`, e);
+            this.log.logToCloud(`${this.LOG_SOURCE}.createLocalWebAuth`, e);
             throw new Error(e);
         }
     }
@@ -130,9 +130,13 @@ export class Auth0AdapterService implements OnDestroy {
             sessionStorage.setItem('localAdminAuth', 'false');
             sessionStorage.setItem('localAuthParams', JSON.stringify(auth0Params));
         }
-        this.webAuth.authorize(
-            auth0Params,
-            () => this.log.logError(`${this.LOG_SOURCE}.showAuth0Modal`, 'auth0 error'));
+        // This Auth0 call gives a warning starts with "Following parameters are not allowed on the `/authorize` endpoint: " 
+        // because custom params are not found in Auth0 parameters-whitelist.js
+        try {
+            this.webAuth.authorize(auth0Params);
+        } catch (error) {
+            this.log.logToCloud(`${this.LOG_SOURCE}.showAuth0Modal`, error);   
+        }
     }
 
     /**
@@ -156,7 +160,7 @@ export class Auth0AdapterService implements OnDestroy {
     public signup(additionalParams?: Record<string, string>): void {
         const temporarySession = this.session.isTemporarySession() ? this.session.session : null;
         if (!temporarySession || !temporarySession.userGuid) {
-            this.log.logError(`${this.LOG_SOURCE}.signup.No temporal user guid`);
+            this.log.logToCloud(`${this.LOG_SOURCE}.signup.No temporal user guid`);
         }
         const params = {
             ...(temporarySession && {
@@ -170,7 +174,7 @@ export class Auth0AdapterService implements OnDestroy {
             // @todo : hack delete when done
             serverUrl: this.configuration.backendUrl
         };
-        this.log.logToCloud(`Auth0 signup modal is open for user: ${JSON.stringify(params)}`, { auth0Mode: Auth0Mode.SignupOnly })
+        this.log.logToCloud(`Auth0 signup modal is open for user ${params.temp_user_guid}: ${JSON.stringify(params)}`, { auth0Mode: Auth0Mode.SignupOnly })
             .pipe(take(1)).subscribe(() => this.showAuth0Modal(Auth0Mode.SignupOnly, params));
     }
 
@@ -192,16 +196,14 @@ export class Auth0AdapterService implements OnDestroy {
                 this.windowRef.nativeWindow.location.hash = '';
                 this.setSession(authResult);
                 this.log.logEvent(`${this.LOG_SOURCE}.handleAuthentication succeeded`);
-                this.log.logToCloud(`${this.LOG_SOURCE}.handleAuthentication, authResult: ${JSON.stringify(authResult)}`)
-                    .pipe(take(1)).subscribe();
                 this.analytics.emitCustomEvent(AnalyticsEventCategories.Authentication, AnalyticsEventActions.Login);
             } else if (err) {
-                this.log.logError(`${this.LOG_SOURCE}.handleAuthentication`, err);
+                this.log.logToCloud(`${this.LOG_SOURCE}.handleAuthentication`, err);
                 let error = null;
                 try {
                     error = JSON.parse(decodeURIComponent(err.errorDescription));
                 } catch (e) {
-                    this.log.logError(`${this.LOG_SOURCE}.handleAuthentication.Problem decoding authentication error`, e);
+                    this.log.logToCloud(`${this.LOG_SOURCE}.handleAuthentication. Problem decoding authentication error`, e);
                 }
                 if (onErrorCallback && error) {
                     // We might encounter errors from Auth0 that is not in expected
@@ -225,7 +227,7 @@ export class Auth0AdapterService implements OnDestroy {
         }
         this.adminWebAuth.authorize(
             auth0Params,
-            () => this.log.logError(`${this.LOG_SOURCE}.adminLogin`, 'auth0 error'));
+            () => this.log.logToCloud(`${this.LOG_SOURCE}.adminLogin auth0 error`));
     }
 
     public handleAdminAuthentication(onErrorCallback?: (e: any | null) => void): void {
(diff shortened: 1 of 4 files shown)


ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
```diff
@@ -89,7 +89,7 @@ export class Auth0AdapterService implements OnDestroy {
                 redirectUri
             });
         } catch (e) {
-            this.log.logError(`${this.LOG_SOURCE}.createWebAuth`, e);
+            this.log.logToCloud(`${this.LOG_SOURCE}.createWebAuth`, e);
```
aweng98, line 92: It seems there have been a couple attempts to send logs to GCP logging.

- One approach uses an experimental `stackdriver-errors-js` lib. `logError()` makes calls to this lib.

- Another approach uses Google `@google-cloud/logging` lib which is triggered by call to the Google cloud function `LoggingService`. `logToCloud()` makes calls this lib.

I think we should be consistent, use one and not both. With `logToCloud()`, we can see more information in logs such as `studyGuid`, `location.href`, etc. In case an exception occurs in `logToCloud()`, `logError()` is called to log error.

Cloud function [LoggingService](https://console.cloud.google.com/functions/details/us-central1/LoggingService?env=gen1&authuser=0&project=broad-ddp-dev&tab=source) on Dev.

Cloud function `LoggingService` log [filter](https://cloudlogging.app.goo.gl/rsrAE6LRQVhWq52HA) on Dev.

ddp-workspace/projects/ddp-sdk/src/lib/services/stackdriverErrorReporter.service.ts
```diff
@@ -30,11 +31,15 @@ export class StackdriverErrorReporterService extends ErrorHandler {
   }
 
   public handleError(error: Error | string): void {
-    if (this.config.doGcpErrorReporting) {
-      this.errorHandler.report(error);
-    }
-    // Pass the error to the original handleError otherwise it gets swallowed in the browser console
     super.handleError(error);
```
aweng98, line 34: I moved `super.handleError(error);` from the last line to the first line. Why is it not called as the first statement in an override function?

ddp-workspace/projects/ddp-sdk/src/lib/services/stackdriverErrorReporter.service.ts
```diff
@@ -30,11 +31,15 @@ export class StackdriverErrorReporterService extends ErrorHandler {
...
+    if (this.config.doGcpErrorReporting) {
+      try {
+        this.errorHandler.report(error);
+      } catch (err) {
+        console.error.apply(window.console, err);
+        throwError(() => err);
+      }
```
aweng98, line 41: Handles thrown error. Otherwise we won't know if it encountered error.

ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
```diff
@@ -59,12 +61,19 @@ export class LoggingService {
...
+            this.logEvent(`${this.LOG_SOURCE}.logToCloud Session is expired`);
+        }
+
         return this.http.post(
             url,
             body,
             { headers: new HttpHeaders({ 'Content-Type': 'application/json' }) }).pipe(
             catchError((error: any) => {
-                this.logError(this.LOG_SOURCE, `HTTP POST: ${url}. Error:`, error);
+                this.logError(this.LOG_SOURCE, `HTTP POST: ${url}`, error);
```
aweng98, line 69: Word `Error:` in `HTTP POST: ${url}. Error:` is redundant.

ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
```diff
@@ -59,12 +61,19 @@ export class LoggingService {
             httpRequest: { requestUrl: location.href, userAgent: navigator.userAgent }
         };
 
+        console.log(JSON.stringify(body, null, 2));
+        console.log("***");
+        
+        if (this.session.isSessionExpired()) {
+            this.logEvent(`${this.LOG_SOURCE}.logToCloud Session is expired`);
+        }
```
aweng98, line 62: I'm not sure what should happen next if session is expired and being caught here. For now, log to Console to know when it has happened. An expired session will cause `http.post` to fail.

ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
```diff
@@ -22,12 +22,10 @@ export class LoggingService {
...
-            const stringifiedArgs = args.map(item => (
-                 (typeof item === 'object') ? this.stringify(item) : item
-            ));
-
-            this.stackdriverErrorReporterService.handleError(stringifiedArgs.join(', '));
-            console.error.apply(window.console, stringifiedArgs);
+            args.forEach(arg => { 
+                console.error.apply(window.console, arg);
+                this.stackdriverErrorReporterService.handleError(arg); 
+            });
```
aweng98, line 28: Old way for stringify is not working if loop `item` does not have a key. 

For example, in the screenshot below,  on the right side in debugger Scope, `stringifiedArgs` is the return value. But it does not contains the the error message anymore.

logError

As the result, exception logs no longer contain the original error message.
```
Error: Logging, HTTP POST: https://us-central1-broad-ddp-dev.cloudfunctions.net/LoggingService. Error:,
```

Screenshot 2023-08-18 at 11 11 08 AM

ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
```diff
@@ -273,17 +275,17 @@ export class Auth0AdapterService implements OnDestroy {
...
-        this.log.logToCloud(`${this.LOG_SOURCE} Successfully updated session token: ${JSON.stringify(decodedJwt)}`)
+        this.log.logEvent(this.LOG_SOURCE, `Successfully updated user ${userGuid} session token`);
+        this.log.logToCloud(`${this.LOG_SOURCE} Successfully updated user ${userGuid} session token: ${JSON.stringify(decodedJwt)}`)
             .pipe(take(1)).subscribe();
     }
 
     public logout(returnToUrl: string = ''): void {
         const baseUrl = this.configuration.baseUrl;
-        this.log.logToCloud(`${this.LOG_SOURCE} logout for user`).pipe(take(1)).subscribe(() => {
+        const userGuid = this.session.session.userGuid;
```
aweng98, line 285: Adding `userGuid` to logs for differentiating multiple current users.

ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
```diff
@@ -22,12 +22,10 @@ export class LoggingService {
...
-            const stringifiedArgs = args.map(item => (
-                 (typeof item === 'object') ? this.stringify(item) : item
-            ));
-
-            this.stackdriverErrorReporterService.handleError(stringifiedArgs.join(', '));
-            console.error.apply(window.console, stringifiedArgs);
+            args.forEach(arg => {
+                console.error.apply(window.console, arg);
+                this.stackdriverErrorReporterService.handleError(arg);
+            });
```
aweng98, line 28: Old way for stringify is not working if loop `item` does not have a `[key: value]` pair.

For example, in the screenshot below,  on the right side in debugger Scope, `stringifiedArgs` is the return value. But it does not contains the the error message anymore.

logError

As the result, exception logs no longer contain the original error message.
```
Error: Logging, HTTP POST: https://us-central1-broad-ddp-dev.cloudfunctions.net/LoggingService. Error:,
```
(note: ticket for above error is PEPPER-1012. Cause is still unknown to me because I am  unable to reproduce it. But I think it's related to expired session so I added code to log that)

Screenshot 2023-08-18 at 11 11 08 AM

ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
```diff
@@ -192,16 +196,14 @@ export class Auth0AdapterService implements OnDestroy {
                 this.windowRef.nativeWindow.location.hash = '';
                 this.setSession(authResult);
                 this.log.logEvent(`${this.LOG_SOURCE}.handleAuthentication succeeded`);
-                this.log.logToCloud(`${this.LOG_SOURCE}.handleAuthentication, authResult: ${JSON.stringify(authResult)}`)
```
aweng98, line 195: Removed for security reason because it contains value of auth0 access token.

ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
GiCharkviani: Alex, the function `logToCloud()` is specifically employed to transmit general information, while `logError()` is designed for recording and tracking errors. If we were to exclusively rely on `logToCloud()`, any errors occurring wouldn't be captured and registered within the cloud-based logs

P.S. If you aim to enhance the comprehensiveness of your logs, rather than using plain strings, you can simply provide an object as an argument to the logError method. This allows you to define and include all the pertinent information

ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
GiCharkviani: Perhaps we should consult the team to determine whether they prefer it to be removed or not.

ddp-workspace/projects/ddp-sdk/src/lib/services/authentication/auth0Adapter.service.ts
GiCharkviani: + `logToCloud()` is observable. It won't work if you didn't subscribe

ddp-workspace/projects/ddp-sdk/src/lib/services/stackdriverErrorReporter.service.ts
GiCharkviani: `throwError(() => err)` should be used within an RxJS observable pipeline to emit observable-based errors.

ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
GiCharkviani: Before proceeding with these modifications, let's delve into scenarios where it might not function as intended and have a discussion about them

ddp-workspace/projects/ddp-sdk/src/lib/services/logging.service.ts
GiCharkviani: Has there been an instance of such a scenario? If it did encounter a failure, we would have an error logged

GiCharkviani (changes requested):
Alex, I appreciate your efforts on this.
I've responded to your comments.
Given that these changes will impact the application's logging approach, I believe it's essential for us to thoroughly discuss and explore each of the changes you've implemented before finalizing our decision to adopt them.

PR #2193, opened 2023-08-21
Playwright_PECGS_Samples received_saliva first and then tumor

Rinchan20:
WIP - should be the first scenario makes sure a study participant can get germline consent if saliva kit is received first and then tumor


Rinchan20:
Easypost is down - resuming WIP once it's back up
`Caused by: com.easypost.exception.EasyPostException: An error occured. Response code: 404 Response body: {"error":{"code":"NOT_FOUND","message":"The requested resource could not be found.","errors":[]}}`

Rinchan20:
dropdowns likely need a tweak

PR #2210, opened 2023-09-12
Add realm to applyDestructionPolicyToAll

denniscunningham:
PEPPER-635

We need to add the realm to this request so DSM can apply the change to the proper study. 
Verified manually locally.


GiCharkviani (approved):
LGTM

PR #2234, opened 2023-09-22
Playwright_RGP_Check proband referral source against ptp entered sources

Rinchan20:
Tweaks the existing RGP proband test in order to check to make sure that study ptp selected referral sources are reflected in the proband tab - still needs tweaks


Rinchan20:
Screenshot 2023-09-22 at 2 08 19 PM
pic of assertion when ptp provided sources is basically "More than one" but referral source in proband tab says "Doctor"

PR #2264, opened 2023-10-25
Pepper-1197 Update and turn on RGP Blood and RNA kit upload test

Rinchan20:
test that handles: https://broadworkbench.atlassian.net/browse/PEPPER-1197

Test is meant to turn back on and do the following:

- fix the part of the test that fails is there are not available kits in Kits w/o Label page
- add create label logic to be able to use the playwright test to test the fix to https://broadworkbench.atlassian.net/browse/PEPPER-1195


Rinchan20:
currently seems to work but the kit was going into kit error page so I need to do something about that

Rinchan20:
I just remembered I was testing this on DSM Test where fix is not applied yet, so that's why it goes into error

Rinchan20:
needs a tweak for when kit queue page does not have any kit requests in it - a few mins

aweng98 (approved):
LGTM. I was also creating Kit Queue page in my PR but I can just use yours after merge.
Maybe for later, file name converntion is kebab-case, like `kit-queue-page`.
Q: new test works on Dev and Test env?

zyme (approved):
LGTM!

PR #2270, opened 2023-10-31
[PEPPER-1155] [Playwright] DSM multi-tabs test

aweng98:
[PEPPER-1155](https://broadworkbench.atlassian.net/browse/PEPPER-1155)

[PEPPER-1155]: https://broadworkbench.atlassian.net/browse/PEPPER-1155?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/playwright-e2e/authentication/auth-base.ts
+++ b/playwright-e2e/authentication/auth-base.ts
@@ -14,6 +14,14 @@ export async function fillInEmailPassword(
 ): Promise<void> {
   const { email, password, waitForNavigation = true, waitForAuth = true } = opts;
 
+  await page.waitForSelector('.auth0-lock-header-welcome', { state: 'visible' });
+
+  const notYourAcctLink = page.locator('#auth0-lock-container-1 .auth0-lock-alternative-link:has-text("Not your account")');
+  try {
+    await notYourAcctLink.click({ timeout: 2000 });
+  } catch (err) {
+    // ignored
+  }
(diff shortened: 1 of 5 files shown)


playwright-e2e/authentication/auth-base.ts
```diff
@@ -14,6 +14,14 @@ export async function fillInEmailPassword(
 ): Promise<void> {
   const { email, password, waitForNavigation = true, waitForAuth = true } = opts;
 
+  await page.waitForSelector('.auth0-lock-header-welcome', { state: 'visible' });
+
+  const notYourAcctLink = page.locator('#auth0-lock-container-1 .auth0-lock-alternative-link:has-text("Not your account")');
+  try {
+    await notYourAcctLink.click({ timeout: 2000 });
+  } catch (err) {
+    // ignored
```
Rinchan20, line 23: should we put a custom error message here in case it fails?

Rinchan20 (approved):
passes for me in dev and test

playwright-e2e/authentication/auth-base.ts
aweng98: IMO, I don't think we need to log a message in catch block. I don't have have a strong preference. Let me know if you think we ought to log here.

PR #2272, opened 2023-11-08
Playwright_Pepper 902_Test to check the sending sample recevied event to dss

Rinchan20:
Relevant ticket: [https://broadworkbench.atlassian.net/browse/PEPPER-902](url)

Playwright test to check that the SAMPLE_RECEIVED event occurs regardless of normal kits (saliva or blood kits) being received first, with tumor samples second - will also check it the other way around (tumor samples received first, normal kits second)

### DIFF
--- a/playwright-e2e/dsm/component/filters/sections/customize-view.ts
+++ b/playwright-e2e/dsm/component/filters/sections/customize-view.ts
@@ -28,6 +28,17 @@ export class CustomizeView {
     await this.closeColumnsGroup();
   }
 
+  public async selectColumnsByID(columnsGroupName: string, columns: string[], id: string, opts: { nth?: number } = {}): Promise<void> {
+    const { nth } = opts;
+    this.activeColumnsGroup = columnsGroupName;
+    await this.openColumnsGroup({nth});
+    for (const column of columns) {
+      const columnOption = this.page.locator(`//ul[@id='${id}']//mat-checkbox[contains(.,'${column}')]`);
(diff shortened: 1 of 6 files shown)


Rinchan20:
Fixing - test currently stops in Onc History -> tissue Request page - doing logic in the first scenario and then will export repeated logic to a function

Rinchan20:
Scenario 1 passes in OS2 dev - implementing for the rest - the only change needed is the order of receiving samples

Rinchan20:
occasionally/frequently fails due to test going too fast for the reat of the tissue request page webelements to be available after inputting tissue received date

aweng98 (approved):
LGTM. nit-picking: a very long test. it will be harder for error troubleshooting in future.

Rinchan20:
each scenario seems to pass now - each takes about 1 min or so - made it only run in LMS so that it increases chance of pts still being available for germline consent manual and automated testing:
Screenshot 2023-11-15 at 3 50 03 PM

Rinchan20:
also - good point about the long test + agreed - the bulk of it isn't necessarily the test part but the prep part: 

1. Find a participant who fits the criteria
2. Give them a normal (saliva or blood) kit set to Sent status
3. Make them a tumor sample made ready to go

After that is just determining which is received first and confirming the germline consent activity was made

aweng98:
LGTM. Feel free to merge when you're ready.

aweng98 (approved):

PR #2276, opened 2023-11-08
Playwright_Attempt at reverting package lock json

Rinchan20:
It seems the updated package-lock.json in a previous PR might be causing build issues for the release - attempting a revert of sorts to see what happens


Rinchan20:
build error was because of something else, so closing this one up

PR #2280, opened 2023-11-14
Playwright_RGP_Fix Blood and RNA kit upload test

Rinchan20:
The RGP blood and RNA kit upload test was failing in dev - this branch is supposed to fix it by having it check Kit Queue page first and then Kit Error page


Rinchan20:
ran it a couple of times - seems to pass regardless of kit being sent to Kit Queue or Kit Error

aweng98 (approved):
LGTM

PR #2289, opened 2023-11-27
[DSM] PEPPER-1231 fixing tissue list loading issues

pegahtah:
The bug happened because parse in tissue-list.model.ts sometimes would not return anything


aweng98 (approved):
approve for now. maybe it's not needed anymore after checking with Whitney.

PR #2302, opened 2023-12-11
[DSM] PEPPER-1166 Phi Manifest Download page

pegahtah:
This PR creates the Page to download the phi manifest as a csv


ddp-workspace/projects/ddp-dsm-ui/src/app/navigation/navigation.component.ts
```diff
@@ -48,8 +48,8 @@ export class NavigationComponent implements OnInit {
 
   }
 
-  public showOncHistoryUploadPage(selectedStudy: string | null): boolean {
-    return selectedStudy && this.oncHistoryUploadPageAllowedStudies.includes(selectedStudy);
+  public showPecgsFeatures(selectedStudy: string | null): boolean {
```
zyme: This looks like it's more of a "should" than a "do it", so naming it `shouldShowPecgs...()` will help readers understand that this is checking auth, not actually showing the pecgs features.

ddp-workspace/projects/ddp-dsm-ui/src/app/phi-manifest/phi-manifest.component.html
```diff
@@ -0,0 +1,48 @@
...
+      <mat-form-field>
+        <input matInput type="text" [(ngModel)]="sequencingOrderId" placeholder="Sequencing Order Id" maxlength="200" style="width: 15cm">
+      </mat-form-field>
+
+      <button mat-raised-button type="button" color="primary"
+              (click)="getPhiReport()"
+              [disabled]="participantId == null || sequencingOrderId == null ">Create PHI Report</button>
+    </div>
+    <div>
+      <p><b>To generate a PHI REPORT:</b></p>
```
zyme: No all caps for our users please.

ddp-workspace/projects/ddp-dsm-ui/src/app/phi-manifest/phi-manifest.component.ts
```diff
@@ -0,0 +1,112 @@
...
+        }
+        this.route.queryParams.subscribe(params => {
+            this.realm = params[DSMService.REALM] || null;
+            if (this.realm != null) {
+                this.checkRight();
+            }
+        });
+    }
+
+    private checkRight(): void {
```
zyme, line 38: What's being checked here?  That this is to the right of the page?  The method name should be descriptive.

ddp-workspace/projects/ddp-dsm-ui/src/app/phi-manifest/phi-manifest.component.ts
```diff
@@ -0,0 +1,112 @@
...
+        if (sessionStorage.getItem(ComponentService.MENU_SELECTED_REALM) != null) {
+            this.realm = sessionStorage.getItem(ComponentService.MENU_SELECTED_REALM);
+            this.checkRight();
+        } else {
+            this.additionalMessage = 'Please select a PE-CGS study';
+        }
+        window.scrollTo(0, 0);
+    }
+
+    getPhiReport(): void {
```
zyme: `get()` methods should return something, but this is void.  Maybe it should be `downloadManifest()`?

ddp-workspace/projects/ddp-dsm-ui/src/app/phi-manifest/phi-manifest.component.ts
```diff
@@ -0,0 +1,112 @@
...
+            this.additionalMessage = 'Error - ' + phiManifestResponseModel.errorMessage;
+            return;
+        }
+        const phiManifestModel: PhiManifestModel = PhiManifestModel.parse(phiManifestResponseModel.data[1]);
+        const map: {}[] = [];
+        map.push(phiManifestModel.getReportAsMap(phiManifestResponseModel.data[0], phiManifestResponseModel.data[1]));
+        Utils.createCSV(
+            phiManifestResponseModel.data[0],
+            map,
+            'PHI_REPORT_PT_' + phiManifestResponseModel.ddpParticipantId + '_ORDER_' +
```
zyme: Maybe stick a datetime in here as well, so that users can download these repeatedly and not have them overwrite one another (especially if the samples/statuses in the order change).

ddp-workspace/projects/ddp-dsm-ui/src/app/phi-manifest/phi-manifest.model.ts
```diff
@@ -0,0 +1,41 @@
...
+    public getReportAsMap(headers: string[], data: string[]): any {
+        const map: Map<string, string> = new Map();
+        for (let i = 0; i < headers.length; i++) {
+            const header = headers[i];
+            map[header] = this.convertValue(data[i]);
+        }
+        return map;
+    }
+
+    convertValue(value: string): string {
```
zyme: If this is the only place that the report json is used, this conversion could be done server side to convert true/false into the string representations that the client is expecting.  PhiManifest or the service could do this, and it could be included in a server-side test.

ddp-workspace/projects/ddp-dsm-ui/src/app/phi-manifest/phi-manifest.component.html
pegahtah: Oops! Sorry! Thanks for catching this

ddp-workspace/projects/ddp-dsm-ui/src/app/phi-manifest/phi-manifest.component.ts
pegahtah: The checkRight methods are a pattern across all of the components in the DSM frontend that check if the user has rights to view this component for this study

ddp-workspace/projects/ddp-dsm-ui/src/app/phi-manifest/phi-manifest.component.ts
pegahtah: The data should not check, since downloading this report is the final step in generating shared learnings, but I will add a date just in case it helps the user when later they want to find the files they downloaded

ddp-workspace/projects/ddp-dsm-ui/src/app/phi-manifest/phi-manifest.component.ts
```diff
@@ -0,0 +1,113 @@
...
+                    const phiManifestResponse = PhiManifestResponseModel.parse(data);
+                    this.loading = false;
+                    this.downloadManifest(phiManifestResponse);
+                },
+                error: err => {
+                    // console.log(err);
+                    if (err._body === Auth.AUTHENTICATION_ERROR) {
+                        this.auth.doLogout();
+                    }
+                    this.errorMessage = 'Error - Failed to download report</b>' + err.error;
```
zyme: Putting a `</b>` might lead to formatting issues.  Where does the boldface begin in code?  If there is an expected format for the message, can we turn it into an object that has a title field and a detail field?

ddp-workspace/projects/ddp-dsm-ui/src/app/phi-manifest/phi-manifest.model.ts
```diff
@@ -0,0 +1,30 @@
...
+    static parse(json): PhiManifestModel {
+        return new PhiManifestModel(json.shortId, json.firstName, json.lastName, json.proxyFirstName,
+            json.proxyLastName, json.dateOfBirth, json.gender, json.dateOfPx, json.facility, json.sampleType,
+            json.accessionNumber, json.histology, json.mfBarcode, json.normalCollaboratorSampleId, json.blockId,
+            json.tumorCollaboratorSampleId, json.tissueSite, json.sequencingResults, json.somaticAssentAddendumResponse,
+            json.somaticConsentTumorPediatricResponse, json.somaticConsentTumorResponse, json.clinicalOrderDate,
+            json.clinicalOrderId, json.clinicalPdoNumber, json.orderStatus, json.orderStatusDate);
+    }
+
+    public getReportAsMap(headers: string[], data: string[]): any {
```
zyme: If this is returning a `Map`, can the method be declared as such?  With `any`, the caller has to dive into the code to figure out what it's actually returning (although the method name does imply it returns a Map).

ddp-workspace/projects/ddp-dsm-ui/src/app/services/dsm.service.ts
```diff
@@ -875,6 +874,15 @@ export class DSMService {
     );
   }
 
+  public getPhiReport(realm: string, guid: string, orderNumber: string): Observable<any> {
```
zyme: Similar comment here to phi-model-manifest in terms of specificity of return type.

ddp-workspace/projects/ddp-dsm-ui/src/app/phi-manifest/phi-manifest.component.spec.ts
```diff
@@ -0,0 +1,7 @@
+
+describe('Component: PhiManifestComponent', () => {
+  // it('should create an instance', () => {
```
zyme: Can we make a real test?

ddp-workspace/projects/ddp-dsm-ui/src/app/phi-manifest/phi-manifest.component.html
```diff
@@ -0,0 +1,48 @@
...
+  <br/>
+</div>
+
+<span [hidden]="!loading" class="Width--100">
+    <div align="center" class="Width--100 Height--100">
+      <i class="fas fa-spinner fa-spin fa-5x Color--primary"></i>
+    </div>
+  </span>
+
+<div *ngIf="allowedToSeeInformation">
```
zyme, line 18: It's been a while since I've looked at the corresponding backend routes--just wanted to triple check that we are doing server-side auth checks here as well so there's no way to get the manifest by messing with client-side state.

pegahtah: the `allowedToSeeInformation` gets its value from the backend call in the checkRight, more specifically here (putting code here to make sure I am addressing the right concern)
```private checkRight(): void {
        this.allowedToSeeInformation = false;
        this.dsmService.getRealmsAllowed(Statics.MEDICALRECORD).subscribe({
            next: data => {
                jsonData = data;
                jsonData.forEach((val) => {
                    if (this.realm === val) {
                        this.allowedToSeeInformation = true;
                    }
                });
                if (!this.allowedToSeeInformation) {
                    this.additionalMessage = 'You are not allowed to see information of the selected study at that category';
                }
            },
            error: () => null
        });
    }```

zyme (approved):

PR #2332, opened 2024-02-16
Playwright_Change samples received test to run on os2 instead of lms

Rinchan20:
basically as title says


aweng98 (approved):

Rinchan20:
the reason for the change is because the test ran out of valid test participants in LMS Test - either because they don't have onc history tab, were exited, or just didn't fulfill the prerequisites needed for a test participant for the samples received test. So changing it to run in OS2 for now.

PR #2354, opened 2024-03-21
Playwright_Fix DSS Tests

Rinchan20:
Errors were due to an error being thrown because in medical release form:

- `Massachusetts` when being inputted into institution state field, was going to be overwritten as `MA`
- `United States` when being inputted into institution country field, was going to be overwritten as `UNITED STATES`

need to look into why it's popping up all the sudden though

### DIFF
--- a/playwright-e2e/data/fake-user.json
+++ b/playwright-e2e/data/fake-user.json
@@ -48,7 +48,10 @@
     "address": "55 Fruit Street, Boston, MA 02114",
     "street": "55 Fruit Street",
     "city": "Boston",
-    "state": "MA",
+    "state": {
+      "name": "MASSACHUSETTS",
+      "abbreviation": "MA"
+    },
     "zip": "02114"
   },
   "adult": {
--- a/playwright-e2e/dss/component/input.ts
+++ b/playwright-e2e/dss/component/input.ts
@@ -77,8 +77,8 @@ export default class Input extends WidgetBase {
 
     // safety check: prevent unintented overwrite value in wrong input field.
     if (!overwrite) {
-      if (oldValue.length > 0 && oldValue !== value) {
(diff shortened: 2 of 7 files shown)


aweng98:
First, fix the `state` in `doctor` in `fake-user.json`.
```
"doctor": {
    "firstName": "John",
    "lastName": "Strong",
    "name": "John Strong",
    "specialty": "PRIMARY_CARE",
    "phone": "6187247123",
    "fax": "3213211234",
    "hospital": "Massachusetts General Hospital",
    "address": "55 Fruit Street, Boston, MA 02114",
    "street": "55 Fruit Street",
    "city": "Boston",
    "state": {
      "name": "MASSACHUSETTS",
      "abbreviation": "MA"
    },
    "zip": "02114"
  },
  ```
Usage:
`await medicalReleaseForm.yourHospitalInstitution().toInput('State').fill(user.doctor.state.name);`

Then, fix `Input.fill()` function with this minor tweak:
```
// safety check: prevent unintented overwrite value in wrong input field.
    if (!overwrite) {
      if (oldValue.length > 0 && !(new RegExp(`^${value.toString()}$`, 'i').test(oldValue))) {
        logError(`Input (Locator: ${this.toLocator()}): Existing value: "${oldValue}". Set "overwrite" param to true for new value: "${value}".`);
      }
    }
```

aweng98 (changes requested):
UI javascript is probably auto-fill fields with saved values from DB.

Rinchan20:
I'm looking at the videos - it's looking like it fills as Massachusetts from autosuggestion - while MA is from the user.doctor.state info in fake-uer.json

Rinchan20:
didn't see earlier comment - looking

aweng98 (approved):

PR #2362, opened 2024-04-03
PEPPER-1332 Add remote logging to somatic file uploads

zyme:
For PEPPER-1332, these changes add remote logging via the LoggingService cloud function into DSM.  Using this CF, we can start capturing more browser-side events in our GCP logs.  Example below.

Screenshot 2024-04-03 at 4 02 12 PM

Although there are already modules in DSS for `logToCloud`, there is a viper's nest of transitive angular injection dependencies that I ran into, so rather than injecting a bajillion incompatible services from DSS into DSM, I created a lightweight method in DSMService that lets us send log messages to stackdriver directly.

Along the way, I updated the `LoggingService` (the CF that receives these log events) to improve debugability.  I also noticed that node 12 is EOL, so LoggingService needs to be deployed in all 4 environments and upgraded to node 20.  Seems to work fine in dev.  I also added `GCP_PROJECT` env var when redeploying the cloud function, since the code appears to break without this in the node runtime.


ddp-workspace/projects/ddp-dsm-ui/README.md
```diff
@@ -1,16 +0,0 @@
-# DSM (DDP Study Manager) Frontend
```
zyme, line 1: Removed this since overall it was not accurate.  Moved the relevant bits into the README at the top level ddp-workspace.

ddp-workspace/README.md
```diff
@@ -11,7 +11,13 @@ npm install
 ```
 ## Launch apps
 
-Don't forget to render `ddpConfig.js/pepperConfig.js`.
+Don't forget to render `ddpConfig.js/pepperConfig.js` for study apps.
+
+For DSM, render the ddp_config.js file like so:
```
zyme, line 16: [See here](https://github.com/broadinstitute/ddp-angular/pull/2362/files#r1550392277)

logging-cloud-function/src/index.ts
```diff
@@ -38,7 +38,7 @@ export const logMessage:  HttpFunction = (req: express.Request, res: express.Res
         try {
             entry = new LogEntry(projectId, req.body);
         } catch(error) {
-            console.error("Could not process log entry: %o", req.body);
```
zyme, line 41: Added some more verbose logging to help debug issues with LoggingService

denniscunningham (approved):

PR #2363, opened 2024-04-04
Playwright_Add Expect.toPass to RGP DSM check

Rinchan20:
so that it can repeatedly query for 2 mins instead of failing the second it doesn't see the ptp in DSM - initially thought a skip was needed (hence branch name) - it just seems to need a min or so to repeatedly check if the ptp is there


playwright-e2e/tests/rgp/adult-enrollment.spec.ts
```diff
@@ -162,15 +162,19 @@ test.describe.serial('Adult Self Enrollment', () => {
...
-    await participantListPage.filterListByParticipantGUID(user.patient.participantGuid);
+    await expect(async () => {
+      await participantListPage.filterListByParticipantGUID(user.patient.participantGuid);
+      //Make sure the newly created participant can be found and it's participant page can be accessed
+      const participantListRowCount = await participantListTable.rowsCount;
+      expect(participantListRowCount).toBe(1);
+    }).toPass({
+      intervals: [10_000],
+      timeout: 120_000
+    });
```
Rinchan20, line 175: instead of checking just the once and failing if it doesn't see the guid - it repeatedly check every 10 seconds for about 2 mins - and then pass or fail

pegahtah (approved):
LGTM

PR #2365, opened 2024-04-18
[DSM] PEPPER-1245

pegahtah:
This PR addresses a bug in the data mismatcg where block IDs, represented as BLOCKID in field settings, were being converted to camel case (blockid) in the JSON documents generated for Elasticsearch (ES). This discrepancy caused issues on the frontend, which was expecting to find BLOCKID but instead received blockid.

I have made changes to convert column names to their camel case for looking at dynamic fields


ddp-workspace/projects/ddp-dsm-ui/src/app/tissue/tissue.component.ts
```diff
@@ -91,11 +92,49 @@ export class TissueComponent {
...
+  }
+
+  makeFirstLetterUpperCase(word: string): string {
+    return word.charAt(0).toUpperCase() + word.slice(1);
+  }
+
+  private makeWordsLowerCase(splittedWords: string[]): Array<string> {
+    const lowerCaseWords = splittedWords.map(word => word.toLowerCase());
+    return lowerCaseWords;
+  }
```
pegahtah, line 136: This is exactly the process the backend has for converting column names to camel case, with the same regular expression.

Rinchan20 (approved):

PR #2375, opened 2024-05-08
[DSM] PEPPER-1140 PEPPER-1256

pegahtah:
The dynamic fields were actually saving but the code to display their value was wrong. I have made a util class to change the column names for field settings and convert them to the camelCase representatin because the additionalValue data is saved with field names as camel case

### DIFF
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/participant-list.component.html
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/participant-list.component.html
@@ -750,17 +750,18 @@
                                     {{tissue[col.participantColumn.name] | noData}}
                                   </ng-container>
                                 </ng-container>
-                                <ng-container *ngIf="col.type === 'ADDITIONALVALUE' && tissue['additionalValuesJson'] != null
-                                && tissue['additionalValuesJson'][col.participantColumn.name] != null; else noTDateTemplate">
+                                <ng-container *ngIf="col.type === 'ADDITIONALVALUE' &&
+                                !DynamicValueUtilModel.isObjectEmpty(tissue['additionalValuesJson'])
+                                && getDynamicValue(tissue, col) != null; else noTDateTemplate">
                                   <ng-container *ngIf="col.additionalType === 'DATE'">
-                                    {{tissue['additionalValuesJson'][col.participantColumn.name] | date:'shortDate' | noData}}
+                                    {{getDynamicValue(tissue, col) | date:'shortDate' | noData}}
                                   </ng-container>
                                   <ng-container *ngIf="col.additionalType === 'CHECKBOX'">
                                     {{getUtil().getYesNo(tissue['additionalValuesJson'][col.participantColumn.name]) | noData}}
                                   </ng-container>
                                   <ng-container
                                     *ngIf="col.additionalType !== 'DATE' && col.additionalType !== 'CHECKBOX'">
-                                    {{tissue['additionalValuesJson'][col.participantColumn.name] | noData}}
+                                    {{getDynamicValue(tissue, col) | noData}}
                                   </ng-container>
                                 </ng-container>
                                 <ng-container
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/participant-list.component.html
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/participant-list.component.html
@@ -750,9 +750,7 @@
                                     {{tissue[col.participantColumn.name] | noData}}
                                   </ng-container>
                                 </ng-container>
-                                <ng-container *ngIf="col.type === 'ADDITIONALVALUE' &&
-                                !DynamicValueUtilModel.isObjectEmpty(tissue['additionalValuesJson'])
-                                && getDynamicValue(tissue, col) != null; else noTDateTemplate">
+                                <ng-container *ngIf="isAdditionalValuePresent(col, tissue); else noTDateTemplate">
                                   <ng-container *ngIf="col.additionalType === 'DATE'">
                                     {{getDynamicValue(tissue, col) | date:'shortDate' | noData}}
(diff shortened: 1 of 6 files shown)


ddp-workspace/projects/ddp-dsm-ui/src/app/tissue/tissue.component.ts
```diff
@@ -76,65 +76,18 @@ export class TissueComponent {
...
-      if (this.tissue.additionalValuesJson[colName] != null) {
-        return this.tissue.additionalValuesJson[colName];
-      } else if (this.tissue.additionalValuesJson[this.convertToCamelCase(colName)] != null) {
-        return this.tissue.additionalValuesJson[this.convertToCamelCase(colName)];
-      }
-    }
-    return null;
-  }
-
-  convertToCamelCase(str: string): string {
```
zyme, line 102: This has moved to `dynamic-value-util.model.ts`, right? 
 Any changes, or just a move?

ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/participant-list.component.ts
```diff
@@ -1321,6 +1323,7 @@ export class ParticipantListComponent implements OnInit {
     } else {
       this.selectedColumns[ parent ].push(column);
     }
+    console.log(this.selectedColumns);
```
zyme: Do we need this in prod code, or was this something we needed along the way during development?

ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/participant-list.component.html
```diff
@@ -750,17 +750,18 @@ <h3 class="Color--primary Line--Break">{{additionalMessage}}</h3>
                                     {{tissue[col.participantColumn.name] | noData}}
                                   </ng-container>
                                 </ng-container>
-                                <ng-container *ngIf="col.type === 'ADDITIONALVALUE' && tissue['additionalValuesJson'] != null
-                                && tissue['additionalValuesJson'][col.participantColumn.name] != null; else noTDateTemplate">
+                                <ng-container *ngIf="col.type === 'ADDITIONALVALUE' &&
```
zyme: Can this code be moved into ts source?  Having logic beyond a single method call inside of the html makes it harder to read, document, and test.

ddp-workspace/projects/ddp-dsm-ui/src/app/tissue/tissue.model.ts
```diff
@@ -41,4 +42,21 @@ export class Tissue {
       null,null,null,null,null,null,
       null,null,null,null,null, false);
   }
+
+  getAdditionalValue(colName: string): string {
```
zyme, line 46: Consider this approach instead, which does the checks once.  The code as written now is error prone because it's got copy-pasted transformation in the if block and in the return statement.  So it's easy to change the conditional and then forget to apply the same change to the return statement.

```
const valueAtColumnKey = this.additionalValuesJson[colName];
    const valueAtCamelCaseKey = this.additionalValuesJson[DynamicValueUtilModel.convertToCamelCase(colName)];
    const valueAtLowerCaseKey = this.additionalValuesJson[colName.toLowerCase()];
    if (this.additionalValuesJson != null) {
      if (valueAtColumnKey != null) {
        return valueAtColumnKey;
      } else if (valueAtCamelCaseKey != null) {
        return valueAtCamelCaseKey;
      } else if (valueAtLowerCaseKey != null) {
        //as a last resort, try to get the value by converting the column name to lowercase
        return valueAtLowerCaseKey;
      }
    }
    return null;
```

ddp-workspace/projects/ddp-dsm-ui/src/app/tissue/tissue.component.ts
pegahtah: Just a move since it was being referenced in both component and model and html files, and I suspect in future we will use that util class in other places like maybe onc history and mr fieldsettings as well (I did not do a full investigation around if the same problem is happening in other places yet, but given the way backend is formed, it might)

ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/participant-list.component.html
pegahtah: Of course!

ddp-workspace/projects/ddp-dsm-ui/src/app/participant-list/participant-list.component.ts
pegahtah: omg I missed this, thanks!

ddp-workspace/projects/ddp-dsm-ui/src/app/utils/dynamic-value-util.model.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+    return this.makeCamelCaseFrom(this.makeWordsLowerCase(splittedWords));
+  }
+
+  private static handleAllUppercase(word: string): string {
+    return this.isCamelCase(word) ? word : word.toLowerCase();
+  }
+
+  private static isCamelCase(word: string): boolean {
+    return this.camelCaseRegEx.exec(word).filter((value) => value === word).length > 0;
```
pegahtah, line 21: @zyme this is the only change when moving the code to a separate class. I found out that using `RegExp.test()` checks if the reg expression is present in the string at all, not that if it matches the full string, so I am doing an exec, which finds all of the matches in the string, and then I check if the full string was one of the matches that were found

ddp-workspace/projects/ddp-dsm-ui/src/app/tissue/tissue.model.ts
pegahtah: Thanks this is very helpful

zyme (approved):

PR #2378, opened 2024-05-10
PEPPER-1332 cors fix and button coloring

zyme:
After a round of integration testing PEPPER-1332, a few things became clear:

1) sending the `x-skip-interceptor` custom header up to the server is a bad idea in general since it's not intended for server-side consumption, but specifically, it causes a bunch of cors problems since it's a nonstandard header.
2) the color of the upload button should indicate an error.

Both things are fixed with this PR.  1) is fixed by removing the header in the same interceptor.  2) is fixed by using the existing css class that is already used for upload failures.

### DIFF
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/interceptors/Http-interceptor.service.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/interceptors/Http-interceptor.service.ts
@@ -23,13 +23,16 @@ export class HttpInterceptorService implements HttpInterceptor {
               private readonly authService: Auth) {}
 
   intercept(req: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
-    return next.handle(req).pipe(
+    var ignoreError = req?.headers?.has(HttpInterceptorService.InterceptorSkipHeader);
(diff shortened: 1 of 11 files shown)


ddp-workspace/projects/ddp-dsm-ui/src/app/interceptors/Http-interceptor.service.ts
```diff
@@ -23,13 +23,16 @@ export class HttpInterceptorService implements HttpInterceptor {
               private readonly authService: Auth) {}
 
   intercept(req: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
-    return next.handle(req).pipe(
+    var ignoreError = req?.headers?.has(HttpInterceptorService.InterceptorSkipHeader);
+    // InterceptorSkipHeader is only used internally to DSM.  don't pass it on to servers.
+    const cleanedHeaders = req.clone().headers.delete(HttpInterceptorService.InterceptorSkipHeader);
+    var cleanReq = req.clone({ headers: cleanedHeaders });
```
zyme, line 29: Seems like a lot of angular code out there does this, although mostly it's do add auth headers.  In this case, we clone the headers, remove the interceptor skip header, and then run the request.

denniscunningham (approved):

PR #2387, opened 2024-06-04
Playwright_Fix pepper-999 and pepper-1000 additions

Rinchan20:
basically, forgot to adjust the parameter for tests that already use the modified method - seems to need at least 1 more fix separate fix to bring the tests back into shape


Rinchan20:
will run all the tests in feature branch after initial circleci tests are done in order to make sure it's all properly fixed

Rinchan20:
confirmed fixed  after running all tests

zyme (approved):

PR #2389, opened 2024-06-05
PEPPER-1314 fix for blank string as dates

pegahtah:
This is a fix for this error on prod which thrown when a blank string is being passed for the request date value. This may or may not fix the PEPPER-1314 but since we can't reproduce the error this is a test to see if it helps

### DIFF
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-page/participant-page.component.ts
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/participant-page/participant-page.component.ts
@@ -652,8 +652,7 @@ export class ParticipantPageComponent implements OnInit, OnDestroy, AfterViewChe
         v = value.checked;
       }
     }
-    if (v !== null) {
-
+    if (v !== null || parameterName.includes('faxSent')) {
       const patch1 = new PatchUtil(
         oncHis.oncHistoryDetailId, this.role.userMail(), {name: parameterName, value: v},
         null, 'participantId', oncHis.participantId, Statics.ONCDETAIL_ALIAS, null,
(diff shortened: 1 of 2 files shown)


ddp-workspace/projects/ddp-dsm-ui/src/app/participant-page/participant-page.component.ts
```diff
@@ -652,8 +652,7 @@ export class ParticipantPageComponent implements OnInit, OnDestroy, AfterViewChe
         v = value.checked;
       }
     }
-    if (v !== null) {
-
+    if (v !== null || parameterName.includes('faxSent')) {
```
zyme: Since we have `faxSent` hardcoded in a few places, it might be helpful to pull this out to a constant.  And since this check will apply to faxSent, faxSent2, and faxSent3, it might be better to check those values explicitly, on the off chance that we make a `faxSent` variant that shouldn't trigger this logic.

ddp-workspace/projects/ddp-dsm-ui/src/app/participant-page/participant-page.component.ts
```diff
@@ -830,18 +829,18 @@ export class ParticipantPageComponent implements OnInit, OnDestroy, AfterViewChe
               // eslint-disable-next-line @typescript-eslint/no-shadow
               const date = new Date();
               const formattedDate = Utils.getFormattedDate(date);
-              if (oncHis.faxSent == null) {
+              if (!oncHis.faxSent) {
                 oncHis.faxSent = formattedDate;
                 oncHis.faxSentBy = this.role.userID();
                 this.oncHistoryValueChanged(oncHis.faxSent, 'faxSent', oncHis);
-              } else if (oncHis.faxSent2 == null) {
+              } else if (!oncHis.faxSent2) {
```
zyme, line 844: Do we need to do the same for `faxSent3`?

ddp-workspace/projects/ddp-dsm-ui/src/app/utils/utils.ts
```diff
@@ -53,7 +53,7 @@ export class Utils {
     if (date != null) {
       return date.toString();
     }
-    return '';
+    return null;
```
zyme: This method looks like it's called in about 40 places.  Are we sure that they all will handle the change from '' to null?

ddp-workspace/projects/ddp-dsm-ui/src/app/participant-page/participant-page.component.ts
zyme: Or a separate method like `isFaxParameter` that can take in a parameter name and return whether or not it's a parameter that is subject to this special handling.

ddp-workspace/projects/ddp-dsm-ui/src/app/utils/utils.ts
pegahtah: Changed this so it is set to null only for fax values in onc history

ddp-workspace/projects/ddp-dsm-ui/src/app/participant-page/participant-page.component.ts
```diff
@@ -829,19 +833,22 @@ export class ParticipantPageComponent implements OnInit, OnDestroy, AfterViewChe
...
                 //If current date is not already on FaxSent1
                 if (oncHis.faxSent !== formattedDate) {
                   oncHis.faxSent2 = formattedDate;
                   oncHis.faxSent2By = this.role.userID();
                   this.oncHistoryValueChanged(oncHis.faxSent2, 'faxSent2', oncHis);
                 }
-              } else if (oncHis.faxSent3 == null) {
+              } else if (!oncHis.faxSent3) {
```
pegahtah, line 851: This is where the faxSent3 is handled

zyme (approved):

PR #2397, opened 2024-07-01
PEPPER-1349 clinical order page updates

ssettipalli:
Add styling to control width for clinical order page json elements and make json elements available only in download

### DIFF
--- a/ddp-workspace/projects/ddp-dsm-ui/src/app/clinical-page/clinical-page.component.html
+++ b/ddp-workspace/projects/ddp-dsm-ui/src/app/clinical-page/clinical-page.component.html
@@ -37,11 +37,9 @@
       <td>{{order.orderId}}</td>
       <td>{{getDateFormatted(order.orderDate)}}</td>
       <td>{{getOrderStatus(order.orderStatus)}}</td>
-      <td class="statusEl">{{order.statusDetail}}</td>
+      <td>{{order.statusDetail}}</td>
       <td class="jsonEl">Click on 'download list' to view order message</td>
       <td class="jsonEl">Click on 'download list' to view status message</td>
-      <!-- <td class="jsonEl">{{order.orderMessage}}</td>
-      <td class="jsonEl">{{order.statusMessage}}</td> -->
     </tr>
     </tbody>
   </table>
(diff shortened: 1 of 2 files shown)


ddp-workspace/projects/ddp-dsm-ui/src/app/clinical-page/clinical-page.component.html
```diff
@@ -38,9 +38,8 @@ <h1> Clinical Orders </h1>
       <td>{{getDateFormatted(order.orderDate)}}</td>
       <td>{{getOrderStatus(order.orderStatus)}}</td>
       <td>{{order.statusDetail}}</td>
-      <td>{{order.orderMessage}}</td>
-      <td>{{order.statusMessage}}</td>
-
+      <td class="jsonEl">Click on 'download list' to view order message</td>
```
ssettipalli: Message can be limited to Click on 'download list' if we want keep it to one line rather than two lines.

ssettipalli: Wondered if these json elements heading should be more of like "Request Message" and "Response Message" rather than "Order Message" and "Status Message" ! DSM DB has it as "OrderMessage" and "StatusMessage"

ssettipalli:
Screenshot 2024-07-01 at 9 10 52 AM

ddp-workspace/projects/ddp-dsm-ui/src/app/clinical-page/clinical-page.component.html
zyme: How about moving the text out of the table, maybe just above the table as a little block of text?

ssettipalli: I like that. 
Does this look ok ?

Screenshot 2024-07-01 at 11 07 16 AM

zyme: Looks great!

zyme (approved):

PR #2408, opened 2024-08-28
fix for when test gets a ptp with v2 instead of v3 of research consent

Rinchan20:
basically as the tittle says - the tests passes fine in dev - but in test env seems to get ptps with v2 of consent - this is a slight change so that whatever version it sees is valid

### DIFF
--- a/playwright-e2e/tests/dsm/participant/participant-phone-number-info-displayed.spec.ts
+++ b/playwright-e2e/tests/dsm/participant/participant-phone-number-info-displayed.spec.ts
@@ -162,30 +162,11 @@ test.describe(`Confirm that participant phone number information is displayed @d
 })
 
 async function getDisplayedConsent(surveyDataTab: SurveyDataTab): Promise<Locator> {
-  const consentFormVersionOne = await surveyDataTab.getActivity({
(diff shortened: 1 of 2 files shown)


playwright-e2e/tests/dsm/participant/participant-phone-number-info-displayed.spec.ts
```diff
@@ -172,3 +160,32 @@ test.describe(`Confirm that participant phone number information is displayed @d
     })
   }
 })
+
+async function getDisplayedConsent(surveyDataTab: SurveyDataTab): Promise<Locator> {
+  const consentFormVersionOne = await surveyDataTab.getActivity({
+    activityName: SurveyName.RESEARCH_CONSENT_FORM,
```
zyme, line 166: Is there a way to make the test pass regardless of version?  It doesn't look like we're doing different assertions based on version, and if we can make the test do what it needs to do without version, we won't have to change the test again for consent v4.

Rinchan20: good point - will make the change

zyme (approved):

PR #2436, opened 2024-12-11
PEPPER-1561 . check for non login/active session before invoking GA

ssettipalli:
PEPPER-1561
check for non active session (not logged in) before invoking GA

### DIFF
--- a/ddp-workspace/projects/ddp-lms/src/app/app.module.ts
+++ b/ddp-workspace/projects/ddp-lms/src/app/app.module.ts
@@ -203,8 +203,8 @@ export class AppModule {
     private router: Router) {
     this.renderer = rendererFactory.createRenderer(null, null);
     this.router.events.subscribe(event => {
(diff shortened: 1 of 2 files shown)


ddp-workspace/projects/ddp-lms/src/app/app.module.ts
```diff
@@ -203,7 +203,7 @@ export class AppModule {
     private router: Router) {
     this.renderer = rendererFactory.createRenderer(null, null);
     this.router.events.subscribe(event => {
-      if (event instanceof NavigationEnd) {
+      if (localStorage.getItem('session_key') == null && event instanceof NavigationEnd) {
         if (event.url !== 'undefined' &&
```
ssettipalli, line 207: zim:
Maybe we can avoid looking for specific static page links but its ok to be extra safer and left it in there

Rinchan20 (approved):

PR #2446, opened 2025-01-27
Playwright-change toHaveScreenshot to toMatchAriaSnapshot

Rinchan20:
change is due to bug occurring with toHaveScreenshot in this specific part. Using toMatchAriaSnaptshot instead was a recommendation seen for getting around bug.

Need to run the tests in circleci to see if it fixes things since that sees to be the only way to reproduce the bug

### DIFF
--- a/playwright-e2e/package.json
+++ b/playwright-e2e/package.json
@@ -15,7 +15,7 @@
   },
   "devDependencies": {
     "@faker-js/faker": "^8.3.1",
-    "@playwright/test": "^1.41.0",
+    "@playwright/test": "^1.50.0",
     "@types/adm-zip": "^0.5.5",
     "@types/file-saver": "^2.0.5",
     "@types/mailparser": "^3.4.0",
--- a/playwright-e2e/tests/pancan/enrollment/redirect-glioblastoma-study.spec.ts
+++ b/playwright-e2e/tests/pancan/enrollment/redirect-glioblastoma-study.spec.ts
@@ -32,7 +32,14 @@ test.describe('Redirect to Brain cancer project', () => {
     // See a popup redirects
     const modal = new Modal(page);
     await expect(modal.toLocator().locator('.confirm-dialog-title')).toHaveText('Your participation');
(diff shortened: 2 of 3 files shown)


Rinchan20:
ran the tests in circleci and it seems to be fixed now

playwright-e2e/package.json
```diff
@@ -15,7 +15,7 @@
   },
   "devDependencies": {
     "@faker-js/faker": "^8.3.1",
-    "@playwright/test": "^1.41.0",
+    "@playwright/test": "^1.50.0",
```
Rinchan20, line 18: this update is so that I could use the toMatchAriaSnapshot() method which was introduced in Playwright version 1.49 from what I understand

playwright-e2e/tests/pancan/enrollment/redirect-glioblastoma-study.spec.ts
```diff
@@ -32,7 +32,14 @@ test.describe('Redirect to Brain cancer project', () => {
     // See a popup redirects
     const modal = new Modal(page);
     await expect(modal.toLocator().locator('.confirm-dialog-title')).toHaveText('Your participation');
-    await expect(modal.toLocator()).toHaveScreenshot('modal-redirect.png');
```
Rinchan20, line 35: switched from toHaveScreenshot to toMatchAriaSnapshot because a bug where the resulting snapshot is stretched and bigger than it should be was occurring. Using toMatchAriaSnapshot seems to have fixed it.

Rinchan20: to be more clear it started looking like the below - and only on a circleci run (practically nowhere else):
Screenshot 2025-01-28 at 9 06 37 AM

Rinchan20: including thread where people seem to be discussing the bug (or a similar bug): https://github.com/microsoft/playwright/issues/33843

denniscunningham (approved):
Nice detective work!

